PR #13, opened 2024-11-27
Do not manually map numeric error

tcharding:
We provide `From` impls for all errors that include a `NumericError` - use it.


apoelstra (approved):
ACK 93af3a84c69c1da75c247bfdf36b971891c282c8; successfully ran local tests

PR #21, opened 2024-12-04
Public export of `serde_json`

### LINKED ISSUE #19: Feature request for `invalidateblock` RPC
A somewhat hidden but useful command for development is the `invalidate_block` call that instructs `bitcoind` to treat a block as if it violates consensus. In combination with `generate_to_address` this may be used to simulate chain reorganizations on Regtest. It does not actually come up when calling `bitcoin-cli help`, but it was present within `bitcoincore-rpc` [here](https://docs.rs/bitcoincore-rpc/latest/src/bitcoincore_rpc/client.rs.html#921-923). I tried taking a stab at adding the macro to `v17` but I got lost as to how to implement it for each subsequent version, and I am also unsure what section it should go under. 

Maybe I am missing a different command that would reorganize blocks, any insight on that front is welcome.

ref: https://github.com/bitcoin/bitcoin/blob/ff873a20a7f48698da3334acc570d97a07016ddf/src/rpc/blockchain.cpp#L1619

rustaceanrob:
Closes #19 

Users may construct `serde_json::Value` to use with `Client::call`.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -18,7 +18,7 @@ use anyhow::Context;
 use corepc_client::client_sync::{self, Auth};
 use log::{debug, error, warn};
 use tempfile::TempDir;
-pub use {anyhow, tempfile, which};
+pub use {anyhow, serde_json, tempfile, which};
 
 #[rustfmt::skip]                // Keep pubic re-exports separate.
 #[doc(inline)]


apoelstra (approved):
ACK ef9d367852060f12ad28de437582542dba0ee2a8; successfully ran local tests

PR #27, opened 2024-12-06
Reduce number of supported minor versions

tcharding:
We support Core back to v0.17.1 and currently explicitly test all minor versions back to v22 - this is probably an overkill and is only going to increase as Core progresses. Instead lets just test all minor versions of the last three releases and then the latest minor version for older releases.

This has the benefit of reducing the number of CI jobs.


tcharding:
I'm going to merge this but if you get a chance @apoelstra could you post-merge concept ack this please. I can then revert if you think it was not a good idea. Thanks man.

apoelstra:
concept ACK. Yeah, let's just support the latest minor version of each major version.

tcharding:
Even for the last three major releases? Currently we do all minor versions for the last 3 major versions and just the latest minor version for all older major versions.

apoelstra:
Hmm, this raises an interesting question -- removing old releases is a breaking change.

So maybe our policy should be "every time there's a major rev we delete all old minor releases".

But yes, probably we should have all the minor releases for the most recent major version. As for older ones, I leave it up to you to decide whether it's worth the maintenance burden.

tcharding:
> So maybe our policy should be "every time there's a major rev we delete all old minor releases".

The only real burden is more CI jobs so I rekon its ok to keep all minor versions for the last 3 major releases. I seem to remember reading somewhere that the last 3 major releases were more actively maintained - or have I just imagined that?

apoelstra:
Sounds good to me!

PR #35, opened 2024-12-13
Run `verify` in CI

tcharding:
The `verify` crate is a dev tool and should not have been included in the workspace because it has no lock file requirements.

Exclude `verify` from the workspace and update the lock files and add a job that runs it from CI.

### DIFF
--- a/.github/workflows/README.md
+++ b/.github/workflows/README.md
@@ -22,6 +22,7 @@ Run from `rust.yml` unless stated otherwise. Total 11 jobs.
 8.  `Docs`
 9.  `Docsrs`
 10. `Format`
+11. `Verify`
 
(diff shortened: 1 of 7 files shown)


apoelstra:
We can do this, but FYI if you remove stuff from the workspace I will never test it again because it's effectively invisible to me.

tcharding:
hmm, I'll put some thought into whether we want to actually run `verify` in CI or whether it being invisible is ok.

tcharding:
Gotta run, if this goes red don't waist your time on in @apoelstra I'll come back.

apoelstra (approved):
ACK b2d962ca633a51512d3bc7db6f8917a2c8d0b65e; successfully ran local tests; though "local tests" no longer include `verify` :)

Issue #41, opened 2025-01-21
Core v28 support is broken


tcharding:
The `node` crate has a unit test `test_multi_p2p` that does not work for Bitcoin Core v28. I _think_ the reason is something to do with how port binding was changed in this release. See [P2P and Network Changes](https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-28.0.md#p2p-and-network-changes) section of the release notes.

To close this issue remove the feature guard on the `test_multi_p2p` unit test.

tcharding:
Suggested fix here: https://github.com/RCasatta/electrsd/pull/89#issuecomment-2606043981

PR #42, opened 2025-01-21
Remove feature gating from `test_p2p` unit test

tcharding:
This test works for `28_0`. It does not need feature gating out.

The reason this was done is because it cannot be run in parallel with `test_multi_p2p`. We can leave that one feature guarded out in v28 builds.

The underlying issue is a problem with how Bitcoin Core binds to ports and is tracked in #41

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -577,7 +577,6 @@ mod test {
     use tempfile::TempDir;
 
     use super::*;
-    #[cfg(not(feature = "28_0"))]
     use crate::P2P;
     use crate::{exe_path, get_available_port, Conf, Node, LOCAL_IP};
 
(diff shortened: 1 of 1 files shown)


apoelstra:
It looks like we need to update our Cargo.toml to fix the "unknown cfg parameter" lint, which is now many months old.

I don't think I'll be able to run my local CI on these things without it (and a new clippy run).

Can we do that first?

tcharding:
Yes definitely. I saw this was red yesterday but was lazy, sorry for taking your time while you looked.

tcharding:
Strange that a v17 unit test failed. I re-ran the test.

tcharding:
I'm going to no-ack merge this, it only removes feature gating from test code and is needed by #45.

Issue #12, opened 2024-11-27
Enable no-std for `types` crate


tcharding:
There is no reason that we should require `std`. Lets keep a hard requirement on an allocator though.

PR #44, opened 2025-01-22
Add `std` feature

tcharding:
`master` is currently broken because the linter is warning because of a `std` feature (which is coming from our usage of `internals::write_err`).

We can fix this breakage after observing that calling macros from other crates is error prone. Lets copy the macro. On the way add a `std` feature and stop using `HashMap`.

This also allows the removal of the `internals` dependency. Sexy PR right here.

Close: #12

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -120,7 +120,6 @@ name = "corepc-types"
 version = "0.5.0"
 dependencies = [
  "bitcoin",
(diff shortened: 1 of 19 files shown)


apoelstra (approved):
ACK 3a72c197a55dc84b7bc5560c205ac6afc37573b8; successfully ran local tests

tcharding:
Mad! thanks man

Issue #40, opened 2025-01-21
node: The default feature is a footgun


tcharding:
`corepc-node` has a default feature `28_0` - this is an unnecessary footgun because it overides the version supplied in another feature e.g., `cargo test --features=25_2` runs tests using Core v28 still because default features are enabled.

tcharding:
Found here: https://github.com/RCasatta/electrsd/pull/89#issuecomment-2603379285

apoelstra:
Good catch. I guess we should disable all features by default and have a compile error telling users that they need to enable one.

PR #45, opened 2025-01-22
node: Remove `default` feature

tcharding:
The default feature in node overides an explicit version feature. This is a footgun. Currently the logic around using `bitcoind` from the environment is borked.

Also, features in the `node` crate are unusual in that at least one version feature is expected. Lets make this a build requirement.

Maintain the current behaviour in relation to `--all-features` i.e., all features implies latest version 28.0
    
- Remove `default` feature
- Remove `download` from version number features
- Use `compile_error` to catch missing version feature
 

Fix: #40

### DIFF
--- a/integration_test/Cargo.toml
+++ b/integration_test/Cargo.toml
@@ -51,7 +51,7 @@ TODO = []                       # This is a dirty hack while writing the tests.
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false, features = ["std", "serde"] }
 client = { package = "corepc-client", version = "0.5.0", default-features = false, features = ["client-sync"] }
-node = { package = "corepc-node", version = "0.5.0", default-features = false, features = [] }
+node = { package = "corepc-node", version = "0.5.0", default-features = false, features = ["download"] }
 rand = "0.8.5"
 env_logger = "0.9.0"
 
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -35,36 +35,37 @@ anyhow = "1.0.66"
 
 # Please note, in this crate the version features are mutally exclusive.
 #
-# - `cargo test --all-features` is the same as `cargo test --features=v28_0`
-# - To use --no-default-features you must configure the bitcoind binary to be version 28 either by having `bitcoind` in your path or using `BITCOIND_EXE=/path/to/bitcoind`.
-# - `cargo test --no-default-features` uses `v28_0` also.
+# - `cargo test --features=27_2,download` to download Bitcoin Core binary v27.
+# - `cargo test --all-features` is the same as `cargo test --features=download,v28_0`
+# - `cargo test --features=28_0` to use `bitcoind`.
+#
+# If you do not enable `download` then you must configure the bitcoind binary by `bitcoind` in your
+# path or using `BITCOIND_EXE=/path/to/bitcoind`. Note also that the feature enabled should match
+# the version from `bitcoind --version`.
 [features]
-default = ["28_0"]
-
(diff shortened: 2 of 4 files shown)


tcharding:
I'm interested to see if this one gets past your CI @apoelstra.

apoelstra:
Sure, will run this through local CI and ack it. But I'm not convinced that `--no-default-features` **should** work. It seems like it would be fine if it were a compiler error telling the user to pick a Core version.

We maybe also ought to warn if multiple features are enabled.

apoelstra (approved):
ACK 8d08689571ae42e106c4913ee2bad58a90a50005; successfully ran local tests

tcharding:
>  But I'm not convinced that --no-default-features should work. 

hmm. You could be right, its `--all-features` that **has** to work isn't it. I'll put some more thought into it.

apoelstra:
Yeah, true, we had that with the `no-std` vs `std` features in rust-bitcoin. I agree it was super annoying.

**But** this is a somewhat-obscure leaf crate, which will make it less annoying, and I also think that users would like to be forced to choose a Core version rather than unknowingly going with "the default" and finding that this version changes out from under them.

tcharding:
Re-done the PR, re-wrote PR description. Implemented your suggestion. `--no-default-features` no longer builds, an explicit version number feature must be provided.

apoelstra (approved):
ACK bd8bdb4b0c7f0c91af1ac1f45e6b9550d2ce46a4; successfully ran local tests; thanks, I think this is the right approach

apoelstra:
Heads up that my local CI didn't really test this change; it does not run workspaces that are in the `exclude` list, which both `node` and `integration_test` are.

I assume we intend downstream users to use the `node` crate -- why is this excluded?

Let me know if you want to change anything related to that or if we should just merge this as is.

tcharding:
I had to take a look to see why we exclude it and I rekon its because it doesn't work with `run_task` since the features are not used like 'normal'. I'm going to merge, I believe the test coverage is good. Its a bit sad we loose linting and whatnot but FTR I un-excluded it and ran `cargo test  --color=always --workspace --all-features --all-targets` as well as `just lint`.

apoelstra:
Can you update `run_task` to allow excluding specific feature combos? Basically we just need to exclude the empty feature set. 

This is a fair bit simpler than the "use arbitrary feature sets but make sure that either `std` or `no-std` is present" situation that we used to have with rust-bitcoin.

tcharding:
> Basically we just need to exclude the empty feature set.

Combinations of features won't work either (the feature matrix stuff).

apoelstra:
We should make it a goal to reenable this. Other than the empty set, every combination of features should work.

tcharding:
I started writing an issue but just ended up throwing #53 up.

tcharding:
In hindsite this may not have been the optimum solution. #53 has been a struggle, some of which could have been avoided if instead of removing the default we defaulted to 0_17_1 (the earliest version).

The reason this still solves the problem solved in this PR is that the higher default feature was a footgun because higher versions override lower ones - defaulting to the lowest does not have this problem.

apoelstra:
That would have helped, but I think the 0_17_1 issue I've solved. I'm struggling now with having a `download` feature that makes it impossible to build anything without Internet access.

tcharding:
What? This definitely builds without `download` - I ran the tests locally with bitcoind v17 without downloading.

apoelstra:
Yes. But you can't build *with* `download` unless you have an Internet connection. You can't even run clippy.

tcharding:
Oh, now I get it. I don't see a solution to that.

PR #50, opened 2025-01-27
fix: fix clippy warnings

rex4539:


types/src/v28/blockchain.rs
```diff
@@ -13,7 +13,6 @@ use super::{GetBlockchainInfoError, Softfork};
 use crate::model;
 
 #[rustfmt::skip]                // Keep public re-exports separate.
-
```
apoelstra, line 16: In a81e7c95517a5237990b7dea58ac6431e7efff4a:

Can you remove this unrelated change?

tcharding: We can just remove the whole attribute. Its meant to be on `pub use Foo` statements. It was incorrectly added in fcd96963

tcharding:
Thanks for the contribution! I'm happy to ack this however you do it since I already looked at the attribute thing. If one was to be picky the attribute change could go in a separate patch.

Thanks again, appreciate your effort.

rex4539:
I reverted that change so anybody feel free to fix it as you find suitable.

tcharding (approved):
ACK 07a8444cd9d5bd3fcc5279da85eb8cdb6febd6d5

apoelstra (approved):
ACK 07a8444cd9d5bd3fcc5279da85eb8cdb6febd6d5; successfully ran local tests

PR #51, opened 2025-01-28
Remove unnecessary #[rustfmt::skip]

jamillambert:
As mention here https://github.com/rust-bitcoin/corepc/pull/50#discussion_r1932753953 this has been removed.

### DIFF
--- a/types/src/v28/blockchain.rs
+++ b/types/src/v28/blockchain.rs
@@ -12,8 +12,6 @@ use serde::{Deserialize, Serialize};
 use super::{GetBlockchainInfoError, Softfork};
 use crate::model;
 
-#[rustfmt::skip]                // Keep public re-exports separate.
-
 /// Result of JSON-RPC method `getblockchaininfo`.
 ///
 /// Method call: `getblockchaininfo`


tcharding:
As an exercise in doing PRs can you write in the git log _why_ it is unnecessary. Well done only linking to github related stuff in the PR description and not in the git log - I like it.

jamillambert:
Git log updated.  No other changes.

apoelstra (approved):
ACK 22e0d07a841f7279e6ac87f349dc6f2ea147e6de; successfully ran local tests

tcharding:
>     This attribute is only meant to be on `pub use Foo` statements so is not needed here.

BOOM! The process worked. This statement is incorrect. I was trying to check to see if you'd thought about it or were just following what I'd said. Never follow exactly what I say, I talk a lot of rubbish.

The attribute is not _doing_ anything here, that is why we can remove it (its not on top of any line of code). The 'only goes on `pub use Foo`' bit is related to the code comment and the policy/style I/we have of abusing the attribute to keep the pub use statements separate. Said another way, it would be nice if rustfmt would format the pub use statements too but in doing so it merges them with the other non-pub use statements, and we don't like that so we block it all together. Note also this effects lines _after_ the marked code too sometimes (in relation to moving lines of code around).

jamillambert:
According to git blame the attribute was added at the same time as the code below and was possibly meant to apply to it, with the mistake being the empty line after it.  The commit message was trying to say that it was not needed, rather than it's placement was incorrect.  I am still not clear why it was ever put there.

I have changed it again to a hopefully more precise message.

tcharding (approved):
ACK 3d0cd56b7816cafced88c5cb2b6c3c71ee3aa199

tcharding:
> According to git blame the attribute was added at the same time as the code below and was possibly meant to apply to it, with the mistake being the empty line after it.

Yeah I assumed that pub statements got removed during dev and the attribute accidentally was left hanging there. But who knows. Either way I am at fault for not seeing it in review. All of this is why I wanted the  commit message to just say exactly what you said: 'for the following technical reason this line of code is wrong and can be removed' - we don't really need to know how it got there. That is my opinion anyway. Thanks man

tcharding:
Woops CI didn't run because of the new contributor rule and I forgot to hit 'run'. My bad.

tcharding:
I'll eat my hat if this patch is wrong in any way.

PR #49, opened 2025-01-23
Release tracking PR: `v0.6.0`

tcharding:
In preparation for release of all three crates add changelog entries, bump the version number, and update the lock files.

Patch 2 closes #69

PR #53, opened 2025-02-05
Include `node` in workspace

tcharding:
Currently we do not include the `node` crate in the workspace because `run_task` did not support testing it.

However `run_task` has been patched. Additionally we discovered that we were not _testing_ the docs with `--all-features` (or any code for that matter other than linting). Fixed in:

https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/pull/20

Which this patch uses.

Include `node` in the workspace. Configure it to work with the newly patched `run_task`. Add a `extra_tests.sh` script that runs the `node` tests.

In order to get MSRV building we make all deps use `default-features = false` then explicitly enable just what we need.

Also change the feature guarding in the build script so that `--all-features` works. This is already supported in the main crate code since the highest version feature overrides the lower ones.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -3,6 +3,18 @@
 version = 3
 
 [[package]]
+name = "adler2"
+version = "2.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "512761e0bb2578dd7380c6baaa0f4ce03e84f95e960231d1dec8bf4d7d6e2627"
+
+[[package]]
+name = "anyhow"
+version = "1.0.95"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "34ac096ce696dc2fcabef30516bb13c0a68a11d30131d3df6f04711467681b04"
+
+[[package]]
 name = "arrayvec"
 version = "0.7.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -92,16 +104,52 @@ dependencies = [
 ]
 
 [[package]]
+name = "bitflags"
+version = "2.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8f68f53c83ab957f72c32642f3868eec03eb974d1fb82e453128456482613d36"
+
+[[package]]
 name = "byteorder"
 version = "1.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
-name = "cc"
-version = "1.0.28"
+name = "bzip2"
+version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "bb4a8b715cb4597106ea87c7c84b2f1d452c7492033765df7f32651e66fcf749"
+checksum = "bdb116a6ef3f6c3698828873ad02c3014b3c85cadb88496095628e3ef1e347f8"
+dependencies = [
+ "bzip2-sys",
+ "libc",
+]
+
+[[package]]
+name = "bzip2-sys"
+version = "0.1.12+1.0.8"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "72ebc2f1a417f01e1da30ef264ee86ae31d2dcd2d603ea283d3c244a883ca2a9"
+dependencies = [
+ "cc",
+ "libc",
+ "pkg-config",
+]
+
+[[package]]
+name = "cc"
+version = "1.2.13"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c7777341816418c02e033934a09f20dc0ccaf65a5201ef8a450ae0105a573fda"
+dependencies = [
+ "shlex",
+]
+
+[[package]]
+name = "cfg-if"
+version = "1.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "baf1de4339761588bc0619e3cbc0120ee582ebb74b53b4efbf79117bd2da40fd"
 
 [[package]]
 name = "corepc-client"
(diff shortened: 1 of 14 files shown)


tcharding:
This is mainly to see what happens. I'll come back to debug this.

tcharding:
BOOM! Mad, no further improvements needed.

apoelstra (approved):
ACK 84ac230f1e154b421a50196d1114001cce105c0d; successfully ran local tests

apoelstra:
Lemme dig into this -- did not expect that local CI run to pass.

apoelstra:
Ah, it looks like you removed `node` from the `exclude` list but did not put it into `exclude`.

tcharding:
> did not expect that local CI run to pass.

Neither did I.

> it looks like you removed node from the exclude list but did not put it into members.

lol, its hard to get good help.

tcharding:
FTR what is the point of having _both_ an explicit 'include' and an 'exclude' list - surely one is default since something must be either in the group or out of the group.

tcharding:
Ah good, my local test command is now failing. 

> BOOM! Mad, no further improvements needed.

Face palm.

apoelstra:
> FTR what is the point of having _both_ an explicit 'include' and an 'exclude' list - surely one is default since something must be either in the group or out of the group.

I think that if you have a crate in neither list, and you try to build that crate directly, it'll complain that it sees a Cargo.toml in a higher directory and isn't sure what to do.

I don't know if this is a bug or intended behavior or what.

Anyway, looks like we're getting the CI failures that we expect now.

tcharding:
Now includes a custom lint script - phew.

tcharding:
I used the changes to `update-lock-files.sh` to update the lock files in #49.

Cargo-minimal.lock
```diff
@@ -197,12 +593,116 @@ dependencies = [
...
+checksum = "809e8dc61f6de73b46c85f4c96486310fe304c434cfa43669d7b40f711150908"
+dependencies = [
+ "aho-corasick",
+ "memchr",
+ "regex-syntax",
+]
+
+[[package]]
+name = "regex-syntax"
+version = "0.8.5"
```
apoelstra, line 636: In 17b94df416eabc5853b86b9aaf38a069b54daf8f:

regex-syntax 0.8.5 does not work with Rust 1.63. (Requires 1.65.)

tcharding: Damn, MSRV is not being checked currently. Back to the drawing board. Thanks

tcharding:
I played with this a bit today but its super boring. I rekon I'll merge #55 tomorrow and then merge this and cut the release. Lets punt this PR to the next release.

apoelstra:
Ok. I will just utACK stuff then.

tcharding:
After sleeping on this I think it is actually a big deal because we (I) claim that the `node` crate is MSRV 1.63 and if its not being tested the the claim is worthless.

tcharding:
@apoelstra please see the question at the bottom of the PR description before we consider this for merge.

tcharding:
hmm, `regext-syntax` is still version 8.5 in `Cargo-minimal.lock` yet the msrv job is green. Something must be wrong with this stil.

tcharding:
I got rid of `regex-syntax` all together using `default-features = false`. The I messed around reducing version numbers until I got MSRV to build. Then I used `cargo upgrade` to try and get a more recent lock file as well. Let's see what CI thinks.

tcharding:
Can you try your CI please @apoelstra?

tcharding:
Rebased on master to pick up #64.

tcharding:
Gentle bump please @apoelstra, I'm itching to release 0.6.0

apoelstra:
Sorry for the delay. It did take me a little bit of thought -- it's not as simple as "at least one feature". It has to be "at least one feature in the `node` workspace which matches the regex `^[0-9_]*$`".

Then it looks like in the root
```
cargo test --doc --all-features
```
fails, which is a real failure (the doctests are not working). I think that our CI should cover this..

tcharding:
Wild discovery bro, we were not actually running tests in `run_task` with `--all-features`. Sounds really bad but we were linting with `--all-features`. Still pretty bad mistake, I'll own that.

node/build.rs
```diff
@@ -1,13 +1,13 @@
 #[cfg(not(feature = "download"))]
 mod download {}
 
-#[cfg(any(not(feature = "download"), feature = "doc"))]
+#[cfg(all(not(feature = "download"), feature = "doc"))]
```
apoelstra, line 4: In 7bd1486e28a9d5d47b031e532001527951a28915:

Now there is no `main` function if neither `doc` nor `download` is provided.

apoelstra: I think my local CI would've found this -- I didn't check because the code was visibly wrong -- but it would be nice if Github could too.

In my local CI I have a complicated setup in which I test with no features, all features, every feature individually, and then two pairs of random features (generated from a hash of cargo.toml or something like that). And then after generating all those, I post-filter them and exclude any feature lists for which no feature matches `^[0-9_]*$`.

Nix is a pretty awful language but bash is even worse, so you might not want to try to reproduce that..

tcharding: Thanks man.

tcharding:
Now uses merge commit after merging https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/pull/20

node/build.rs
tcharding: I added two lines to the `extra_test.sh` script

```rust
    $cargo check --all-features
    $cargo check --doc
```

And a bunch more feature gate ugly hacks to get `compile_error!` to be the only output when using `--no-default-features`.

apoelstra:
Lol, nice, adding those redudant imports so the compiler output isn't such a mess without features.

apoelstra:
In 84c45ba5eda074a462c89fba6bfd81c8ed3e16f0:

This does not build if you enable `0_17_1 download doc` and nothing else.

It may help to draw out an ascii-art truth table in the file.

tcharding:
> This does not build if you enable 0_17_1 download doc and nothing else.

There is no reason to enable `doc` unless one is building the docs. But I can have another go, no sweat. Sorry this is dragging out so much.

tcharding:
I think its correct as it is man. We are abusing the `download` and `doc` features by using `not` on them so as to support `--all-features`. Said another way, we have to have at least one `not` statement on `main` otherwise it triggers for `--all-features` and for this we use `doc` and `download`.

apoelstra:
I was ok with saying "you need at least one of the version features" but not with this byzantine set of rules for which feature combinations are permissible. I can update my local CI but I haven't even gotten that far because manually understanding the feature gates takes so much work and I haven't completed it without encountering something surprising.

At the very least, if we don't want certain combinations to work then we should have an explicit `compile_error` about them.

tcharding:
Let me try solve this at a higher level. We have a bunch of smellyness in the features. `doc` is unusual and the feature ones are unusual in multiple ways (require at least one and enable all versions lower).

tcharding:
Hoorah!! We got there. I removed the `doc` feature all together. The fail was not noticing this when I introduced `rustdoc-args = ["--cfg", "docsrs"]` in the manifest ages ago.

tcharding:
Hopefull the docs build done by docs.rs does not run the tests during its build process. Because this crate's docs currently exist online I rekon this is true.

apoelstra:
I've almost got this. My local CI has a `cargo test --doc` line which has taken a bunch of iteration, since it doesn't work with `--all-features` (that enables `download` which doesn't work inside Nix) nor does it work with no features (that triggers the compile-fail error) and obtaining a specific feature list is a bit of a trick.

There are similar lines all over the place which cover things like doctests which crate2nix/cargo don't test unless you go out of your way to explicitly specify, and they all break in some way with this crate.

tcharding:
Holy smokes this one should not have been so hard should it. Here is another attempt. Please see https://github.com/rust-bitcoin/corepc/pull/45#issuecomment-2679758153 and also the additional patch second to last in this PR.

- Now `cargo test` should work and tests pass if you have `bitcoind` in your path and its version 17 (works for 18 and likely a few others too FWIW). Implies `cargo test --doc` works too.

- Features stack correctly in downstream crates (ie the footgun fixed in #45 is still gone)

I owe you a beer for this one.

apoelstra:
>Now cargo test should work and tests pass if you have bitcoind in your path and its version 17 

LOL ok, yeah, I'll go digging around to find a copy of boost 1.75 and miniupnpc 2.2.7 and gcc 11 and whatever libc I need to work with it.

(These changes don't actually make the situation any worse -- if it's true that I need to somehow obtain and run extremely old versions of Core to make the tests pass then that's going to be a problem for me no matter what.)

tcharding:
How about we not run `node` builds/tests in your CI and rely on GitHub for testing it?

apoelstra:
As long as `node` is in the `workspace.members` part of the manifest then I have to deal with the existence of the `download` feature.

If I can't build specific versions of Core then sure, I can just disable those.

tcharding:
I don't know what to do about this now man.

apoelstra:
I'm working on it.

tcharding:
No stress, no rush. FTR I'm not itching to release now, that itch went away.

tcharding:
Hey man, I'm going to go ahead and merge this. I need some time away from `rust-bitcoin` and this repo is the other thing I can work on without waiting on or arguing with other people. I'm totally drained right now. I just want to crawl off into a corner and write some code. Please don't take me pushing forward as any comment on you or your CI, both of which I'm happy to work with. Its the greater rust-bitcoin 'community' that has me zonked.

apoelstra:
No problem. I can test my local CI against master while I'm getting it working.

PR #55, opened 2025-02-10
Update docs and verify infrastructure

tcharding:
For Core versions 19 - 25 inclusive update the rustdocs in `types/src/vxy/mod.rs to match the style we introduced in v17 and v18 already. And add support to the `verify` tool to check the work.

This is all version upto the latest 3.


tcharding:
Ha! 4,000 line diff. This is unreviewable. I rekon you should just trust me that its all docs and tooling related.

tcharding:
I pulled #56 out of this, they were unrelated.

PR #56, opened 2025-02-10
Clean up tests

tcharding:
Do some trivial integration test clean ups.

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -87,7 +87,11 @@ pub fn get_address_info() {
 fn get_balance() {
     let node = Node::new_with_default_wallet();
     let json = node.client.get_balance().expect("getbalance");
-    assert!(json.into_model().is_ok())
+    assert!(json.into_model().is_ok());
+
+    node.fund_wallet();
+    let json = node.client.get_balance().expect("getbalance");
+    assert!(json.into_model().is_ok());
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


integration_test/tests/wallet.rs
```diff
@@ -74,13 +74,25 @@ pub fn get_addresses_by_label() {
 }
 
 #[test]
-// TODO: Consider testing a few different address types.
 #[cfg(feature = "TODO")]
+// TODO: We are getting an error for missing `pubkeys` field but the field is optional?
```
apoelstra, line 78: In 86dc357af3eb900e1e606776a3490a4bff3bb802:

You need `#[serde(default)]` to make an `Option<T>` field actually optional. Otherwise it is required to be present and allowed to be `null`.

integration_test/tests/wallet.rs
```diff
@@ -74,13 +74,25 @@ pub fn get_addresses_by_label() {
...
+
+    // node.fund_wallet();
+    // let _ = node
+    //     .client
+    //     .send_to_address(&address, Amount::from_sat(10_000))
+    //     .expect("sendtoaddress");
+    // node.mine_a_block();
+
+    // let json = node.client.get_address_info(&address).expect("getaddressinfo");
+    // assert!(json.into_model().is_ok());
```
apoelstra, line 95: In 86dc357af3eb900e1e606776a3490a4bff3bb802:

Did you intend to add so much commented-out code in a PR titled "clean up tests"? Did you mean to call it "make tests less clean"? ;)

integration_test/tests/wallet.rs
tcharding: Legend. So that means all the other `Option` fields must be getting `null` because this is the first time I had this problem and its definitely not the first usage of `Option`. TIL. Thanks.

apoelstra:
utACK 13c8e9f5d3e936703ace759d3b8ea89a4c18004b

PR #74, opened 2025-02-28
Update to base64 0.22.1

AaronDewes:
This is just simple dependency update to avoid people having duplicate (old) versions of base64 in their dependency tree if they use this crate.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -32,15 +32,15 @@ dependencies = [
 
 [[package]]
 name = "base64"
-version = "0.13.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


apoelstra:
I think this is fine, but can you update the included lockfiles `Cargo-minimal.lock` and `Cargo-recent.lock`?

You can run `just update-lock-files` to do this.

tcharding:
Verified that this is the same version we depend on in `bitcoin`.

tcharding:
I rebase on `master`, updated the lock files, and force pushed to your branch. Hope you don't mind. You can use `git range-diff aeaf57c...@` to see what changes I pushed and you can verify aeaf57c is your patch by looking at the original push above. You are still the patch author, and I don't know where github gets the data to include my avatar on the patch.

Thanks for the PR!! Appreciate it.

Issue #59, opened 2025-02-12
types: Are docs on the version specific types ok?


tcharding:
Currently the `types` crate uses re-exports so that for a specific Core version the type exists e.g., `v18::CreateWallet` is a re-export of `v17::CreateWallet` but `v25::CreateWallet` is not (because the JSON returned in Core v25 changed).

Note also that while writing the structs in `types` I used the docs from `bitcoin-cli` hitting a bitcoind instance for the version in question. This means that the docs on `v18::CreateWallet` are actually the docs returned by `bitcoind v17`. 

This would not be a problem if the docs didn't change meaningfully between Core versions without a code change as well but I do not know if that is the case.

GideonBature:
I think docs on version specifics where a method changed significantly is very okay. I feel reflecting that change brings about more clarity.

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/59

PR #89, opened 2025-03-10
Add missing transaction categories

tnull:
Contrary to Bitcoin Core's v17 docs, there are *five* transaction categories that might be returned as part of a response to `listtransactions`, `listsinceblock`, or `gettransaction`. While Core fixed this omission in the docs in https://github.com/bitcoin/bitcoin/pull/14653, we didn't account for them in the respective model's `enum`, leading to calls to `get_transaction` randomly failing with the rather obscure error message:
```
Err(JsonRpc(Json(Error("unknown variant `generate`, expected `send` or `receive`", line: 1, column: 682))))
```

Here, we fix this omission and add the missing categories. Note that while the doc fix https://github.com/bitcoin/bitcoin/pull/14653 was shipped as part of `v18`, the categories are pre-existing, which is why we simply fix the `v17` and `model` variants in-place here.

@tcharding This might warrant a patch release, IMO?

### DIFF
--- a/types/src/v17/wallet/mod.rs
+++ b/types/src/v17/wallet/mod.rs
@@ -38,10 +38,16 @@ pub enum AddressPurpose {
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "lowercase")]
 pub enum TransactionCategory {
-    /// Transaction is a send.
+    /// Transactions sent.
     Send,
-    /// Transactions is a receive.
+    /// Non-coinbase transactions received.
     Receive,
+    /// Coinbase transactions received with more than 100 confirmations.
+    Generate,
+    /// Coinbase transactions received with 100 or fewer confirmations.
+    Immature,
+    /// Orphaned coinbase transactions received.
+    Orphan,
 }
 
 /// Whether this transaction can be RBF'ed.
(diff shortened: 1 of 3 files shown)


tcharding:
This is broken in more than just the last release and also since the docs build is broken I rekon we just fix both and do a new release - and make the policy for now that only the latest version is supported. Ok with you?

types/src/v17/wallet/mod.rs
```diff
@@ -38,10 +38,16 @@ pub enum AddressPurpose {
 #[derive(Copy, Clone, PartialEq, Eq, Debug, Deserialize, Serialize)]
 #[serde(rename_all = "lowercase")]
 pub enum TransactionCategory {
-    /// Transaction is a send.
+    /// Transactions sent.
     Send,
-    /// Transactions is a receive.
+    /// Non-coinbase transactions received.
     Receive,
+    /// Coinbase transactions received with more than 100 confirmations.
```
tcharding, line 45: I think we should merge this but just flagging that now there are docs in `v17` that came from a later version of Core. We have an outstanding issue #59 that describes how the re-exports include docs so version specific types have the 'wrong' docs on them.

tcharding (approved):
ACK f31fe4687fdf0e6e5c798b58a93e6df2e7aad6db

tcharding:
> failing with the rather obscure error message:
  ```
  Err(JsonRpc(Json(Error("unknown variant `generate`, expected `send` or `receive`", line: 1, column: 682))))
  ```

Yeah this is always the kind of error message, its not great is it. That is the reson I want this crate _really_ well tested (its not currently at all) so downstream never sees this - that is the dream anyway.

tcharding:
Oh woops, we need a point release to fix the docs so I'll include this patch too.

tcharding:
Backported to 0.6.0 in #91

PR #90, opened 2025-03-10
node: Fix the rustdocs build

tcharding:
Currently the docs.rs build is not working. From the log it can be seen that the build is trying to download a bitcoind binary. I cannot see why this is happening, it does not happen locally. Anyway we can just change the docs build to use `28_0` instead of all features and this bypasses the problem.

While we are at it, change the feature gating in `build.rs` on the `download` module because it still seems wrong. Phew logically combining negative truth values is hard to reason about.

### DIFF
--- a/node/Cargo.toml
+++ b/node/Cargo.toml
@@ -65,5 +65,5 @@ download = ["anyhow", "bitcoin_hashes", "flate2", "tar", "minreq", "zip"]
 0_17_1 = []
 
 [package.metadata.docs.rs]
-all-features = true
+features = ["28_0"]
 rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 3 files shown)


tcharding:
Backported to 0.6.0 in #92

apoelstra:
>I cannot see why this is happening, it does not happen locally.

It happens for me. Whenever the `download` feature is enabled then build.rs attempts to contact the Internet. This is one reason that my local CI has been such a struggle to get working.

tcharding:
The problem is/was that `docsrs` is set using `RUSTDOCSFLAGS` so is not set in `build.rs` - so the `cfg(not(docsrs))` attribute in `build.rs` is meaningless.

PR #93, opened 2025-03-10
types: Update status for `getblockstats`

tcharding:
Recently we added tests for `getblockstats` but forgot to update the status.

There is still some confusion in my mind as what should be done about versions 19-24. Currently the method is untested and not working for these versions but we still re-export the `GetBlockStats` type and the method exists in `bitcoincli help getblockstats`.


tcharding:
cc @tnull in case you have any suggestions.

tcharding:
OMG I did not realize that - that is a super easy fix then. Thanks man.

Issue #97, opened 2025-03-12
`client` does not turn on `types` `std` feature


tcharding:
The `client` crate re-exports `types` so some users will not use it directly. This is currently broken because `std` crates will be missing `std::error::Error` impls in `types` because `std` is not enabled.

```rust
[dependencies]
bitcoin = { version = "0.32.0", default-features = false, features = ["std", "serde"] }
types = { package = "corepc-types", version = "0.6.1", default-features = false, features = ["std"] }
```

Bug found while trying to upgrade `electrsd` to use the `0.6.x` series.

PR #98, opened 2025-03-13
client: Enable `std` feature in `types` crate

tcharding:
The `client` is std only but we don't enable the `std` feature in `types` - this is a bug.

Fix: #97


tcharding:
Needs backport to `0.6.x` branch and release

PR #106, opened 2025-03-18
Implement methods from the mining section

tcharding:
Add types, implement in client, and verify all methods from the mining section except `submitheader` and `submitblock`. `submitblock` is done but I can't get the test to pass all the time.

- `submitblock` has flaky (excluded) tests.
- `submitheader` is untested.

One the way I fixed a few things in the network section (`getpeerinfo`) because I ran into issues there at some stage.


tcharding:
Debugging `submitblock` is killing me!

PR #108, opened 2025-03-19
Fix `{create,load}wallet` on `v25`

tnull:
With `v25` the `warning` field was deprecated for `warnings`. While it should only be available when the `-deprecatedrpc=walletwarningfield` is passed, I was hitting parse errors even though I did not find where I
passed this argument. Here, we add the `warning` field to the `v25` types (although we ignore it as it's literally duplicative of what's in `warnings` already) to avoid the parse error.

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -2,8 +2,6 @@
 
 //! Tests for methods found under the `== Wallet ==` section of the API docs.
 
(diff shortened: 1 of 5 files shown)


tcharding (approved):
ACK 99a1cda6998bc2476e6eaefc271578ec29eec862

tcharding:
Nice! Thanks man. I did #110 motivated by this PR.

PR #111, opened 2025-03-26
Retry initial client connections

tnull:
I'm not sure what of the behavior really changed since the `bitcoind` days, but here's yet another thing I stumbled across while trying to making the switch to `corepc-node`:

Previously, `corerpc-node` would try to establish a connection via the given `Auth` cookie, and would panic if it couldn't setup a initial connection. Only the rest of the initial setup steps would be retried with time-delay.

This however was race-y, as the auth cookie might not already be synced to disk on first startup of a fresh `bitcoind` instance (with fresh data dir), when we try to connect for the first time. Here, we move the intial connection logic into the retry loop to simply also retry it instead of panicking.

@tcharding Seems this might be the last issue (fingers crossed) that keeps LDK and LDK Node from switching our CI over to `corepc-node`. As the switch is also blocking some features, it would be much appreciated if this could be released soonish after it's reviewed.

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -383,45 +383,61 @@ impl Node {
         thread::sleep(Duration::from_millis(1000));
         assert!(process.stderr.is_none());
 
-        let mut i = 0;
+        let mut tries = 0;
         let auth = Auth::CookieFile(cookie_file.clone());
-        let client_base =
-            Client::new_with_auth(&rpc_url, auth.clone()).expect("failed to create client");
 
         let client = loop {
+            tries += 1;
+
+            if tries > 10 {
+                error!("failed to get a response from bitcoind");
+                return Err(Error::NoBitcoindInstance.into());
+            }
+
+            let client_base = match Client::new_with_auth(&rpc_url, auth.clone()) {
+                Ok(client) => client,
+                Err(e) => {
+                    error!("failed to create client: {}. Retrying!", e);
+                    thread::sleep(Duration::from_millis(1000));
+                    continue;
+                }
+            };
+
             // Just use serde value because changes to the GetBlockchainInfo type make debugging hard.
             let client_result: Result<serde_json::Value, _> =
                 client_base.call("getblockchaininfo", &[]);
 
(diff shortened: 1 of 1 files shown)


nervana21:
tACK https://github.com/rust-bitcoin/corepc/pull/111/commits/26ceb4d7a67ad7404228418dc006d09df2ab2518

Thanks @tnull! 

PR tested and reviewed by cloning main ([2b410f2](https://github.com/rust-bitcoin/corepc/commit/2b410f23e251baa747832153e06c7a978bf5f1cd)), then applying two patches:


[failing-tests.patch](https://github.com/user-attachments/files/19561420/failing-tests.patch) – adds two tests that fail on main.
[pr-111.patch](https://github.com/user-attachments/files/19561421/pr-111.patch) – fixes those failures by adding retry logic implemented in this PR.


Below is the reproduction process (requires `sleep` and `bitcoind` in PATH on Unix-like systems):

```bash
# 1) Clone the repo and switch to main
git clone https://github.com/rust-bitcoin/corepc.git
cd corepc
git checkout main

# 2) Apply the failing-tests patch (two tests that fail on main)
git apply failing-tests.patch

# 3) Run tests -- these new tests should fail before fix is applied
cargo test --lib test_with_conf_ -- --test-threads=1

# 4) Apply pr-111.patch (the retry logic fix)
git apply pr-111.patch

# 5) Run tests again -- now they pass, after the fix has been applied
cargo test --lib test_with_conf_ -- --test-threads=1

tcharding:
Mad, thanks man. You'll have to give me a minute please. I'll get to this soon as I can.

tcharding:
utACK 26ceb4d7a67ad7404228418dc006d09df2ab2518

tcharding:
> As the switch is also blocking some features, it would be much appreciated if this could be released soonish after it's reviewed.

Sure can. I've got a bunch of patches locally that I might push up first but I'll try and release tomorrow for you. No guarantees though.

tcharding:
Tagged and published.

PR #116, opened 2025-04-05
Implement version specific types

### LINKED ISSUE #115: Implement version specific types for status 'omitted' methods
Currently we have a status 'omitted' that was meant to imply that we intentionally omitted adding a type. For example if method returns a std lib type

```
$ bt17 help pruneblockchain
pruneblockchain

Arguments:
1. "height"       (numeric, required) The block height to prune up to. May be set to a discrete height, or a unix timestamp
                  to prune blocks whose block time is at least 2 hours older than the provided timestamp.

Result:
n    (numeric) Height of the last block pruned.
```

This has the problem that later versions of the method might change and we will not catch it.

It would be better to add a version specific type, implement in the client, and an integration test that calls the method. Even for `null`, which is currently does as follows: 

```rust
                match self.call("submitblock", &[into_json(hex)?]) {
                    Ok(serde_json::Value::Null) => Ok(()),
                    Ok(res) => Err(Error::Returned(res.to_string())),
                    Err(err) => Err(err.into()),
                }
```

Note that doing this also requires updating the method in `verify` (e.g., `verify::metho::v17`) to use `Method::new_no_model`.

GideonBature:
## Implement Version-Specific Types for Status `omitted`

This PR addresses **[issue #115 ](https://github.com/bitcoin/rust-bitcoincore-rpc/issues/115)** by implementing the methods with status of `'omitted'` in methods with version-specific types. Previously, `'omitted'` was used in cases where return types were standard or seemingly inconsequential (e.g., `null`, standard library types). However, this approach introduced a risk: **if the return type of such methods changes in future Bitcoin Core versions, we would not catch it.**

To fix this, this PR:
- Defines **version-specific types** for affected methods.
- Updates method definitions to use `Method::new_no_model` in corresponding methods.
- Adds **integration tests** for each updated method to verify correctness and catch future regressions.

This improves type safety, clarity, and future-proofing of the client interface.

---

#### 🚀 Blockchain Methods  
Tested with:  
```bash
cargo test --features=28_0 --test blockchain
```

| S/N | Method Name     | Implemented | Test Written |
|-----|------------------|-------------|---------------|
| 1   | pruneblockchain  | ✅           | ✅             |
| 2   | savemempool      | ✅           | ✅             |
| 3   | scantxoutset     | ✅           | ✅             |
| 4   | verifychain      | ✅           | ✅             |

---

#### 🌐 Network Methods  
Tested with:  
```bash
cargo test --features=28_0 --test network
```

| S/N | Method Name        | Implemented | Test Written |
|-----|---------------------|-------------|---------------|
| 5   | addnode             | ✅           | ✅             |
| 6   | clearbanned         | ✅           | ✅             |
| 7   | disconnectnode      | ✅           | ✅             |
| 8   | getconnectioncount  | ✅           | ✅             |
| 9   | listbanned          | ✅           | ✅             |
| 10  | ping                | ✅           | ✅             |
| 11  | setban              | ✅           | ✅             |
| 12  | setnetworkactive    | ✅           | ✅             |

---

#### 💰 Wallet Methods
Tested with:  
```bash
cargo test --features=28_0 --test wallet
```

| S/N | Method Name               | Implemented | Test Written |
|-----|----------------------------|-------------|---------------|
| 13  | abandontransaction         | ✅           | ✅             |
| 14  | abortrescan                | ✅           | ✅             |
| 15  | backupwallet               | ✅           | ✅             |
| 16  | encryptwallet              | ✅           | ✅             |
| 17  | importaddress              | ✅           | ✅             |
| 18  | importmulti                | ✅           | ✅             |
| 19  | importprivkey              | ✅           | ✅             |
| 20  | importprunedfunds          | ✅           | ✅             |
| 21  | importpubkey               | ✅           | ✅             |
| 22  | importwallet               | ✅           | ✅             |
| 23  | keypoolrefill              | ✅           | ✅             |
| 24  | lockunspent                | ✅           | ✅             |
| 25  | removeprunedfunds          | ✅           | ✅             |
| 26  | sethdseed                  | ✅           | ✅             |
| 27  | settxfee                   | ✅           | ✅             |
| 28  | walletlock                 | ✅           | ✅             |
| 29  | walletpassphrase           | ✅           | ✅             |
| 30  | walletpassphrasechange     | ✅           | ✅             |

---

✅ Ready for review

Closes #115


client/src/client_sync/v17/blockchain.rs
```diff
@@ -297,3 +297,16 @@ macro_rules! impl_client_v17__verifytxoutproof {
...
 }
+
+/// Implements Bitcoin Core JSON-RPC API method `pruneblockchain`
+#[macro_export]
+macro_rules! impl_client_v17__pruneblockchain {
+    () => {
+        impl Client {
+            /// Instructs the node to prune the blockchain up to a specified height or timestamp.
+            pub fn prune_blockchain(&self, target: u64) -> Result<PruneBlockchain> {
+                self.call("pruneblockchain", &[into_json(target)?])
```
tcharding, line 308: FYI for std types we can just call `into()`. `into_json` is for using `serde` serialization on `rust-bitcoin` types.

```suggestion
                self.call("pruneblockchain", &[target.into())
```

integration_test/tests/blockchain.rs
```diff
@@ -255,3 +255,30 @@ fn verify_tx_out_proof() {
     let txids = node.client.verify_tx_out_proof(&proof).expect("verifytxoutproof");
     assert_eq!(txids.0.len(), 1);
 }
+
+// Implicitly tests the omitted method `pruneblockchain`
```
tcharding, line 259: Why 'Implicitly'? Also no need to comment tests, the test name is correct - its the name of the RPC method in snakcase.

integration_test/tests/blockchain.rs
```diff
@@ -255,3 +255,30 @@ fn verify_tx_out_proof() {
...
+#[test]
+fn prune_blockchain() {
+    let node = Node::with_wallet(Wallet::Default, &["-prune=550"]);
+    // Generate 1001 blocks to exceed regtest's nPruneAfterHeight (1000)
+    let num_blocks_to_mine = 1001;
+    // (Using a loop with mine_a_block is necessary as generate() is deprecated in v28)
```
tcharding, line 265: You can use `generatetoaddress` instead.

node/src/versions.rs
```diff
@@ -1,7 +1,7 @@
 // An explicit version of Bitcoin Core must be selected by enabling some feature.
 // We check this here instead of in `lib.rs` because this file is included in `build.rs`.
 #[cfg(all(
-    not(feature = "28_0"),
+    feature = "28_0",
```
tcharding, line 4: This looks suspicious. What are you trying to achive?

types/src/model/blockchain.rs
```diff
@@ -486,3 +486,7 @@ pub struct GetTxOutSetInfo {
 /// Models the result of JSON-RPC method `verifytxoutproof`.
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct VerifyTxOutProof(pub Vec<Txid>);
+
+/// Models the result of JSON-RPC method `pruneblockchain`.
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Hash, Deserialize, Serialize)]
```
tcharding, line 491: Better to use the same derives as other stucts in the file.

types/src/v17/blockchain/into.rs
```diff
@@ -545,3 +545,13 @@ impl VerifyTxOutProof {
         Ok(model::VerifyTxOutProof(proofs))
     }
 }
+
+impl PruneBlockchain {
+    /// Converts version specific type to a version nonspecific, more strongly typed type.
+    pub fn into_model(self) -> Result<model::PruneBlockchain, PruneBlockchainError> {
```
tcharding, line 551: We don't really need to provide this. The only types that get a struct in `model` are types that use strongly type fields (ie types from `rust-bitcoin`). Since this is just an int we don't need to model it.

types/src/v17/blockchain/mod.rs
```diff
@@ -667,3 +667,20 @@ pub struct GetTxOutSetInfo {
...
+/// Result of JSON-RPC method `pruneblockchain`.
+///
+/// > pruneblockchain height
+/// >
+/// > Arguments:
+/// > 1. "height"       (numeric, required) The block height to prune up to. May be set to a discrete height, or a unix timestamp
+/// >                   to prune blocks whose block time is at least 2 hours older than the provided timestamp.
+/// >
+/// > Result:
+/// > n    (numeric) Height of the last block pruned.
```
tcharding, line 681: Just to be uniform with the rest of the crate can we leave the `Result: ` bit out.

verify/rpc-api-v17.txt
```diff
@@ -18,7 +18,7 @@ gettxout "txid" n ( include_mempool )
 gettxoutproof ["txid",...] ( blockhash )
 gettxoutsetinfo
 preciousblock "blockhash"
-pruneblockchain
+pruneblockchain height
```
tcharding, line 21: These file was created using `bt17 help > rpc-api-v17.txt`. It should not need updating.

verify/src/method/v17.rs
```diff
@@ -29,11 +29,12 @@ pub const METHODS: &[Method] = &[
     Method::new_string("gettxoutproof", "get_tx_out_proof"),
     Method::new_modeled("gettxoutsetinfo", "GetTxOutSetInfo", "get_tx_out_set_info"),
     Method::new_nothing("preciousblock", "precious_block"),
-    Method::new_numeric("pruneblockchain", "prune_blockchain"),
     Method::new_nothing("savemempool", "save_mempool"),
     Method::new_modeled("scantxoutset", "ScanTxOutSet", "scan_tx_out_set"),
     Method::new_bool("verifychain", "verify_chain"),
     Method::new_modeled("verifytxoutproof", "VerifyTxOutProof", "verify_tx_out_proof"),
+    Method::new_no_model("pruneblockchain", "PruneBlockchain", "prune_blockchain"),
+
```
tcharding, line 37: This is the correct usage of `Method` (since we don't need a model type). But please don't move it to a different place, these are in the same order as in `rpc-api-v17.txt`.

tcharding (commented):
Looking good mate. I wouldn't personally bother pruning at the moment. At this stage just checking that the returned data is the correct shape is enough.

client/src/client_sync/v17/blockchain.rs
GideonBature: True, thank you! I have resolved it.

integration_test/tests/blockchain.rs
GideonBature: True, wanted to follow the last test that was added thinking it's a standard for the codebase. I will remove it.

integration_test/tests/blockchain.rs
GideonBature: Thank you, although I observed that `generatetoaddress` only works for v17 - v22, from v23 upwards, it's not available for use, that was why I took this approach instead.  But I can still go ahead and implement the `generatetoaddress` if it's okay to ignore the other versions for now.

types/src/model/blockchain.rs
GideonBature: Adjusted... Thank you....

types/src/v17/blockchain/into.rs
GideonBature: Noted! Thank you...

types/src/v17/blockchain/mod.rs
GideonBature: Noted!

verify/src/method/v17.rs
GideonBature: Noted! Thank you so much for the reviews....

GideonBature:
I have effected the changes reviewed, please do have a check when you are chanced.

types/src/model/blockchain.rs
```diff
@@ -486,3 +486,7 @@ pub struct GetTxOutSetInfo {
 /// Models the result of JSON-RPC method `verifytxoutproof`.
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct VerifyTxOutProof(pub Vec<Txid>);
+
+/// Models the result of JSON-RPC method `pruneblockchain`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct PruneBlockchain(pub u32);
```
tcharding, line 492: We don't need this.

types/src/model/mod.rs
```diff
@@ -27,7 +27,7 @@ pub use self::{
         GetBlockchainInfo, GetChainTips, GetChainTxStats, GetDifficulty, GetMempoolAncestors,
         GetMempoolAncestorsVerbose, GetMempoolDescendants, GetMempoolDescendantsVerbose,
         GetMempoolEntry, GetMempoolInfo, GetRawMempool, GetRawMempoolVerbose, GetTxOut,
-        GetTxOutSetInfo, MempoolEntry, MempoolEntryFees, Softfork, SoftforkType, VerifyTxOutProof,
+        GetTxOutSetInfo, MempoolEntry, MempoolEntryFees, Softfork, SoftforkType, VerifyTxOutProof, PruneBlockchain,
```
tcharding, line 30: Or this.

types/src/v17/blockchain/error.rs
```diff
@@ -560,3 +560,37 @@ impl std::error::Error for GetTxOutSetInfoError {
 impl From<NumericError> for GetTxOutSetInfoError {
     fn from(e: NumericError) -> Self { Self::Numeric(e) }
 }
+
+/// Error when converting the `PruneBlockchain` JSON type to the Model type.
+#[derive(Debug)]
+pub enum PruneBlockchainError {
```
tcharding, line 566: Don't need any of this since there is no need for conversion to model type.

types/src/v17/blockchain/mod.rs
```diff
@@ -667,3 +667,17 @@ pub struct GetTxOutSetInfo {
 /// Inner field is the txid(s) which the proof commits to, or empty array if the proof can not be validated.
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct VerifyTxOutProof(pub Vec<String>);
+
+
```
tcharding, line 671: nit: Only one line of whitespace :)

verify/rpc-api-v17.txt
```diff
@@ -85,7 +84,7 @@ abandontransaction "txid"
 abortrescan
 addmultisigaddress nrequired ["key",...] ( "label" "address_type" )
 backupwallet "destination"
-bumpfee "txid" ( options ) 
+bumpfee "txid" ( options )
```
tcharding, line 87: This shouldn't be in here. I'm guessing you forgot to read the diff before force pushing. No stress I do that all the time.

tcharding (commented):
Getting there man, keep at it.

verify/rpc-api-v17.txt
GideonBature: True, hehehe. Thank you so much for the reviews...

types/src/v17/blockchain/mod.rs
GideonBature: :) removed...

types/src/v17/blockchain/error.rs
GideonBature: True, thank you.

types/src/model/mod.rs
GideonBature: Thank you!

types/src/model/blockchain.rs
GideonBature: True 🫣

GideonBature:
Changes effected... Thank you so much for all the reviews, I am really learning and understanding the codebase more.

tcharding:
I just introduced a new style of writing test code in #125. Please follow the same style when you write your new tests. Thanks man!

tcharding:
You got a rebase mistake in there mate. In `verify/src/metho/`

tcharding:
In case its useful to you I have this alias

```bash
which mc
mc: aliased to git grep -l '<<<<'
```

Then I use `mc` to check for merge conflicts after rebasing.

tcharding:
This is going to be hard for me to review because its so big but also I appreciate that it is hard to work on this crate without doing big changesets so I'm happy to review this but there may be a few iterations.

tcharding:
Some things I found:

- Need to set the status to `done` e.g. `types/src/v17/mod.rs` when each method is added and tested
- If you could use the new test format I'd appreciate it, otherwise I have to go over all the tests and add it

```rust
    let json: VerifyTxOutProof = node.client.verify_tx_out_proof(&proof)?;
    let model: Result<mtype::VerifyTxOutProof, _> = json.into_model();
    let txids = model.unwrap();
```

The point to this is to check that all the re-exports were done correctly.

Or using `PruneBlockchain`

```rust
    let _: PruneBlockchain = node.client.prune_blockchain(target_height);
```

It would probably be best to push up an initial patch that does just one method completely so I can review that and then you could do all the others as a second patch. Sorry I should have said that weeks ago, my bad. There is a way to stage code chunks so you can pull the changes out - let me know if you want more tips on that.

GideonBature:
> You got a rebase mistake in there mate. In `verify/src/method/`

Noted, thank you for spotting it out.

GideonBature:
> This is going to be hard for me to review because its so big but also I appreciate that it is hard to work on this crate without doing big changesets so I'm happy to review this but there may be a few iterations.

No Problem, I will make all changes necessary as you review! :).

GideonBature:
> Some things I found:
> 
> * Need to set the status to `done` e.g. `types/src/v17/mod.rs` when each method is added and tested
> * If you could use the new test format I'd appreciate it, otherwise I have to go over all the tests and add it
> 
> ```rust
>     let json: VerifyTxOutProof = node.client.verify_tx_out_proof(&proof)?;
>     let model: Result<mtype::VerifyTxOutProof, _> = json.into_model();
>     let txids = model.unwrap();
> ```
> 
> The point to this is to check that all the re-exports were done correctly.
> 
> Or using `PruneBlockchain`
> 
> ```rust
>     let _: PruneBlockchain = node.client.prune_blockchain(target_height);
> ```
> 
> It would probably be best to push up an initial patch that does just one method completely so I can review that and then you could do all the others as a second patch. Sorry I should have said that weeks ago, my bad. There is a way to stage code chunks so you can pull the changes out - let me know if you want more tips on that.

True, almost forgot changing their status to done.

Okayyy, now I see the reason why, for this particular methods, most of them returns `null` or just a `standard type` so I will implement for the others just like the example you provided. Thank you so much for the reviews.

I will also push up an initial patch with single method so you can review, before doing the others. I will look it up i.e. stage code chucks, and if there is anything I don't understand, I will definitely ask for clarity.

Once again thank you always :).

tcharding:
Closing as stale.

PR #133, opened 2025-04-26
Add check for new_no_model methods with model types

### LINKED ISSUE #70: verify: Check no model
If a method is marked as `new_no_model` (in `verify::method`) then we should check that a type does not exist in `model`. This is catch changes over versions when a type now requires a model type but we forgot to use `new_modelled` in `method`.

GideonBature:
This PR adds a verification step to ensure that methods marked as `new_no_model` don't have corresponding model types. If a model type is found for such methods, the verification fails with a clear error message suggesting to change the method marking to `new_modelled` instead.

### Changes

- Added a new verification function `verify_no_model_methods` that checks for this inconsistency
- Integrated this check into the main verification flow
- Added a helper function `type_physically_exists` to check if a model type exists regardless of whether it's required
- Fixed an inconsistency found during testing of the newly added check: changed `getblockfilter` for v19 from `new_no_model` to `new_modelled` as it actually has a model type

This helps prevent inconsistencies between method type annotations and the actual model implementation. When methods evolve across versions and begin to require model types, this check will catch cases where we forgot to update the method annotation from `new_no_model` to `new_modelled`.

Closes #70

### DIFF
--- a/verify/src/main.rs
+++ b/verify/src/main.rs
@@ -83,6 +83,14 @@ fn verify_version(version: Version, test_output: Option<&String>, quiet: bool) -
     verify_status(version, test_output)?;
     close(correct, quiet);
 
+    let msg = "Verifying no model types for no_model methods";
+    check(msg, quiet);
+    let correct = verify_no_model_methods(version)?;
+    close(correct, quiet);
+    if !correct {
+        process::exit(1);
+    }
+
     Ok(())
 }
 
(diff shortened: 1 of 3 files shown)


tcharding:
Mad, you found a bug in `verify`. We can fix it in a much more simple way however. This diff fixes the bug but does not fix the new warnings from `verify` - want to use this?

```diff
diff --git a/verify/src/main.rs b/verify/src/main.rs
index 5ab4de8..8015ae3 100644
--- a/verify/src/main.rs
+++ b/verify/src/main.rs
@@ -166,2 +166,6 @@ fn check_types_exist_if_required(version: Version, method_name: &str) -> Result<
         }
+    } else {
+        if model::type_exists(version, method_name)? {
+            eprintln!("found model type when non expected: {}", output_method(out));
+        }
     }
diff --git a/verify/src/model.rs b/verify/src/model.rs
index fc753b2..7080ef3 100644
--- a/verify/src/model.rs
+++ b/verify/src/model.rs
@@ -42,5 +42,3 @@ pub fn type_exists(version: Version, method_name: &str) -> Result<bool> {
     if let Some(Return::Type(s)) = method.ret {
-        if method.requires_model {
-            return crate::grep_for_string(&path(), s);
-        }
+        return crate::grep_for_re_export(&path(), s);
     }
```

GideonBature:
> Mad, you found a bug in `verify`. We can fix it in a much more simple way however. This diff fixes the bug but does not fix the new warnings from `verify` - want to use this?
> 
> ```diff
> diff --git a/verify/src/main.rs b/verify/src/main.rs
> index 5ab4de8..8015ae3 100644
> --- a/verify/src/main.rs
> +++ b/verify/src/main.rs
> @@ -166,2 +166,6 @@ fn check_types_exist_if_required(version: Version, method_name: &str) -> Result<
>          }
> +    } else {
> +        if model::type_exists(version, method_name)? {
> +            eprintln!("found model type when non expected: {}", output_method(out));
> +        }
>      }
> diff --git a/verify/src/model.rs b/verify/src/model.rs
> index fc753b2..7080ef3 100644
> --- a/verify/src/model.rs
> +++ b/verify/src/model.rs
> @@ -42,5 +42,3 @@ pub fn type_exists(version: Version, method_name: &str) -> Result<bool> {
>      if let Some(Return::Type(s)) = method.ret {
> -        if method.requires_model {
> -            return crate::grep_for_string(&path(), s);
> -        }
> +        return crate::grep_for_re_export(&path(), s);
>      }
> ```

Thank you for this :).

tcharding:
What is the status of this PR  mate?

GideonBature:
> What is the status of this PR mate?

I have used the changes you recommend, tested it and it's working quite well.

tcharding:
I expected to see only the changes I suggested and not all the original code you wrote. Was there some reason I'm missing that you kept all the additional code?

tcharding:
I went back to double check that my comment above was correct and in order to do so I had to re-code my solution. So I just pushed #139.

PR #135, opened 2025-04-28
Fix version for 27.2 feature flag

GideonBature:
When running with the "27_2" feature flag enabled, the VERSION constant was incorrectly set to "27.1". This PR fixes the version string to properly reflect "27.2" when the corresponding feature flag is active.


tcharding (approved):
ACK 698d931b39f4688e81a50236c05dc42ad693c6d5

PR #139, opened 2025-04-30
verify: Check that no model type exists

tcharding:
If a method does not require a model we currently do not check this. Add logic to do so.

Fix the warnings found:

- Re-export `ListUnspent`
- Use the correct `Method` constructor in v19

Fix: #70


tcharding:
This PR replaces #133

PR #143, opened 2025-05-01
chore(deps): bump zip in light of RUSTSEC-2020-0071

storopoli:
There's a RUSTSEC advisory on time that should be updated to >=0.2.23. It gets pulled down from corepc-node with download feature from the zip dependency, see the cargo tree below:

```
time 0.1.45
└── zip 0.5.13
    └── corepc-node 0.7.0
```

Hence we need to update the zip dependency.

I don't know how to generate the `Cargo-*.lock` files.
Can you please advise?

Refs: https://rustsec.org/advisories/RUSTSEC-2020-0071

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -200,6 +200,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "crossbeam-utils"
+version = "0.8.21"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "d0a5c400df2834b80a4c3327b3aad3a4c4cd4de0629063962b03235697506a28"
+
+[[package]]
 name = "env_logger"
 version = "0.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


tcharding (approved):
ACK 4d30a0c50714050e9455a11bb3d743d8917414ae

tcharding:
Just need to update the lock files please bro.

storopoli:
@tcharding done!
I had to go over some MSRV stuff to find a `zip` version that is unaffected by the RUSTSEC and supports 1.63.0 MSRV.

storopoli:
We should tag a new version `0.8.0`.
Shall we also yank previous versions from crates.io due to this RUSTSEC advisory?
NOTE: this only affects `corepc-node` with the `download` feature flag.

tcharding:
We can just release `corepc-node 0.7.1`. No need for any yanking I don't believe. IIUC that is exactly what the patch version is for.

tcharding (approved):
ACK c8a8aa9582ced41e6f799d05bfe37c171e8e89f0

PR #152, opened 2025-05-08
Do a few trivial cleanups 

tcharding:
Follow up from #148 to save doing another round of review/dev.


tcharding:
cc @GideonBature, I just did these changes instead of asking you to do them because they are trivial. Also I should have noticed the version alias thing in patch 2 straight away in review but I did not.

PR #79, opened 2025-03-06
Expose remaining `v17` `blockchain` APIs for `v19`+

tnull:
While these API methods were previously exposed for a `v17` client, they were not for `v19`+.

Here, we add the `blockchain` APIs present in `v17` for `v19`+.

PR #153, opened 2025-05-08
client: Remove duplicate macros

tcharding:
The client method macros do not need to be duplicated in different versions because our imports correctly work with the produced code meaning the methods return the correct version specific type.


tcharding:
This fixes mistakes that I missed during review of #79. cc @GideonBature, no wonder you were confused about the client.

GideonBature:
> This fixes mistakes that I missed during review of #79. cc @GideonBature, no wonder you were confused about the client.

Oh! hahaha.... At least it's fixed now.

PR #154, opened 2025-05-09
Implement getnodeaddresses method and test

jamillambert:
Implement the `getnodeaddresses` method added in V18, including integration test and reexports in newer versions.

### DIFF
--- a/client/src/client_sync/v18/network.rs
+++ b/client/src/client_sync/v18/network.rs
@@ -1,0 +1,22 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Macros for implementing JSON-RPC methods on a client.
+//!
+//! Requires `Client` to be in scope.
+//!
+//! Specifically this is methods found under the `== Network ==` section of the
+//! API docs of Bitcoin Core `v0.18`.
+//!
+//! See, or use the `define_jsonrpc_minreq_client!` macro to define a `Client`.
+
+/// Implements Bitcoin Core JSON-RPC API method `getnodeaddresses`
+#[macro_export]
+macro_rules! impl_client_v18__getnodeaddresses {
+    () => {
+        impl Client {
+            pub fn get_node_addresses(&self) -> Result<GetNodeAddresses> {
+                self.call("getnodeaddresses", &[])
+            }
+        }
+    };
+}
--- a/types/src/v18/network/mod.rs
+++ b/types/src/v18/network/mod.rs
@@ -1,0 +1,28 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! The JSON-RPC API for Bitcoin Core `v0.18` - network.
+//!
+//! Types for methods found under the `== Network ==` section of the API docs.
+
+use serde::{Deserialize, Serialize};
+
+/// Result of JSON-RPC method `getnodeaddresses`.
+///
+/// > getnodeaddresses ( count )
+/// >
+/// > Return known addresses which can potentially be used to find new nodes in the network.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct GetNodeAddresses(pub Vec<NodeAddress>);
+
+/// An item from the list returned by the JSON-RPC method `getnodeaddresses`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct NodeAddress {
+    /// Timestamp in seconds since epoch (Jan 1 1970 GMT) when the node was last seen.
+    pub time: u64,
+    /// The services offered.
+    pub services: u64,
+    /// The address of the node.
+    pub address: String,
+    /// The port of the node.
+    pub port: u16,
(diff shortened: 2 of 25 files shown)


client/src/client_sync/v18/network.rs
```diff
@@ -0,0 +1,26 @@
...
+//! API docs of Bitcoin Core `v0.18`.
+//!
+//! See, or use the `define_jsonrpc_minreq_client!` macro to define a `Client`.
+
+/// Implements Bitcoin Core JSON-RPC API method `getnodeaddresses`
+#[macro_export]
+macro_rules! impl_client_v18__getnodeaddresses {
+    () => {
+        impl Client {
+            pub fn get_node_addresses(&self, count: Option<u32>) -> Result<GetNodeAddresses> {
```
tcharding, line 17: While correct, just in general I have been excluding any optional args unless they are needed for integration testing. And later in the integration test you call this with `None`.

types/src/v18/network/mod.rs
```diff
@@ -0,0 +1,35 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! The JSON-RPC API for Bitcoin Core `v0.18` - network.
+//!
+//! Types for methods found under the `== Network ==` section of the API docs.
+//!
+/// These types do not implement `into_model` because apart from fee rate there is no additional
+/// `rust-bitcoin` types needed.
+use serde::{Deserialize, Serialize};
```
tcharding, line 9: nit: newline please.
```suggestion

use serde::{Deserialize, Serialize};
```

types/src/v18/network/mod.rs
```diff
@@ -0,0 +1,35 @@
...
+/// Result of JSON-RPC method `getnodeaddresses`.
+///
+/// > getnodeaddresses ( count )
+/// >
+/// > Return known addresses which can potentially be used to find new nodes in the network.
+/// >
+/// > Arguments:
+/// > 1. count    (numeric, optional, default=1) How many addresses to return. Limited to the smaller of 2500 or 23% of all known addresses.
+/// >
+/// > Result: Array of objects with time, services, address, and port.
```
tcharding, line 20:
```suggestion
/// Result of JSON-RPC method `getnodeaddresses`.
///
/// > getnodeaddresses ( count )
/// >
/// > Return known addresses which can potentially be used to find new nodes in the network.
```

tcharding (commented):
Mad, too easy! I would probably squash all the commits except the formatting one. Thanks man.

client/src/client_sync/v18/network.rs
jamillambert: I think I addressed this correctly, and do now remember you saying to leave out optional args, but it got lost in all the other info.

types/src/v18/network/mod.rs
jamillambert: Done

jamillambert: This failed format CI with the space. I copied the file from v17 which has no newline after, but the commenting changes from `//!` to `///`. I have now changed this file to use the module comments, which cargo format does want a new line after.

tcharding: Oh I did not notice the weird comment `///` thing. I've done #158, I'll merge that then you can rebase on master and exclude the comment all together here as well.

jamillambert:
Rebased on #158 and removed same comment for v18.

tcharding (approved):
ACK 4c4d8ba31eae4b05f38b180beeb9f5f05e412e91

types/src/v18/network/mod.rs
```diff
@@ -0,0 +1,28 @@
...
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct GetNodeAddresses(pub Vec<NodeAddress>);
+
+/// An item from the list returned by the JSON-RPC method `getnodeaddresses`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct NodeAddress {
+    /// Timestamp in seconds since epoch (Jan 1 1970 GMT) when the node was last seen.
+    pub time: u64,
+    /// The services offered.
+    pub services: u64,
```
tcharding, line 23: There as an issue when working with Core and numeric types that Core sometimes returs -1 signalling an error (because its C++ code). But from the RPC docs one never knows which numeric types can do this. Then to complicate the matter one only hits this issue 'in the wild' when an error is actually returned. The solution I've come up with (actually this was suggested to me by Steven who wrote the crate rust-bitcoincore-rpc that this crate is hoping to improve upon) is to use signed integers anytime its not obvious. So in this case I have not clue from reading the docs on v17 what `services` is so I would have used an i64. This solution works very nicely for fields that have a rust-bitcoin type that is  a u32, see `to_u32` in `types/src/lib.rs` and the `NumericError` type.

However I only came up with this solution after having done a bunch of this repo so this is not at all uniform.

PR #155, opened 2025-05-09
Implement verifychain method and test

GideonBature:
This is based on PR https://github.com/rust-bitcoin/corepc/pull/116 on implementing one method at a time for easier review: This is the third method implementation: `verifychain` which returns a (bool type). Once this is approved, I’ll proceed with the next one.

### DIFF
--- a/types/src/v17/blockchain/mod.rs
+++ b/types/src/v17/blockchain/mod.rs
@@ -650,6 +650,10 @@ pub struct PruneBlockchain(
     pub i64,
 );
 
+/// Result of JSON-RPC method `verifychain`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct VerifyChain(pub bool);
+
 /// Result of JSON-RPC method `verifytxoutproof`.
 ///
 /// > verifytxoutproof "proof"
--- a/client/src/client_sync/v17/blockchain.rs
+++ b/client/src/client_sync/v17/blockchain.rs
@@ -318,6 +318,16 @@ macro_rules! impl_client_v17__savemempool {
     };
 }
 
+/// Implements Bitcoin Core JSON-RPC API method `verifychain`
+#[macro_export]
+macro_rules! impl_client_v17__verifychain {
+    () => {
+        impl Client {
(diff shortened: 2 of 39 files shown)


types/src/v17/blockchain/mod.rs
```diff
@@ -650,6 +650,22 @@ pub struct PruneBlockchain(
...
+/// >
+/// > Arguments:
+/// > 1. checklevel (numeric, optional, default=3, range=0-4) How thorough the block verification is:
+/// - level 0 reads the blocks from disk
+/// - level 1 verifies block validity
+/// - level 2 verifies undo data
+/// - level 3 checks disconnection of tip blocks
+/// - level 4 tries to reconnect the blocks
+/// - each level includes the checks of the previous levels
+/// > 2. nblocks (numeric, optional, default=6, 0=all) The number of blocks to check.
```
tcharding, line 665:
```suggestion
/// Result of JSON-RPC method `verifychain`.
```

I've not been perfectly uniform but lets try to not include docs for optional args. Just because its too noisy and even more of a problem with going stale between versions.

So for these 'thin' methods perhaps just the one line rustdoc is good enough?

tcharding (commented):
Except the docs thing everything here looks good - too easy!

types/src/v17/blockchain/mod.rs
GideonBature: True, let me do this...

GideonBature:
> Except the docs thing everything here looks good - too easy!

removed the unnecessary lines of the rust docs, leaving just a single line, as you suggested....

tcharding:
While reviewing this I saw that we have the integration test for `gettxoutproof` in the wrong place (it shows up in the diff of this patch). Want to fix that? Either after this merges or as a separate patch at the front of this?

FTR I'm as fallible as the next bloke, if you see mistakes please clean em up - its super appreciated. This repo is even more necessary to keep clean than most because its so big, boring, boilerplate.

tcharding (approved):
ACK ae286f1d70bcedcf4a3486764e00d518be09893f

tcharding:
Thanks man, you are killing it!

GideonBature:
> Thanks man, you are killing it!

Thank you.... I am learning from you...

PR #151, opened 2025-05-06
Move TemplateRequest and TemplateRules into their proper module

GideonBature:
This pull request moves the `TemplateRequest` struct and `TemplateRules` enum, into their appropriate module to improve code organization and maintainability.

PR #156, opened 2025-05-11
client: Move types to version specific module

tcharding:
Currently we have a bit of confusion around types that are defined and used by the various client version modules.

- Some currently don't change e.g., `Input`
- Some are due to change in v29 e.g., `TemplateRequest`
- Some change already e.g. `AddressVersion`

And to make matters worse, I did 'custom' re-exports in `node` which is very confusing when one is reading code in `client`.

Put all `client` defined types in the version module they first appear in and re-export them as we do for types from `types`. Then use the type naked (without qualifying the path).

### DIFF
--- a/client/src/client_sync/mod.rs
+++ b/client/src/client_sync/mod.rs
@@ -16,13 +16,9 @@ pub mod v26;
 pub mod v27;
 pub mod v28;
 
-use std::collections::HashMap;
 use std::fs::File;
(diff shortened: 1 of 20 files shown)


tcharding:
I noticed this confusion while reviewing #151. I take all responsibility for it, both `AddressType` and the new code introduced in #151.

tcharding:
I'll come back after the weekend.

Issue #147, opened 2025-05-05
node: Build hangs if `download` feature is enabled


tcharding:
Something is wrong with the build lately but I've no clue what. It may be being obfuscated (or caused) by unrelated network issues I'm experiencing with my network.

tcharding:
My hunch is it is to do with `--all-features` and chanegs in `node v0.7.0` to how features are handled.

I manually updated the lockfiles for #146 using `cargo check`

storopoli:
I've seen flaky downloads from `https://bitcoincore.org/bin/` URLs in my corepc-node builds

tcharding:
Oh is that right, good to know. Thanks man. That might also explain the flakey CI jobs I've been seeing lately.

tcharding:
Perhaps we should cache the Core binaries so we don't download them ever CI run.

tcharding:
Now I'm seeing the build hang in `integration_test` not in `node` and not when downloading binaries.

```bash
BITCOIND_EXE=/opt/bitcoin-0.17.2/bin/bitcoind cargo test --features=0_17_2
```

Is hanging - so the problem is with us.

tcharding:
#193 is a quick bandaid for local dev.

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/147

PR #159, opened 2025-05-12
Document local integration test strategy

tcharding:
Our current setup supports running the integration tests using a local Core binary but the way to do so may not be obvious. So much so that I myself have been downloading in each worktree.

Add a README to document how to speed up local dev.


tcharding:
Done while investigating #147

PR #163, opened 2025-05-12
Implement all v17 util functions

jamillambert:
Add all of the `Util` methods from core v17, macros for implementation, required models and tests. Update the table in types updated for v17.

Run cargo fmt to correctly order the new method names in the lists.


types/src/model/util.rs
```diff
@@ -4,3 +4,35 @@
...
 //! These structs model the types returned by the JSON-RPC API but have concrete types
 //! and are not specific to a specific version of Bitcoin Core.
+
+use serde::{Deserialize, Serialize};
+
+/// Models the result of JSON-RPC method `createmultisig`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct CreateMultisig {
+    /// The value of the new multisig address.
+    pub address: String,
```
tcharding, line 14: Should be `Address`

types/src/model/util.rs
```diff
@@ -4,3 +4,35 @@
...
+    /// The string value of the hex-encoded redemption script.
+    #[serde(rename = "redeemScript")]
+    pub redeem_script: String,
```
tcharding, line 17: Should be `ScriptBuf`

types/src/model/util.rs
```diff
@@ -4,3 +4,35 @@
...
+}
+
+/// Models the result of JSON-RPC method `validateaddress`.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ValidateAddress {
+    /// If the address is valid or not. If not, this is the only property returned.
+    pub isvalid: bool,
```
tcharding, line 24:
```suggestion
    pub is_valid: bool,
```

types/src/model/util.rs
```diff
@@ -4,3 +4,35 @@
...
+    /// The bitcoin address validated.
+    pub address: Option<String>,
+    /// The hex encoded scriptPubKey generated by the address.
+    #[serde(rename = "scriptPubKey")]
```
tcharding, line 28: We don't need these in `model` because we control the field names. These serde renames are because Core controls the field names when doing the RPC calls but the model types are never used directly against Core.

types/src/model/util.rs
```diff
@@ -4,3 +4,35 @@
...
+    pub script_pub_key: Option<String>,
+    /// If the key is a script.
+    pub isscript: Option<bool>,
+    /// If the address is a witness address.
+    pub iswitness: Option<bool>,
+    /// The version number of the witness program.
+    pub witness_version: Option<u8>,
+    /// The hex value of the witness program.
+    pub witness_program: Option<String>,
+}
```
tcharding, line 38: Same for all other fields, should be types from `rust-bitcoin` and use Rust idiomatic field names.

integration_test/tests/util.rs
```diff
@@ -0,0 +1,93 @@
...
+#![allow(non_snake_case)] // Test names intentionally use double underscore.
+#![allow(unused_imports)] // Because of feature gated tests.
+
+use bitcoin::hex::FromHex as _;
+use bitcoin::{Amount};
+use integration_test::{Node, NodeExt as _, Wallet};
+use node::vtype::*; // All the version specific types.
+
+#[test]
+fn utils__createmultisig__modelled() {
```
tcharding, line 14:
```suggestion
fn utils__create_multisig__modelled() {
```

I tried to use Rust idiomatic names (excluding the double underscore) instead of Core names (all squashed together). I tried to find a balance of readable identifiers and uniformity with Core. I guess I messed up by naming all the `client` macros how they are - in hindsighte they should have underscores in them.

integration_test/tests/util.rs
```diff
@@ -0,0 +1,93 @@
...
+    let addr1 = node.client.new_address().expect("new_address 1");
+    let addr2 = node.client.new_address().expect("new_address 2");
+    let info1 = node.client.validate_address(&addr1).expect("validateaddress 1");
+    let info2 = node.client.validate_address(&addr2).expect("validateaddress 2");
+
+    let pubkey1 = info1.address.as_ref().unwrap();
+    let pubkey2 = info2.address.as_ref().unwrap();
+
+    // For test purposes, just use the addresses as dummy pubkeys (in real use, use actual pubkeys).
+    let keys = vec![pubkey1.clone(), pubkey2.clone()];
```
tcharding, line 28: Once `validateaddress` has strong types in `model` then we can pass `PublicKey`s in here.

integration_test/tests/util.rs
```diff
@@ -0,0 +1,93 @@
...
+
+    let json: CreateMultisig = node.client.create_multisig(2, &keys, Some("legacy")).expect("createmultisig");
+    let model: Result<mtype::CreateMultisig, _> = serde_json::to_value(&json)
+        .and_then(|v| serde_json::from_value(v));
```
tcharding, line 32: I'm not sure what the intention is here. We need `into_model` to be implemented for `CreateMultisig` then I think my confusion will resolve.

integration_test/tests/util.rs
```diff
@@ -0,0 +1,93 @@
...
+    model.expect("CreateMultisig into model");
+}
+
+#[test]
+fn utils__estimatesmartfee__modelled() {
```
tcharding, line 37: Here you have the method marked as modellled which is correct. But in `verify/src/method/v17.rs` it is marked as returns nothing (my bug not yours). For all methods when removeing 'TODO' we have to check `verify` is correct because I made tons of mistakes there. Once you are done `verify` should pass.

types/src/v17/mod.rs
```diff
@@ -137,11 +137,11 @@
...
 //! | JSON-PRC Method Name               | Returns         | Notes                                  |
 //! |:-----------------------------------|:---------------:|:--------------------------------------:|
-//! | createmultisig                     | version + model | TODO                                   |
-//! | estimatesmartfee                   | returns nothing | TODO                                   |
-//! | signmessagewithprivkey             | returns string  | TODO                                   |
-//! | validateaddress                    | version + model | TODO                                   |
-//! | verifymessage                      | returns boolean | TODO                                   |
+//! | createmultisig                     | version + model |                                        |
+//! | estimatesmartfee                   | returns nothing |                                        |
+//! | signmessagewithprivkey             | returns string  |                                        |
```
tcharding, line 142: Should be modelled - using `ecdsa::Signature` (I think).

types/src/v17/mod.rs
```diff
@@ -137,11 +137,11 @@
...
+//! | validateaddress                    | version + model |                                        |
+//! | verifymessage                      | returns boolean |                                        |
```
tcharding, line 144: You've added a type for this so this should be 'version'. (Adding the type is correct and the most thorough solution)

types/src/v17/util.rs
```diff
@@ -3,3 +3,120 @@
...
+/// >                    higher feerate and is more likely to be sufficient for the desired
+/// >                    target, but is not as responsive to short term drops in the
+/// >                    prevailing fee market.  Must be one of:
+/// >        "UNSET" (defaults to CONSERVATIVE)
+/// >        "ECONOMICAL"
+/// >        "CONSERVATIVE"
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct EstimateSmartFee {
+    /// Estimate fee rate in BTC/kB.
+    pub feerate: Option<f64>,
```
tcharding, line 55:
```suggestion
    #[serde(rename = "feerate")]
    pub fee_rate: Option<f64>,
```

types/src/v17/util.rs
```diff
@@ -3,3 +3,120 @@
...
+    /// Errors encountered during processing.
+    pub errors: Option<Vec<String>>,
+    /// Block number where estimate was found.
+    pub blocks: Option<u64>,
```
tcharding, line 59: Why `Option`? I checked for v17 and v28 and didn't see 'optional'.

types/src/v17/util.rs
```diff
@@ -3,3 +3,120 @@
...
+/// > 1. "address"                    (string, required) The bitcoin address to validate
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ValidateAddress {
+    /// If the address is valid or not. If not, this is the only property returned.
+    pub isvalid: bool,
+    /// The bitcoin address validated.
+    pub address: Option<String>,
+    /// The hex encoded scriptPubKey generated by the address.
+    #[serde(rename = "scriptPubKey")]
+    pub script_pub_key: Option<String>,
```
tcharding, line 97:
```suggestion
    pub script_pubkey: Option<String>,
```

tcharding (commented):
EDIT: I didn't know exactly what type of review/help you were after.

Here's a few comments man, good effort. I struggled reviewing without feeling like my comments were too savage so take them as I intended, just pointing out the things that are needed. This repo is convoluted AF so don't stress about getting it all straight away. This PR is great.

types/src/model/util.rs
tcharding: Sometimes we need `serde(with ...` which is used by the serde derives to know how to de/serialize.

In explanation. The sered traits are implemented for `model` types so that users of the lib can store (eg `model::GetTransaction` in a data base or send it over the wire. However these types are never meant to be sent to Core. Since the types in e.g., `types/src/v17/foo.rs` are received over the wire from Core the stuff in there is to support that use case.

jamillambert (commented):
Thanks for all the review comments. It was even more than I was looking for. I'll let you know when it is ready for review again.

jamillambert:
I've made a lot of changes, but have not finished the tests yet. Just wanted some feedback, in particular if I have done the model and into_model parts right.  I was not sure what type to use for the smart fee, other functions use `SignedAmount`, `Amount` and `FeeRate`. I went with `FeeRate`. Also their into methods varied a lot, I used the one from `Network` `GetNetworkInfo`.

types/src/v17/util/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+    pub script_pubkey: String,
+    /// If the key is a script.
+    pub is_script: bool,
+    /// If the address is a witness address.
+    pub is_witness: bool,
+    /// The version number of the witness program.
+    pub witness_version: Option<i64>,
+    /// The hex value of the witness program.
+    pub witness_program: Option<String>,
+}
```
tcharding, line 113: The three predicates will need to use `rename` e.g, `isvalid`.

types/src/v17/util/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+/// >                    target, but is not as responsive to short term drops in the
+/// >                    prevailing fee market.  Must be one of:
+/// >        "UNSET" (defaults to CONSERVATIVE)
+/// >        "ECONOMICAL"
+/// >        "CONSERVATIVE"
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct EstimateSmartFee {
+    /// Estimate fee rate in BTC/kB.
+    #[serde(rename = "feerate")]
+    pub smart_fee: f64,
```
tcharding, line 62: Looks like this is optional.

types/src/v17/util/error.rs
```diff
@@ -0,0 +1,122 @@
...
+            E::RedeemScript(ref e) => Some(e),
+        }
+    }
+}
+
+/// Error when converting an `EstimateSmartFee` type into the model type.
+#[derive(Debug)]
+pub enum EstimateSmartFeeError {
+    /// Conversion of the `smart_fee` field failed.
+    SmartFee(amount::ParseAmountError),
```
tcharding, line 45: While  not incorrect, since there is only one error I'd just return the `ParseAmountError` directly. Like you do with `SignMessageWithPrivKey`.

tcharding (commented):
Yeah mate this looks good. R

jamillambert:
I have left `signmessagewithprivkey` and `verifymessage` untested. I had issues with the signature returned by the RPC call. The similar method `Wallet::signmessage` is also untested.

tcharding:
> This is not reexported at the bitcoin create root, should it be?

I don't believe so, we are trying to be a bit conservative with what is re-exported at the crate root.

types/src/v17/mod.rs
tcharding: This is not fixed yet mate. We need to propogate all these table changes upto v28 also. The `verify` tool is really letting us down here, it should have a way to catch this.

verify/src/method/v17.rs
```diff
@@ -81,8 +81,8 @@ pub const METHODS: &[Method] = &[
     Method::new_modelled("testmempoolaccept", "TestMempoolAccept", "test_mempool_accept"),
     // util
     Method::new_modelled("createmultisig", "CreateMultisig", "create_multisig"),
-    Method::new_nothing("estimatesmartfee", "estimate_smart_fee"),
-    Method::new_string("signmessagewithprivkey", "sign_message_with_priv_key"),
+    Method::new_modelled("estimatesmartfee", "EstimateSmartFee", "estimate_smart_fee"),
+    Method::new_modelled("signmessagewithprivkey", "SignMessageWithPrivKey", "sign_message_with_priv_key"),
     Method::new_modelled("validateaddress", "ValidateAddress", "validate_address"),
     Method::new_bool("verifymessage", "verify_message"),
```
tcharding, line 87:
```suggestion
    Method::new_no_model("verifymessage", "VerifyMessage", "verify_message"),
```

integration_test/tests/util.rs
```diff
@@ -0,0 +1,86 @@
...
+    let pubkey2 = "02fe6f0a5a297eb38c391581c4413e084773ea23954d93f7753db7dc0adc188b2f"
+        .parse::<PublicKey>()
+        .unwrap();
+
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    let json: CreateMultisig = node
+        .client
+        .create_multisig(nrequired, vec![pubkey1, pubkey2])
+        .expect("createmultisig");
+    let res: Result<mtype::CreateMultisig, _> = json.into_model();
```
tcharding, line 29:
```suggestion
    let res: Result<mtype::CreateMultisig, CreateMultisigError> = json.into_model();
```

Use `_` if the error is a rust-bitcoin error and use the explicit error if it is from `types` - that proves we re-exported the error correctly.

integration_test/tests/util.rs
```diff
@@ -0,0 +1,86 @@
...
+}
+
+#[test]
+fn util__validate_address__modelled() {
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    node.fund_wallet();
+
+    let addr = node.client.new_address().expect("new_address");
+    let json: ValidateAddress = node.client.validate_address(&addr).expect("validateaddress");
+    let res: Result<mtype::ValidateAddress, _> = json.into_model();
```
tcharding, line 80:
```suggestion
    let res: Result<mtype::ValidateAddress, ValidateAddressError> = json.into_model();
```

integration_test/tests/util.rs
```diff
@@ -0,0 +1,86 @@
...
+    let _ = res.expect("ValidateAddress into model");
+}
+
+// This is tested in util__sign_message_with_priv_key__modelled()
+#[test]
+fn util__verify_message() {}
```
tcharding, line 86: Well done! For others, this enables `verify` to succeed when checking the tests with the `--tests` flag.

tcharding (commented):
Looking good man, few more minor things and we are there.

PR #164, opened 2025-05-12
Implement scantxoutset method and test

GideonBature:
Going by the conversations with tcharding for PR https://github.com/rust-bitcoin/corepc/pull/116 on implementing one method at a time for easier review: This is the fourth method implementation `scantxoutset`. Once this is approved, I’ll proceed with the next one.

### DIFF
--- a/integration_test/tests/blockchain.rs
+++ b/integration_test/tests/blockchain.rs
@@ -308,6 +308,35 @@ fn blockchain__savemempool() {
 }
 
 #[test]
+fn blockchain__scan_tx_out_set_modelled() {
+    #[cfg(any(
+        feature = "v17",
+        feature = "v18",
+        feature = "v19",
+        feature = "v20",
+        feature = "v21"
+    ))]
+    let node = Node::with_wallet(Wallet::None, &[]);
+
+    #[cfg(not(any(
+        feature = "v17",
+        feature = "v18",
+        feature = "v19",
+        feature = "v20",
+        feature = "v21"
+    )))]
+    let node = Node::with_wallet(Wallet::None, &["-coinstatsindex=1"]);
+
+    let dummy_pubkey_hex = "0279be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798";
+    let scan_desc = format!("pkh({})", dummy_pubkey_hex);
+
+    let scan_objects: node::ScanObject = node::ScanObject::Descriptor(scan_desc);
+    let action: node::ScanAction = node::ScanAction::Start;
+
+    let _: Result<ScanTxOutSet, _> = node.client.scan_tx_out_set(action, &[scan_objects]);
+}
+
+#[test]
 fn blockchain__verify_tx_out_proof__modelled() {
     let node = Node::with_wallet(Wallet::Default, &[]);
     node.fund_wallet();
--- a/types/src/v17/blockchain/mod.rs
+++ b/types/src/v17/blockchain/mod.rs
@@ -650,6 +650,53 @@ pub struct PruneBlockchain(
     pub i64,
 );
 
+/// Result of JSON-RPC method `scantxoutset`.
+///
+/// > scantxoutset "action" ( [scanobjects,...] )
+/// >
+/// > Arguments:
+/// > 1. "action"                       (string, required) The action to execute
+/// > 2. "scanobjects"                  (array, required) Array of scan objects
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+#[serde(untagged)]
+pub enum ScanTxOutSet {
+    /// Returns after scan completes
+    Start(ScanTxOutSetStart),
+    /// True (scan will be aborted), False (no scan to abort)
+    Abort(bool),
+    /// Scan in progress or Completed
+    Status(Option<ScanTxOutSetStatus>),
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ScanTxOutSetStart {
+    /// The unspents
+    pub unspents: Vec<ScanTxOutSetUnspent>,
+    /// The total amount of all found unspent outputs in BTC
+    pub total_amount: f64,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct ScanTxOutSetStatus {
+    /// Approximate percent complete
+    pub progress: f64,
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ScanTxOutSetUnspent {
+    /// The transaction id
+    pub txid: String,
+    /// The vout value
+    pub vout: u32,
+    /// The script key
+    #[serde(rename = "scriptPubKey")]
+    pub script_pubkey: String,
+    /// The total amount in BTC of unspent output
+    pub amount: f64,
+    /// Height of the unspent transaction output
+    pub height: u64,
+}
+
 /// Result of JSON-RPC method `verifychain`.
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
 pub struct VerifyChain(pub bool);
--- a/client/src/client_sync/v17/blockchain.rs
+++ b/client/src/client_sync/v17/blockchain.rs
@@ -318,6 +318,30 @@ macro_rules! impl_client_v17__savemempool {
     };
 }
 
+/// Implements Bitcoin Core JSON-RPC API method `scantxoutset`
+#[macro_export]
+macro_rules! impl_client_v17__scantxoutset {
+    () => {
+        impl Client {
+            pub fn scan_tx_out_set(
+                &self,
+                action: ScanAction,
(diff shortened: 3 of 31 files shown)


tcharding:
Why is there identical code in the v17 and v18 `blockchain` modules of the client?

GideonBature:
> Why is there identical code in the v17 and v18 `blockchain` modules of the client?

They basically work the same way, the difference is in their return type, v18 has an additional field (`desc`) in it's result, whereas it is lacking in v17. Or is there a way to reuse the client code in v17 for v18 and still be able to return the type that is meant for v18?

tcharding:
Bro I explained this to you already on the other two PRs you did. Perhaps go back over them and see if you can work it out.

GideonBature:
> Bro I explained this to you already on the other two PRs you did. Perhaps go back over them and see if you can work it out.

oh, I now understand, I was looking at how `gettxout` was used in the client, where despite the code is similar, they used it for v17 and again for v22. So I thought that for any change in the fields of return type of a method, there should also be a code in the module of the client for that particular version with the change. But I can see that it's not always the case.

tcharding:
Oh well that would explain your confusion, the `gettxout` in v22 is just plain wrong. I'll remove it.

tcharding:
Done in #169. FTR bugs happen, if you see things that are wrong or look weird just ask (or PR to fix them). Apologies for being so fierce yesterday.

tcharding:
Returns column needs updating.

integration_test/tests/blockchain.rs
```diff
@@ -307,6 +307,35 @@ fn blockchain__savemempool() {
     }
 }
 
+#[test]
+fn blockchain__scantxoutset_modelled() {
```
tcharding, line 311:
```suggestion
fn blockchain__scan_tx_out_set__modelled() {
```

types/src/v17/blockchain/mod.rs
```diff
@@ -650,6 +650,53 @@ pub struct PruneBlockchain(
...
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ScanTxOutSetUnspent {
+    /// The transaction id
+    pub txid: String,
+    /// The vout value
+    pub vout: u32,
+    /// The script key
+    #[serde(rename = "scriptPubKey")]
+    pub script_pub_key: String,
```
tcharding, line 693:
```suggestion
    pub script_pubkey: String,
```

types/src/v17/blockchain/mod.rs
```diff
@@ -650,6 +650,53 @@ pub struct PruneBlockchain(
...
+/// Result of JSON-RPC method `scantxoutset`.
+///
+/// > scantxoutset "action" ( [scanobjects,...] )
+/// >
+/// > Arguments:
+/// > 1. "action"                       (string, required) The action to execute
+/// > 2. "scanobjects"                  (array, required) Array of scan objects
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+#[serde(untagged)]
+pub enum ScanTxOutSet {
```
tcharding, line 662: Oh this is interesting. This is a new idea for the crate, it could be done for all the methods that have verbose results. I'm not sure it adds any value though. I'd just remove this and leave the three individual return types. Plus I'd document 'returned when the foo action as taken'.

tcharding (commented):
Holy moly you picked a difficult one. I'm going to have to come back to this when I can put more thought into it. Feel free to ignore my suggestions until then.

types/src/v17/blockchain/mod.rs
tcharding: after looking at v28 I'm not sure - leave it with me for a bit.

GideonBature:
> Done in #169. FTR bugs happen, if you see things that are wrong or look weird just ask (or PR to fix them). Apologies for being so fierce yesterday.

No problem. I will do this going forward.

GideonBature:
> Returns column needs updating (and Notes).

Noted!

GideonBature:
> Holy moly you picked a difficult one. I'm going to have to come back to this when I can put more thought into it. Feel free to ignore my suggestions until then.

Noted....

types/src/v17/blockchain/mod.rs
GideonBature: Alright... Noted. That would be great using it for the methods having verbose results.

GideonBature:
> Returns column needs updating (and Notes).

for the notes, is there any need to change it? since it describes perfectly that the scantxoutset is experimental.

tcharding:
Oh you are right, lets leave the notes as they are.

GideonBature:
> Man this method is really not that simple to add support for because it changes so often. Can we put this one on hold until you've got a few more easy ones done?

Understood... sure, let's do that.

GideonBature:
Complete the `scantxoutset` method and model.

types/src/model/blockchain.rs
```diff
@@ -621,3 +621,57 @@ pub struct ReceiveActivity {
...
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ScanTxOutSetUnspent {
+    /// The transaction id
+    pub txid: Txid,
+    /// The vout value
+    pub vout: u32,
+    /// The script key
+    #[serde(rename = "scriptPubKey")]
+    pub script_pubkey: ScriptBuf,
```
tcharding, line 664: `model` types don't get renames because they are never used against Core.

types/src/model/blockchain.rs
```diff
@@ -621,3 +621,57 @@ pub struct ReceiveActivity {
...
+    /// The hash of the block at the tip of the chain. For v19 onwards.
+    pub bestblock: Option<BlockHash>,
+    /// The unspents
+    pub unspents: Vec<ScanTxOutSetUnspent>,
+    /// The total amount of all found unspent outputs in BTC
+    pub total_amount: Amount,
+}
+
+#[derive(Deserialize, Debug, Clone, PartialEq)]
+pub struct ScanTxOutSetStatus {
```
tcharding, line 651: Observation: This struct never  changes

Use `git grep -A 4 'pub struct ScanTxOutSetStatus'` to see.

types/src/v18/blockchain/mod.rs
```diff
@@ -70,3 +70,52 @@ pub struct MempoolEntry {
...
+    /// Returns after scan completes
+    Start(ScanTxOutSetStart),
+    /// True (scan will be aborted), False (no scan to abort)
+    Abort(bool),
+    /// Scan in progress or Completed
+    Status(Option<ScanTxOutSetStatus>),
+}
+
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ScanTxOutSetStart {
```
tcharding, line 93: This looks wrong. Perhaps add `#[serde(deny_unknown_fields)]` to all your structs and that will help find the bugs.

tcharding (commented):
You've done a bunch of work, well done. This one is pretty curly, here are some ideas.

I think it would be more simple if we were to split up the return types for the three different commands (start, abort, status). That way we only have to re-implement the ones that change.

In the client I think it would be more simple to have three methods, one for each command and not have a `ScanAction` type.

Instead of `ScanObject` I think we can just use a `String` and ignore the added complexity of an object (the `WithDesc` variant) since it is unused. In general we don't want to add code to the client that we do not use in testing.

That should give you a bit to work on!

jamillambert (commented):
There are some missing renames, relates to issue #298. I haven't necessarily highlighted all of them.

types/src/model/blockchain.rs
GideonBature: Noted!

GideonBature:
Implemented suggested changes.

types/src/v18/blockchain/mod.rs
jamillambert: Yeah I added it and tested, there are two missing fields, `success` and `searched_items`.

types/src/v25/blockchain/mod.rs
```diff
@@ -115,3 +117,374 @@ pub struct GetBlockStats {
...
 }
+
+/// Result of JSON-RPC method `getblockchaininfo`.
+///
+/// > getblockchaininfo
+/// >
+/// > Returns an object containing various state info regarding blockchain processing.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct GetBlockchainInfo {
```
jamillambert, line 128: Most of this is not part of this PR, left over from a copy and paste from v19?

jamillambert (commented):
I made a few comments, they apply to all affected versions.

GideonBature:
> I had a quick look at this and there have been so many changes to master in the meantime the merge conflicts are going to be a pain. I would suggest dropping all exports/imports from the commit fixing the remaining merge conflicts and then adding them back manually afterwards.

Noted!

jamillambert:
Continued in #428

PR #46, opened 2025-01-23
Enable running multiple Core v28 nodes

tcharding:
Currently the `node` crate cannot run multiple instances of Bitcoin Core v28 because of changes to bind logic in that release.

The solution is to use `-bind` instead of `-port`.

For more information on the relevant changes in Core v28 see:

https://github.com/bitcoin/bitcoin/blob/master/doc/release-notes/release-notes-28.0.md#p2p-and-network-changes

This patch was created with help of MaxFangX - mad props.

Fix: #41

PR #175, opened 2025-05-15
run-bitcoind: Fix startup of Core nodes

tcharding:
Core v28 introduced changes making node startup fail if bind fails. We fixed this in `node` in #46 but in `run-bitcoind.sh` we still use `-port`. This didn't cause any problems because we only pass the arg to one of the two nodes so port 18445 was available. When we try to add support for v29 things break.

Jamil was experiencing this problem (I think) while running v28 nodes but I don't know why.

Use `-bind` instead of `-port`.

### DIFF
--- a/contrib/run-bitcoind.sh
+++ b/contrib/run-bitcoind.sh
@@ -186,7 +186,7 @@ run_bitcoind() {
     echo "Starting bitcoind v${version_number} (alias: ${version})..."
     "$bitcoind" -regtest $fallback_fee_arg $block_filter_arg \
                 -datadir="${test_dir}/1" \
-                -port="${version_id}48" \
+                -bind="127.0.0.1:${version_id}48" \
                 -server=0 \
                 -printtoconsole=0 &
 
--- a/node/contrib/extra_tests.sh
+++ b/node/contrib/extra_tests.sh
@@ -32,7 +32,7 @@ main() {
 
     # But only run tests for the latest version. This is ok because we are mainly checking
     # MSRV and docs with this script. Integration test will check every Core version.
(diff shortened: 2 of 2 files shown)


tcharding:
cc @jamillambert I knew I was hitting this but I forgot it was when testing the v29 PR not when trying to run v28. I'm not sure why you are hitting it on v28 or even if that is what you are hitting - lets see.

jamillambert (commented):
There is a release note about it in 28.1 I found after letting you know about the issue: 
>When the -port configuration option is used, the default onion listening port will now be derived to be that port + 1 instead of being set to a fixed value (8334 on mainnet). This re-allows setups with multiple local nodes using different -port and not using -bind, which would lead to a startup failure in v28.0 due to a port collision.
>
>Note that a HiddenServicePort manually configured in torrc may need adjustment if used in connection with the -port option. For example, if you are using -port=5555 with a non-standard value and not using -bind=...=onion, previously Bitcoin Core would listen for incoming Tor connections on 127.0.0.1:8334. Now it would listen on 127.0.0.1:5556 (-port plus one). If you configured the hidden service manually in torrc now you have to change it from HiddenServicePort 8333 127.0.0.1:8334 to HiddenServicePort 8333 127.0.0.1:5556, or configure bitcoind with -bind=127.0.0.1:8334=onion to get the previous behavior. ([#31223](https://github.com/bitcoin/bitcoin/pull/31223))

tcharding:
ooph, that reads like the issue shouldn't have existed in 28.1 and 29. Anyways lets leave it with `-bind` - it seems to work.

PR #178, opened 2025-05-15
Test labeler

0xZaddyy:
running test to check if labels work


tcharding:
You need to PR against a repo that has the labeller merged already. That is why I suggested merging it into your own tree to test.

PR #191, opened 2025-05-20
Add model for getnodeaddresses

jamillambert:
GetNodeAddresses returns an Address.  Create a bitcoin::Address model for it and modify the test to use it.

Implement for v18 to v21. In v22 the return changes, label v22 to v28 as TODO.

### DIFF
--- a/integration_test/tests/network.rs
+++ b/integration_test/tests/network.rs
@@ -37,10 +37,11 @@ fn network__get_network_info__modelled() {
 fn network__get_node_addresses() {
     let node = Node::with_wallet(Wallet::None, &[]);
     let json: GetNodeAddresses = node.client.get_node_addresses().expect("getnodeaddresses");
-    assert!(json.0.len() <= 2500);
-    if let Some(addr) = json.0.first() {
+    let res: Result<mtype::GetNodeAddresses, _> = json.into_model();
+    let model = res.expect("GetNodeAddresses into model");
+    assert!(model.0.len() <= 2500);
+    if let Some(addr) = model.0.first() {
         assert!(addr.port > 0);
-        assert!(!addr.address.is_empty());
         assert!(addr.time > 1231006505);
     }
 }
--- a/types/src/model/mod.rs
+++ b/types/src/model/mod.rs
@@ -34,7 +34,9 @@ pub use self::{
         BlockTemplateTransaction, GetBlockTemplate, GetPrioritisedTransactions,
         PrioritisedTransaction,
     },
-    network::{GetNetworkInfo, GetNetworkInfoAddress, GetNetworkInfoNetwork},
+    network::{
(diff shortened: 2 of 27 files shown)


jamillambert (commented):
The CI fail looks unrelated

tcharding:
Yeah we have intermitent CI failures atm, in #147 it was suggested its to do with flaky downloads of Core.

tcharding (approved):
ACK 054862f212ac5e8c7911fae7a0b64f565e3f9b7b

jamillambert:
> Yeah we have intermitent CI failures atm, in #147 it was suggested its to do with flaky downloads of Core. 

I've had them locally too running the tests.

tcharding:
Do you download locally or set `BITCOIND_EXE`?

jamillambert:
Download locally. I'll try and dig a bit and see if I can find out more. But always hard with intermittent errors.

tcharding:
See the alias' in `integration_test/README.md` for running locally without downloading. Otherwise you'll download for every worktree and its too slow.

tcharding:
>  I'll try and dig a bit and see if I can find out more

I would guess we should just be adding repeat/timeout logic to the download code (in build.rs).

tcharding:
Ouch, in review I did not notice that the address returned is **not** a Bitcoin address but a 'node' address. I don't actually know what that means, is it IP address? Eitherway the `model` type is wrong and since the only reason it exists is because of the `bitcoin::Address` we don't need it at all.

integration_test/tests/network.rs
```diff
@@ -37,10 +37,11 @@ fn network__get_network_info__modelled() {
 fn network__get_node_addresses() {
     let node = Node::with_wallet(Wallet::None, &[]);
     let json: GetNodeAddresses = node.client.get_node_addresses().expect("getnodeaddresses");
-    assert!(json.0.len() <= 2500);
-    if let Some(addr) = json.0.first() {
+    let res: Result<mtype::GetNodeAddresses, _> = json.into_model();
+    let model = res.expect("GetNodeAddresses into model");
+    assert!(model.0.len() <= 2500);
```
tcharding, line 42: What does this do?

jamillambert: Nothing the list is empty 😂. It was an AI suggested sanity check that the list was smaller than the max allowed size.

integration_test/tests/network.rs
```diff
@@ -37,10 +37,11 @@ fn network__get_network_info__modelled() {
...
+    if let Some(addr) = model.0.first() {
         assert!(addr.port > 0);
-        assert!(!addr.address.is_empty());
         assert!(addr.time > 1231006505);
     }
```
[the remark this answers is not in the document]
jamillambert: I opened #320 to follow this up. Adding a multi-node network did not work for me either.

PR #194, opened 2025-05-22
Implement addnode method and test

GideonBature:
Going by the conversations with tcharding for PR [#116 ](https://github.com/rust-bitcoin/corepc/pull/116) on implementing one method at a time for easier review: This is first method under network section, addnode which is a specific type that returns a (json null). Once this is approved, I’ll proceed with the next one.


tcharding (approved):
ACK 91d2ac156542ad51ff533bc23a5a967bbe88fcf2

PR #197, opened 2025-05-23
Update integration test feature gating

### LINKED ISSUE #195: integration_test: Improve feature gating
In the integration tests we have feature gating that is prone to going stale. Something like this will silently go stale when we add v30.

```rust
    #[cfg(any(
        feature = "v26",
        feature = "v27",
        feature = "v28",
        feature = "v29",
    ))]
    { .... }
```

Also its annoying to type all this out especially when one has to list every version from 17 - 28 like we have in one place. We need to come  up with a better solution.

jamillambert:
Change the version feature gates to be that version and below to make feature gating the tests easier.

Closes #195


tcharding (approved):
ACK e9196806c87897bab2847911b21546f69323cd8f

tcharding:
So hot right now.

PR #201, opened 2025-05-24
PR Labeler workflow: Fix label name

jamillambert:
The label name in the workflow had an `_` but should be a `-` to match the repo labels. This causes the workflow to fail if this tag is required.

Correct the name to use a -


tcharding (approved):
ACK e6f938126646aaaff2272b53bfce79618d9c7b50

Issue #137, opened 2025-04-30
Test that re-exported `types` auxiliary types are correct


tcharding:
Given RPC method `foobar` returning type `FooBar` with conversion error `FooBarError`.

Our integration testing strategy verifies that `FooBar` is re-exported correctly. If `FooBar` has a field with type `FooBarAuxType` then we do not currently verify that it is correctly re-exported  - i would be nice to come up with a way to prove we got it right. Extremely high chance I made mistakes in this area. Not super likely downstream will notice.

GideonBature:
One way I am thinking is to also use the explicit type check we are using for the normal types. But it might make the test clumsy. To avoid this, I am thinking of creating another test file to check that all re-exports is done correctly including the Auxiliary types.

Another way I am looking at is to use a custom macro method of `assert_field_is_exact_type!`, which we can define and then call for each re-exported types having auxiliary types.

tcharding:
Go for it, happy to review what you come up with.

PR #202, opened 2025-05-24
Add test to verify re-exported auxiliary types

GideonBature:
In order to address the issue in #137 to ensure that re-exported types auxiliary types are correct, I simply added a standalone test file where I defined a macro to assert that fields of re-exported types are of the expected auxiliary types.

This is still a WIP, still need to know if doing something like this aligns with the project and if I am on the right course before I continue...

### DIFF
--- a/integration_test/tests/aux_export.rs
+++ b/integration_test/tests/aux_export.rs
@@ -1,0 +1,196 @@
+// SPDX-License-Identifier: CC0-1.0
+//! Compile-time assertions for type re-exports and auxiliary types.
+
+#![allow(dead_code)]
+
+use node::mtype::*;
+#[allow(unused_imports)]
+use std::collections::BTreeMap;
+
+#[macro_export] macro_rules! assert_field_is_exact_type {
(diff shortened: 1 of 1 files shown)


tcharding:
To review this I asked myself:

1. What are we trying to prove?
2. What does the `integration_test/tests/aux_export.rs` file prove
3. How do I know we didn't miss any of the auxiliary types?
4. When we add a newtype that has an auxiliary type what will stop us from forgetting to test it i.e., does this new file add any protection against forgetting?

If you ask your self these questions perhaps you can come up with the same review I would give?

GideonBature:
> To review this I asked myself:
> 
> 1. What are we trying to prove?
> 2. What does the `integration_test/tests/aux_export.rs` file prove
> 3. How do I know we didn't miss any of the auxiliary types?
> 4. When we add a newtype that has an auxiliary type what will stop us from forgetting to test it i.e., does this new file add any protection against forgetting?
> 
> If you ask your self these questions perhaps you can come up with the same review I would give?

I see, thank you for these questions, I will look into 3 and 4 and come up with a better solution.

PR #206, opened 2025-05-27
integration_test: Pause after node comes up

tcharding:
We are getting intermittent test failures stating that the node is not available. `node` already includes code that hits the RPC API after the node is up so this is a little surprising. However it doesn't hurt to try waiting a little still.

Just sleep for half a second after each node spins up and see if the intermittent failures go away.

### DIFF
--- a/integration_test/src/lib.rs
+++ b/integration_test/src/lib.rs
@@ -69,7 +69,11 @@ impl NodeExt for Node {
             conf.args.push(arg);
         }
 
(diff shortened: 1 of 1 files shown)


jamillambert (approved):
ACK c188d911cd893d84fd03145c5c02f3c3d5928178

This doesn't fix the issue, but it did improve it in my testing. I tried changing it to a 2s wait and still got failures.  I think merge this, but keep the issue open.

PR #207, opened 2025-05-28
Improve node error message

jamillambert:
Include the upstream error message in the NoBitcoindInstance error to aid in debugging.


tcharding:
That retry loop is horrible - it does two totally different things. I think we should spit it up then the error added in this PR doesn't need to contain an `Option`.

PR #208, opened 2025-05-28
Fix typo

jamillambert:
There was a typo in verify, found when renaming the macros using this list. To be honest I think I stared at it for an age before I saw the typo, the compiler was complaining but the macro was there! or so I thought.


tcharding (approved):
ACK 7cb2eae2d5b678d4fed3d729497aeb98bd731bef

PR #210, opened 2025-05-28
Implement disconnectnode method and test

GideonBature:
Going by the conversations with tcharding for PR [#116 ](https://github.com/rust-bitcoin/corepc/pull/116) on implementing one method at a time for easier review: This is the disconnectnode method which is a specific type that returns a (json null). Once this is approved, I’ll proceed with the next one.


integration_test/tests/network.rs
```diff
@@ -33,6 +33,16 @@ fn network__clear_banned() {
     node.client.clear_banned().expect("clearbanned");
 }
 
+#[test]
+fn test_disconnect_node() {
```
tcharding, line 37:
```suggestion
fn network__disconnect_node() {
```

tcharding (commented):
Looks good except the test name. Perhaps since you are using the same PR description for each of these PRs you do we could make it more terse and more meaningful.  Something like

'JSONRPC method foo does not return anything, we would like to test this in case a later version of Core changes this behaviour. Add a function to the client that errors if anything but null is returned then add an integration test that calls this new function.'

GideonBature:
> Looks good except the test name. Perhaps since you are using the same PR description for each of these PRs you do we could make it more terse and more meaningful. Something like
> 
> 'JSONRPC method foo does not return anything, we would like to test this in case a later version of Core changes this behaviour. Add a function to the client that errors if anything but null is returned then add an integration test that calls this new function.'

This is better... Will adopt this.

tcharding (approved):
ACK 359ac383f28c579d5ed79ded0bba0a38cb765f0c

PR #212, opened 2025-05-29
Implement getconnectioncount method and test

GideonBature:
The JSON-RPC method `getconnectioncount` does return a numeric. We want to test this to catch any changes in behavior in future Core versions.

This PR adds a client function that errors if the return value is anything other than `numeric`, along with an integration test that calls this function.


tcharding:
Hey mate, from now on, to make reviewing the re-export change easier can you just throw the re-export in at the end of the list on its own line. Then in a separate commit run `just fmt`. That way we can be sure we don't accidentally introduce any mistakes. Holla if you don't understand what I mean.

GideonBature:
> Hey mate, from now on, to make reviewing the re-export change easier can you just throw the re-export in at the end of the list on its own line. Then in a separate commit run `just fmt`. That way we can be sure we don't accidentally introduce any mistakes. Holla if you don't understand what I mean.

I understand, instead of inserting it manually in the midst of existing re-exports. I will do just that... Thank you.

tcharding:
`version` please.

GideonBature:
Alright, noted. Will update all the existing ones.

PR #213, opened 2025-05-29
Fix integration test failures by making Node::with_conf more robust

### LINKED ISSUE #205: Random integration test failures because: `it appears that bitcoind is not reachable`
There are random integration test failures both in CI and locally, it fails once every 5 or 10 runs. Issue #147 may be related but the suggestion there does not seem to fix this.
The error message is always:
```
failed to create node: it appears that bitcoind is not reachable
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
Running the tests in only a single thread always passes in my testing, so I thought it might be due to the race condition fixed in https://github.com/bitcoin/bitcoin/pull/30394, but it still happens in v27.2 and v28.1. Using only a single thread takes about 4 times as long so is not really a solution.

jamillambert:
The intermittent integration test failures are caused by race conditions in the creation of the node, wallet or the cookie file. Change `Node::with_conf` to check if there is a problem and retry the section where it occurred, or if that still fails retry the whole process of starting the node.

First patch: 
- Split out some sections into helper functions.

Second patch:
- Remove the loop in `with_conf()` that creates `client`, instead loop over the individual parts.
- Create helper functions for creating the `client`, and creating or loading the client `wallet`, both with their own retry logic.
- Reduce the sleep time between retries to reduce the delay, since it looks like the errors happen over shorter time periods than the existing 1 second wait.
- Add functions to wait until the client and the cookie file are available.
- Add a loop to retry the whole process if it fails, since the integration tests pass when repeated, just repeat this section where the failures occur instead.
- Remove all the debugging outputs since it all works now they are not needed.
- Update the rustdocs for the function.

Third patch: 
- Revert #206, since the issue is now solved at it’s source the pause is no longer needed.

Closes #205

### DIFF
--- a/node/src/lib.rs
+++ b/node/src/lib.rs
@@ -282,18 +282,7 @@ impl Node {
 
     /// Launch the bitcoind process from the given `exe` executable with given [Conf] param and create/load the "default" wallet.
     pub fn with_conf<S: AsRef<OsStr>>(exe: S, conf: &Conf) -> anyhow::Result<Node> {
-        let tmpdir =
-            conf.tmpdir.clone().or_else(|| env::var("TEMPDIR_ROOT").map(PathBuf::from).ok());
-        let work_dir = match (&tmpdir, &conf.staticdir) {
-            (Some(_), Some(_)) => return Err(Error::BothDirsSpecified.into()),
-            (Some(tmpdir), None) => DataDir::Temporary(TempDir::new_in(tmpdir)?),
-            (None, Some(workdir)) => {
-                fs::create_dir_all(workdir)?;
-                DataDir::Persistent(workdir.to_owned())
-            }
-            (None, None) => DataDir::Temporary(TempDir::new()?),
-        };
-
+        let work_dir = Self::init_work_dir(conf)?;
         let work_dir_path = work_dir.path();
         if !work_dir_path.exists() {
             panic!("work dir does not exist");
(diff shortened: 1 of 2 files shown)


node/src/lib.rs
```diff
@@ -280,8 +280,104 @@ impl Node {
         Node::with_conf(exe, &Conf::default())
     }
 
-    /// Launch the bitcoind process from the given `exe` executable with given [Conf] param and create/load the "default" wallet.
+    /// Launch the bitcoind process from the given `exe` executable with given [Conf] param and
+    /// create/load the "default" wallet.
```
tcharding, line 284:
```suggestion
    /// create/load the "default" wallet.
    ///
```

node/src/lib.rs
```diff
@@ -280,8 +280,104 @@ impl Node {
...
+                    let _ = process.kill();
+                    continue;
+                }
+                Ok(None) => {
+                    // Process is still running, proceed
+                }
+            }
+
+            if Self::wait_for_cookie_file(cookie_file.as_path(), Duration::from_secs(5)).is_err() {
+                let _ = process.kill();
```
tcharding, line 339: Maybe a code comment here saying why we kill the process if waiting for the cookie file fails?

node/src/lib.rs
```diff
@@ -280,8 +280,104 @@ impl Node {
...
+            let client_base = Self::create_client_base(&rpc_url, &auth)?;
+            let client = match &conf.wallet {
+                Some(wallet) =>
+                    match Self::create_client_wallet(&client_base, &rpc_url, &auth, wallet) {
+                        Ok(client) => client,
+                        Err(e) =>
+                            if attempt == conf.attempts - 1 {
+                                return Err(e);
+                            } else {
+                                let _ = process.kill();
```
tcharding, line 353: And here maybe a comment saying why we kill after waiting for client fails.

node/src/lib.rs
```diff
@@ -280,8 +280,104 @@ impl Node {
...
+                                continue;
+                            },
+                    },
+                None => client_base,
+            };
+            if Self::wait_for_client(&client, Duration::from_secs(5)).is_err() {
+                let _ = process.kill();
```
tcharding, line 360: And here.

node/src/lib.rs
```diff
@@ -293,170 +389,119 @@ impl Node {
...
+            match Client::new_with_auth(rpc_url, auth.clone()) {
+                Ok(client) => return Ok(client),
+                Err(e) =>
+                    if tries == 10 {
+                        return Err(Error::NoBitcoindInstance(e.to_string()).into());
+                    },
+            }
+            thread::sleep(Duration::from_millis(200));
+        }
+        unreachable!()
```
tcharding, line 474: Perhaps this is cleaner?

```rust
        for _ in 0..9 {
            if let Ok(client) = Client::new_with_auth(rpc_url, auth.clone()) {
                return Ok(client);
            }
            thread::sleep(Duration::from_millis(200));
        }
        Client::new_with_auth(rpc_url, auth.clone()).map_err(|e| Error::NoBitcoindInstance(e.to_string()).into())
```

tcharding (commented):
Mad, good work man.

node/src/lib.rs
jamillambert: I have added comments to all the checks where it kills the process and retries.

PR #220, opened 2025-05-31
Implement ping method and test

GideonBature:
The JSON-RPC method `ping` does return null. We want to test this to catch any changes in behavior in future Core versions.


tcharding (approved):
ACK 8b1f5782afd21aba158fe1b23103997067e414c5

tcharding:
Grinding them out! Well done, keep at it.

PR #225, opened 2025-06-01
Implement listbanned method and test

GideonBature:
The JSON-RPC method `listbanned` does return a type. We want to test this to catch any changes in behavior in future Core versions.


GideonBature:
oh my... Let me fix them up.

integration_test/tests/network.rs
```diff
@@ -113,6 +113,16 @@ fn get_peer_info_three_node_network() {
     assert!(node3.peers_connected() >= 1);
 }
 
+#[test]
+fn network__list_banned() {
+    let node = Node::with_wallet(Wallet::None, &[]);
+    let dummy_subnet = "192.0.2.5";
+
+    node.client.set_ban(dummy_subnet, SetBanCommand::Add).expect("setban add");
+    let _: ListBanned = node.client.list_banned().expect("listbanned");
```
tcharding, line 122: Can we add an assertion here that the banned list length is greater than 0.

integration_test/tests/network.rs
```diff
@@ -113,6 +113,16 @@ fn get_peer_info_three_node_network() {
...
+    node.client.set_ban(dummy_subnet, SetBanCommand::Remove).expect("setban remove");
```
tcharding, line 123: If its worth removing the ban then its worth testing banned list length is 0 again here.

types/src/v17/network/mod.rs
```diff
@@ -263,4 +263,13 @@ pub struct ListBanned(pub Vec<Banned>);
 
 /// An item from the list returned by the JSON-RPC method `listbanned`
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
-pub struct Banned(String); // FIXME: The docs are empty so I don't know what shape this is.
+pub struct Banned {
```
tcharding, line 266: Can we have some docs saying that the shape of this type was pulled from Core source code because the method does not show up in the Bitcoin Core RPC CLI docs.

types/src/v17/network/mod.rs
```diff
@@ -263,4 +263,13 @@ pub struct ListBanned(pub Vec<Banned>);
...
+    /// The IP/Subnet of the banned node.
+    pub address: String,
+    /// The UNIX epoch time the ban was expires.
+    pub banned_until: u32,
+    /// The UNIX epoch time the ban was created.
+    pub ban_created: u32,
+    /// The reason for the ban.
+    pub ban_reason: String, // FIXME: No docs for this field.
```
tcharding, line 274: We are never going to be able to remove that FIXME so best not to add it in the first place I rekon.

types/src/v20/network.rs
```diff
@@ -0,0 +1,26 @@
...
+
+use serde::{Deserialize, Serialize};
+
+/// Result of JSON-RPC method `listbanned`.
+///
+/// > listbanned
+///
+/// > List all banned IPs/Subnets.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+pub struct ListBanned(pub Vec<Banned>);
```
tcharding, line 15: This is the same shape for all versions so I'd just use the v17 one everywhere and only add the new `Banned` type.

tcharding (commented):
Looking much better man.

types/src/v17/network/mod.rs
GideonBature: True!

types/src/v20/network.rs
tcharding: oooph you are correct, my bad.

GideonBature:
Woaw... Let confirm this...

tcharding:
The fact that the list is not empty should imply that at least one `Banned` struct was deresrialized. That is why I thought it strange that removing fields didn't make the test fail.

tcharding:
Needs rebase. Can you post when you force push with what changed please. I'm still confused by this PR.

integration_test/tests/network.rs
GideonBature: missed that. Thank you.

types/src/v17/network/mod.rs
```diff
@@ -263,7 +263,17 @@ pub struct ListBanned(pub Vec<Banned>);
 
 /// An item from the list returned by the JSON-RPC method `listbanned`
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
-pub struct Banned(String); // FIXME: The docs are empty so I don't know what shape this is.
+pub struct Banned {
+    // NOTE: Shape taken from Core source code,as method is undocumented in the Bitcoin RPC CLI.
```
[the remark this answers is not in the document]
GideonBature: thank you! Done.

types/src/v22/network.rs
```diff
@@ -0,0 +1,30 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! The JSON-RPC API for Bitcoin Core `v0.22` - network.
```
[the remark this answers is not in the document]
GideonBature: Thank you...

tcharding:
In v17 `listbanned` returns null so I'd expect to see that verified in the client macro. Then in the v17 `types` docs I expected to see 'returns nothing'. Then in what ever version introduces the result I'd expect to see an additional client macro for that version and have it used for subsequent versions. Then the test would probably have to handle the difference macro calls also.

jamillambert:
> I think something is still wrong here. If I comment out the fields in `Banned` and run the tests it still passes?

This seems to be a bigger issue with the tests in general and not just localized to this PR. See issue #241

integration_test/tests/network.rs
```diff
@@ -113,6 +113,23 @@ fn get_peer_info_three_node_network() {
...
+    node.client.set_ban(dummy_subnet, SetBanCommand::Add).expect("setban add");
+    let json: ListBanned = node.client.list_banned().expect("listbanned");
+    assert!(json.0.len() > 0);
+
+    let banned = &json.0[0];
+    assert_eq!(banned.address, dummy_subnet);
+
+    node.client.set_ban(dummy_subnet, SetBanCommand::Remove).expect("setban remove");
+    let json: ListBanned = node.client.list_banned().expect("listbanned");
+    assert!(json.0.len() == 0);
```
jamillambert, line 130: I had another look at this and you could check that the banned address is in the returned list, and then not in the list after removing it. Rather than just checking the size of the list, e.g. 
```rust
#[test]
fn network__list_banned() {
    let banned_subnet = "192.0.2.5/32";

    node.client.set_ban(banned_subnet, SetBanCommand::Add).expect("setban add");
    let json: ListBanned = node.client.list_banned().expect("listbanned");
    assert!(json.0.iter().any(|item| item.address == banned_subnet));

    node.client.set_ban(banned_subnet, SetBanCommand::Remove).expect("setban remove");
    let json: ListBanned = node.client.list_banned().expect("listbanned");
    assert!(json.0.iter().all(|item| item.address != banned_subnet));
}
```

tcharding:
Sorry mate, needs rebase again. I'll try get this one in next.

integration_test/tests/network.rs
GideonBature: Niceee... Thank you. This is much better.

jamillambert (approved):
ACK f76918a186f935dc0da65b4b74ed645f59b137f7

tcharding (approved):
ACK f76918a186f935dc0da65b4b74ed645f59b137f7

tcharding:
Well done for sticking at this one, good to get it over the line.

GideonBature:
Thank you... :).

PR #228, opened 2025-06-03
feat: Add end-to-end `bitcoin-rpc-midas` integration test

### LINKED ISSUE #4: Consider using code generation to produce Rust code from structured RPC information exported from Bitcoin Core
I've been considering an alternative approach to writing a Bitcoin Core JSON RPC client, namely, exporting the structured API information available within Bitcoin Core as JSON, and using that to auto-generate the API.

I created a branch of Core [here](https://github.com/casey/bitcoin/tree/rpc-json) which implements a new JSON RPC command, `api`, which dumps structured API information as JSON.

You can see the output [here](https://github.com/casey/bitcoin/blob/rpc-json/api.json). It contains types and documentation for all RPC commands.

These could be used to automatically generate Rust code, instead of writing them manually and keeping them up to date between versions of Core. Types are not ideal, for example, addresses have type "string". However, you could write a complementary JSON "patch" file which described what the user facing types should be.

For example, `getaddressinfo` takes a single argument which is of type `string`. The patch file could contain instructions for the code generator to expose this as a `bitcoin::Address`:

```json
{
  "getaddressinfo": {
    "arguments": {
      "address": {
        "type": "address",
      }
    }
  }
}
```

The code generator would then load the JSON file dumped by `bitcoin-cli api`, as well as the patch file, and use them to generate user-facing Rust code which exposed the appropriate types, and performed the appropriate conversions when calling Bitcoin Core.

There are a bunch of advantages to this approach. You would be guaranteed to be up-to-date with Core, and could automatically expose documentation from Core as Rust doc comments, without needing to copy them manually. Since everything is auto-generated, you would likely only need a minimum of tests to be confident that things were working.

Some types would require special considerations, like addresses, fee rates, and amounts (which are exposed by the Core RPC with different units in different places). So, it might be wise to consider the patchfile to be a kind of whitelist, and require entries in the patch file for all commands before allowing to code generator to expose them, to ensure that a human had manually confirmed that no types require special handling.

I don't know if I'm going to keep working on this myself. Since `rust-bitcoincore-rpc` has been deprecated, I went down a bit of a rabbit hole thinking about how the whole RPC situation could be improved, and this is what I came up with. However, in reality, we'll probably keep using `rust-bitcoincore-rpc` for the time being, and just hack around the limitations until they become intolerable.

If this approach is fruitful, it would be nice to get the branch which adds the `api` command merged into Core, so it doesn't require continual rebasing. I've been away from Core development for a while, and have been spoiled rotten by Rust, so I might not be the right person to do that.

nervana21:
- Introduces a new test (`bitcoin_rpc_midas.rs`) demonstrating a full PSBT lifecycle:
  - start bitcoind
  - create wallet
  - mine blocks
  - generate address
  - create and sign PSBT
  - finalize and broadcast
  - confirm transaction
- All functionality is exercised through the generated `midas` client with no manual RPC calls.

The test can be run with the following command:

cargo test --test bitcoin_rpc_midas --features 28_0 -- --nocapture

Closes #4

### DIFF
--- a/integration_test/Cargo.toml
+++ b/integration_test/Cargo.toml
@@ -55,11 +55,13 @@ TODO = []                       # This is a dirty hack while writing the tests.
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false, features = ["std", "serde"] }
+bitcoin-rpc-midas = "0.1.0"
 node = { package = "corepc-node", version = "0.8.0", default-features = false }
 rand = "0.8.5"
 env_logger = "0.9.0"
 
 [dev-dependencies]
+tokio = { version = "1.39.0", features = ["full"] }
 
 [patch.crates-io.corepc-client]
 path = "../client"
--- a/integration_test/tests/bitcoin_rpc_midas.rs
+++ b/integration_test/tests/bitcoin_rpc_midas.rs
@@ -1,0 +1,95 @@
+use bitcoin::Amount;
+use bitcoin_rpc_midas::test_node::test_node::BitcoinTestClient;
+use node::anyhow::{anyhow, Result};
+use node::serde_json::json;
+
+#[tokio::test]
+async fn midas_test() -> Result<(), Box<dyn std::error::Error>> {
+    println!("\n=== Starting Midas Test ===");
+
+    let mut client = BitcoinTestClient::new().await?;
+    let info = client.getblockchaininfo().await?;
+    println!("Initial blockchain state:\n{:#?}\n", info);
+
+    let _ = client.mine_blocks(101, 2000).await?;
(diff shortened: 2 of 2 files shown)


tcharding:
I'm not exactly sure what this PR is attempting to show? IIUC `bitcoin-rpc-midas` is a competitor and complete replacement for this repo? Or am I missing something?

nervana21:
Hi @tcharding,
I believe you're right, but that's not a claim I felt confident making on my own.

This PR was an attempt at making the value-add as easy as possible to evaluate for those who understand the ecosystem better than I do. Happy to adapt if you have any bandwidth for guidance 🙂

tcharding:
I'm still unsure what you want me to do with this PR. I posted on #4, maybe that clears up the confusion.

PR #230, opened 2025-06-08
Implement setnetworkactive method and test

GideonBature:
The JSON-RPC method `setnetworkactive` does return a boolean. We want to test this to catch any changes in behavior in future Core versions.

This PR adds a client function that errors if the return value is anything other than `boolean`, along with an integration test that calls this function.


tcharding (approved):
ACK e071183e45a2a0e1937a433007f4617ff9de6f50

PR #233, opened 2025-06-10
Implement importprivkey and test

GideonBature:
The JSON-RPC method `importprivkey` does return null. We want to test this to catch any changes in behavior in future Core versions.

### DIFF
--- a/client/src/client_sync/v17/mod.rs
+++ b/client/src/client_sync/v17/mod.rs
@@ -123,6 +123,7 @@ crate::impl_client_v17__get_received_by_address!();
 crate::impl_client_v17__get_transaction!();
 crate::impl_client_v17__get_unconfirmed_balance!();
 crate::impl_client_v17__get_wallet_info!();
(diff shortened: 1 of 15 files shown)


client/src/client_sync/v17/mod.rs
```diff
@@ -123,6 +123,7 @@ crate::impl_client_v17__get_received_by_address!();
 crate::impl_client_v17__get_transaction!();
 crate::impl_client_v17__get_unconfirmed_balance!();
 crate::impl_client_v17__get_wallet_info!();
+crate::impl_client_v17__import_priv_key!();
```
tcharding, line 126: Since 'pubkey' is a single word I think this should be 'privkey' everywhere.

GideonBature:
completed all suggestions made.

tcharding (approved):
ACK 4b4e0840dd37d7da60af329fbc24dad4c798b3a9

PR #234, opened 2025-06-10
Add missing v18 util methods

jamillambert:
Add `deriveaddresses` and `getdescriptorinfo` methods and tests.

Include `getdescriptorinfo` return shape changes in v19 and v29.

`deriveaddresses` return shape change in v29 TODO


tcharding (approved):
ACK b6dbf1aa9457cebee1ecc644283135af2d854601

PR #235, opened 2025-06-10
Standardize version number format in docs

jamillambert:
There is a variation in the way the version numbers are written in the docs. Some are preceded by a 0 and some aren't. [Core](https://bitcoin.org/en/version-history) states the versions up to and including 21 with preceding 0, and for 22 and above without.

Change all version numbers in the text to be v0.17 - v0.21, and v22 - v29 to match core.


tcharding (approved):
ACK 55298d17e745d9f3da312981b43be0ac87e28516

tcharding:
This is particularly correct, careful you'll be as anal a programmer as me if you are not careful.

Issue #241, opened 2025-06-12
Missing fields in types structs not tested


jamillambert:
Integration tests still pass if there are missing return fields in the JSON-RPC method structs in `types`. 

e.g. `types/src/v17/network/mod.rs:65` `getnettotals`, all the fields can be deleted and the test still passes: 
```rust
#[test]
fn network__get_net_totals() {
    let node = Node::with_wallet(Wallet::None, &[]);
    let _: GetNetTotals = node.client.get_net_totals().expect("getnettotals");
}
```

NB. if there is an extra field or incorrectly named field the test does fail.

tcharding:
I like this solution.

jamillambert:
All of the missing fields have been added in #243, #251, #255, #267 and #272. And #273 adds the serde deny.

PR #243, opened 2025-06-12
Add missing return fields in v17

jamillambert:
When looking into the fix for #241 some missing fields in the v17 types were found.

Add the missing fields.

Remove the now unnecessary change in v29 to `getblocktemplate` since the fields have been there since v17, just undocumented.


types/src/v17/mining/into.rs
```diff
@@ -17,6 +17,7 @@ impl GetBlockTemplate {
         use GetBlockTemplateError as E;
 
         let version = block::Version::from_consensus(self.version);
+        let capabilities = self.capabilities.into_iter().map(|s| s.to_string()).collect::<Vec<_>>();
```
tcharding, line 20: This is mapping vector of `String` to a vector of `String`, we can just use `self.capabilities` directly below.

types/src/v17/wallet/into.rs
```diff
@@ -338,6 +338,11 @@ impl GetTransaction {
             self.block_index.map(|idx| crate::to_u32(idx, "block_index")).transpose()?;
 
         let txid = self.txid.parse::<Txid>().map_err(E::Txid)?;
+        let wallet_conflicts = self
+            .wallet_conflicts
+            .into_iter()
+            .map(|s| s.parse::<Txid>().map_err(E::Txid))
```
tcharding, line 344: We don't want to re-use the `Txid` error here. The error variants are name after the field not after the type. So we need a new variant `WalletConflicts`.

types/src/v17/mining/mod.rs
```diff
@@ -99,6 +104,11 @@ pub struct GetBlockTemplate {
     pub bits: String,
     /// The height of the next block,
     pub height: i64,
+    /// Optional signet challenge
+    #[serde(rename = "signet_challenge")]
+    pub signet_challenge: Option<String>,
```
tcharding, line 109: Why the rename?

tcharding: Oh the mistake was there already, can you remove it anyways please.

types/src/v17/wallet/into.rs
jamillambert: New error variant added.

jamillambert:
Suggested changes all made

tcharding (approved):
ACK e9575cd06e9bfce9e579eb3df4eb3d99f5882680

tcharding:
The capabilities change got mixed up, added in one commit then removed in the next but the commit history in this repo is not awesome anyways so review, merge, YOLO.

PR #244, opened 2025-06-13
Implement abortrescan method and test

GideonBature:
The JSON-RPC method `abortrescan` does return a type. We want to test this to catch any changes in behavior in future Core versions.


integration_test/tests/wallet.rs
```diff
@@ -36,6 +36,14 @@ fn wallet__abandon_transaction() {
     node.client.abandon_transaction(txid).expect("abandontransaction");
 }
 
+#[test]
+fn wallet__abort_rescan() {
+    let node = Node::with_wallet(Wallet::Default, &[]);
+
+    let json: AbortRescan = node.client.abort_rescan().expect("abortrescan");
+    assert!(json.0 == false);
```
jamillambert, line 44:
```suggestion
    assert!(!json.0); // No rescan running, abort should return false
```

It was not clear to me why this should be false and had to look it up. Therefore the comment.

GideonBature: Alright. Noted. Thank you!

jamillambert (approved):
ACK 9a24cfd8715ee4a45a96d392e0f1146c932eb1ac

tcharding (approved):
ACK 9a24cfd8715ee4a45a96d392e0f1146c932eb1ac

PR #249, opened 2025-06-16
Remove redundant tmp file existence check in wallet integration test

GideonBature:
This PR removes the redundant `if file_path.exists()` check before deleting a temporary file.

As pointed out during review of `backupwallet` integration test [here](https://github.com/rust-bitcoin/corepc/pull/247#discussion_r2148959542), the check is unnecessary.


tcharding (approved):
ACK 7eb37025b185462575e9d0f7961f85c0886a2caa

PR #250, opened 2025-06-16
Implement encryptwallet method and test

GideonBature:
The JSON-RPC method `encryptwallet` does return a type. We want to test this to catch any changes in behavior in future Core versions.


jamillambert:
Nice, you're on a roll.

GideonBature:
> Nice, you're on a roll.

Thank you!

integration_test/tests/wallet.rs
```diff
@@ -155,6 +155,25 @@ fn wallet__dump_wallet() {
...
+    let node = Node::with_wallet(Wallet::None, &[]);
+
+    let client = match () {
+        #[cfg(feature = "v20_and_below")]
+        () => node.create_wallet(wallet_name).expect("createwallet"),
+        #[cfg(not(feature = "v20_and_below"))]
+        () => {
+            node.client.create_wallet(wallet_name).expect("createwallet");
+            &node.client
+        },
```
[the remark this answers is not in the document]
GideonBature: True. The `create_wallet` function in the `node/src/lib.rs` works just fine for all versions, just that I wanted to be more consistent with using the create_wallet that is from client. So I will in the test just call the `rpc_url_with_wallet` from the `node/src/lib.rs` to construct the url to the particular wallet and use it, that should work across all versions.

integration_test/tests/wallet.rs
```diff
@@ -151,6 +151,18 @@ fn wallet__dump_wallet() {
...
+    let node = Node::with_wallet(Wallet::None, &[]);
+
+    // NOTE: Create a new wallet and returns a dedicated client instance for it.
+    // This client will send RPCs to the `/wallet/test-wallet` endpoint.
+    let wallet_client = node.create_wallet("test-wallet").expect("createwallet");
+
+    // Use the dedicated client to encrypt the new wallet
+    let _: EncryptWallet = wallet_client.encrypt_wallet("test-passphrase").expect("encryptwallet");
+}
+
```
jamillambert, line 165:
```suggestion
#[test]
fn wallet__encrypt_wallet() {
    let node = Node::with_wallet(Wallet::Default, &[]);
    let _: EncryptWallet = node.client.encrypt_wallet("test-passphrase").expect("encryptwallet");
}
```

I looked into this and the reason for the feature gating on the above tests `dumpprivkey` and `dumpwallet`  is because they are only supported on legacy wallets, which is default on v22 and below, v23-25 with a dedicated function, and not possible above 25. So we need to keep the feature gating on tests themselves to work in v23-25.

The test here works on all versions and doesn't need the feature gating at all.

The current change works, but the whole test can just be as above.

GideonBature: Interesting... Thank you for checking it out and for the suggestion.

jamillambert (approved):
ACK e513cbe112096cd1dcfd7230dac6c420c614d1bb

tcharding (approved):
ACK e513cbe112096cd1dcfd7230dac6c420c614d1bb

PR #251, opened 2025-06-16
Add missing return fields in v18

jamillambert:
Some of the types in v18 had missing return fields.

Add all of the missing fields and update the models, into function and reexports accordingly. Redefinitions of the types, errors and into functions are copies of the v17 versions with the missing field added.


types/src/v18/blockchain/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+    pub ancestor_count: i64,
+    /// Virtual transaction size of in-mempool ancestors (including this one).
+    #[serde(rename = "ancestorsize")]
+    pub ancestor_size: i64,
+    /// DEPRECATED: Modified fees (see above) of in-mempool ancestors (including this one).
+    #[serde(rename = "ancestorfees")]
+    pub ancestor_fees: f64,
+    /// Hash of serialized transaction, including witness data.
+    pub wtxid: String,
+    /// (No docs in Core v0.17.)
```
tcharding, line 60: I merged but perhaps we want to follow up and change this to either make up some docs or remove the version number?

jamillambert: I have redone this method for the changes in all version in #252 and added docs to the `fees` field.

PR #254, opened 2025-06-18
Implement importaddress method and test

GideonBature:
The JSON-RPC method `importaddress` does return null. We want to test this to catch any changes in behavior in future Core versions.

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -279,6 +279,30 @@ fn wallet__get_transaction__modelled() {
     model.unwrap();
 }
 
+#[test]
+fn wallet__import_address() {
+    let node = match () {
+        #[cfg(feature = "v22_and_below")]
+        () => Node::with_wallet(Wallet::Default, &[]),
+        #[cfg(not(feature = "v22_and_below"))]
+        () => {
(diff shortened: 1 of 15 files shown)


integration_test/tests/wallet.rs
```diff
@@ -279,6 +279,30 @@ fn wallet__get_transaction__modelled() {
...
 }
 
+#[test]
+fn wallet__import_address() {
+    let node = match () {
+        #[cfg(feature = "v22_and_below")]
+        () => Node::with_wallet(Wallet::Default, &[]),
+        #[cfg(not(feature = "v22_and_below"))]
+        () => {
+            let node = Node::with_wallet(Wallet::None, &["-deprecatedrpc=create_bdb"]);
```
jamillambert, line 289: Just to comment that this feature gating is required since the RPC is only supported on legacy wallets. Opposed to `encryptwallet` in the other recent PR.

This method of creating a legacy wallet works up to v29 and the other tests like `dumpwallet` and `dumpprivkey` could be updated to match this format.

jamillambert (approved):
ACK 859ec74c8fcd956bcde8479ca76b4851f2e32428

tcharding (approved):
ACK 859ec74c8fcd956bcde8479ca76b4851f2e32428

PR #255, opened 2025-06-18
Add missing return fields in v19

jamillambert:
Some of the types in v19 had missing return fields.

- Remove a redundant v29 `network::GetNetworkInfo` method that is already defined the same in v28.
- Add all of the missing fields from v19 and update the models, into function and reexports accordingly. Redefinitions of the types and into functions are copies of the v17 versions with the missing field added.
- Run the formatter in a separate commit to make the changes to the reexports easier to review.


tcharding (dismissed):
ACK c791ce36796345c808ee3138f0d0aca08809a377

jamillambert:
Rebased on master, only change is reorganising one reexport in v28 types.

Also confirms that the approval is dismissed now when there is a change.

PR #257, opened 2025-06-20
integration test: Document incompatibility of `--all-features`

tcharding:
And I always forget that the integration test crate cannot be built with all features enabled.

Add a section to the README in an attempt to save poor future Tobin the time to work this out again.


jamillambert (approved):
ACK 4430f61e5c949f56e785ea9f7c4c9e109e4e42f0

PR #258, opened 2025-06-20
Add script to lint integration test

tcharding:
The integration test crate is not part of the workspace and also cannot be built wit all features so linting it must be done for each feature.

Add a script that runs the linter.


jamillambert (approved):
ACK 93554f521d44fa24be42f8ada51c89669b4ca26c

PR #260, opened 2025-06-20
Add '()' annotation and error handling to submit_header

GideonBature:
This is a patch for [PR #259](https://github.com/rust-bitcoin/corepc/pull/259) that adds a `let _: () = ...` annotation and an `.expect("submitheader")` call to the integration test for `submit_header`.

The `.expect()` call is necessary to extract the `()` value from the result returned by `submit_header`, making the test both explicit and consistent with the other null-returning client function tests.

Also, Introduced a for loop to increment the header’s nonce until the PoW is valid. This ensures the header meets Bitcoin’s difficulty target before calling `submit_header`.


jamillambert (approved):
ACK d3a047e26bf5cf76080440933a2aa045448defbe

PR #262, opened 2025-06-20
Fix clippy lints in `verify`

0xZaddyy:
This PR resolves all Clippy warnings in the verify crate by applying idiomatic Rust fixes and cleanup where necessary.


jamillambert:
There is an issue with the lockfiles verify was added to the workspace members.

This crate uses `Cargo-minimal.lock` and `Cargo-recent.lock` which need to be updated using the script `contrib/update-lock-files.sh`.

The main Cargo.lock should not be tracked by git so the last patch needs to be dropped.

And can you reword the commit log of the first patch to have a heading with ≤ 50 chars, then a blank line then the rest with ≤ 72 chars per line.

0xZaddyy:
> There is an issue with the lockfiles because verify was added to the workspace members.
> 
> This crate uses `Cargo-minimal.lock` and `Cargo-recent.lock` which need to be updated using the script `contrib/update-lock-files.sh`.
> 
> The main Cargo.lock should not be tracked by git so the last patch needs to be dropped.
> 
> And can you reword the commit log of the first patch to have a heading with ≤ 50 chars, then a blank line then the rest with ≤ 72 chars per line.

Thank you boss, i've fixed it.

tcharding:
A diff with 468 green and 68 red is suspicious for a lint PR. I did not look closely but at least these things are wrong:
- formatting changes
- probably changes to the root manifest are wrong though I didn't think much about it

0xZaddyy:
I added `verify` as a workspace member so when I pushed CI was failing because there was issues with the lock files
So I had to run `.contrib/update-lock-files.sh` to update `cargo-minimal.lock` and `cargo-recent.lock` .

PR #263, opened 2025-06-21
Implement importmulti method and test

GideonBature:
The JSON-RPC method `importmulti` does return a type. We want to test this to catch any changes in behavior in future Core versions.

### DIFF
--- a/client/src/client_sync/v17/mod.rs
+++ b/client/src/client_sync/v17/mod.rs
@@ -128,6 +128,7 @@ crate::impl_client_v17__get_transaction!();
 crate::impl_client_v17__get_unconfirmed_balance!();
 crate::impl_client_v17__get_wallet_info!();
 crate::impl_client_v17__import_address!();
+crate::impl_client_v17__import_multi!();
 crate::impl_client_v17__import_privkey!();
 crate::impl_client_v17__list_address_groupings!();
 crate::impl_client_v17__list_labels!();
@@ -244,4 +245,43 @@ pub enum AddNodeCommand {
 pub enum SetBanCommand {
     Add,
     Remove,
+}
+
+/// Args for the `importmulti` method
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct ImportMultiRequest {
+    /// Descriptor to import. If using descriptor, donot also provide address/scriptPubKey, scripts, or pubkeys.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub desc: Option<String>,
+    /// Type of scriptPubKey (string for script, json for address). Should not be provided if using descriptor.
+    #[serde(rename = "scriptPubKey", skip_serializing_if = "Option::is_none")]
+    pub script_pub_key: Option<ImportMultiScriptPubKey>,
+    /// Creation time of the key expressed in UNIX epoch time, or the string "now" to substitute the current synced blockchain time.
+    pub timestamp: ImportMultiTimestamp,
+}
+
+/// `scriptPubKey` can be a string for script or json for address.
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum ImportMultiScriptPubKey {
+    /// The script.
+    Script(String),
+    /// The address.
+    Address { address: String },
+}
+
+/// `timestamp` can be a number (UNIX epoch time) or the string `"now"`
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+#[serde(untagged)]
+pub enum ImportMultiTimestamp {
+    /// The string "now".
+    Now(String),
+    /// The UNIX epoch time.
+    Time(u64),
+}
+
+impl Default for ImportMultiTimestamp {
+    fn default() -> Self {
+        ImportMultiTimestamp::Now("now".to_string())
+    }
 }
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -8,7 +8,7 @@
 use bitcoin::address::{Address, NetworkChecked};
 use bitcoin::{Amount, PrivateKey};
 use integration_test::{Node, NodeExt as _, Wallet};
-use node::{mtype,AddressType};
+use node::{mtype, AddressType, ImportMultiRequest, ImportMultiScriptPubKey, ImportMultiTimestamp};
 use node::vtype::*;             // All the version specific types.
 use std::fs;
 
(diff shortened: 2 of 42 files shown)


client/src/client_sync/v17/mod.rs
```diff
@@ -245,3 +246,40 @@ pub enum SetBanCommand {
     Add,
     Remove,
 }
+
+/// Args for the `importmulti` method
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct ImportMultiRequest {
```
jamillambert, line 252: A question for @tcharding what is the reason for these argument structs being in `client` and not in `types`?

integration_test/tests/wallet.rs
```diff
@@ -329,6 +329,46 @@ fn wallet__list_received_by_label__modelled() {
     assert!(model.0.iter().any(|item| item.label == label));
 }
 
+#[test]
+fn wallet__import_multi() {
```
jamillambert, line 333: This test looks good in that it is testing using the three argument types of a descriptor, scriptPubkey and address. But this is testing the functionality of core. It would be nice to instead/as well add test cases that have Some `warning` and Some `error` to test the new structs added to `types`.

types/src/v17/wallet/mod.rs
```diff
@@ -547,6 +547,48 @@ pub struct GetWalletInfo {
...
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+pub struct ImportMultiEntry {
+    /// The success.
+    pub success: bool,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// The warnings.
+    pub warnings: Option<Vec<String>>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    /// The error.
+    pub error: Option<JsonRpcError>,
```
jamillambert, line 577: I don't understand the use of `#[serde(default, skip_serializing_if...`, this is the result of the RPC call and therefore `deserialized` from json.

If we want fields that are `None` in the result types to be omitted from the serialization this should be done crate wide and not have different functionality for different RPC calls.  I believe core always omits them?

@tcharding thoughts?

integration_test/tests/wallet.rs
GideonBature: Actually the second request returns an error/warning, while the first and third/last request returns a success.

GideonBature: do you think it's better to use comment to indicate the request that passes and the one that fails in the test? as all is being returned using same structure, although different fields in the structure.

client/src/client_sync/v17/mod.rs
tcharding: Because they are used with the `client` and if one writes their own `client` (which I hope folk will) and just depend on `types` we don't want a bunch of 'random' stuff in there. On the other hand wsince we have the `client-sync` feature we could expose them and then writers of a `client` could still depend on our client to grab them? I'd be open to that. (By expose I mean put them in `lib.rs` or someplace so they are not hidden by the feature.)

types/src/v17/wallet/mod.rs
tcharding: My understanding is, and it may be wrong, that this is a no-op because serde already omits `None` when serializing options. I'm guessing you got this from an LLM @GideonBature? I've seen  LLMs add this when using them.

integration_test/tests/wallet.rs
jamillambert: Yes, and could you add a comment saying why it should return an error.
For the error case can you add an assert that `warnings` and `error` do contain data.
Thanks.

types/src/v17/wallet/mod.rs
GideonBature: Yes, picked it up from #116.

integration_test/tests/wallet.rs
GideonBature: Definitely, let me do that.

client/src/client_sync/v17/mod.rs
```diff
@@ -246,3 +247,40 @@ pub enum SetBanCommand {
...
+    /// Descriptor to import. If using descriptor, donot also provide address/scriptPubKey, scripts, or pubkeys.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub desc: Option<String>,
+    /// Type of scriptPubKey (string for script, json for address). Should not be provided if using descriptor.
+    #[serde(rename = "scriptPubKey", skip_serializing_if = "Option::is_none")]
+    pub script_pub_key: Option<ImportMultiScriptPubKey>,
```
tcharding, line 259: This arg is required by v17 and v29 so shouldn't be `Option`, right?

client/src/client_sync/v17/mod.rs
```diff
@@ -246,3 +247,40 @@ pub enum SetBanCommand {
     Add,
     Remove,
 }
+
+/// Args for the `importmulti` method
+#[derive(Clone, Debug, Default, Deserialize, PartialEq, Serialize)]
+pub struct ImportMultiRequest {
+    /// Descriptor to import. If using descriptor, donot also provide address/scriptPubKey, scripts, or pubkeys.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub desc: Option<String>,
```
tcharding, line 256: Perhaps throw a comment on this 'from Core vXYZ onwards'.

client/src/client_sync/v17/mod.rs
```diff
@@ -246,3 +247,40 @@ pub enum SetBanCommand {
...
+    /// The address.
+    Address { address: String },
+}
+
+/// `timestamp` can be a number (UNIX epoch time) or the string `"now"`
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
+#[serde(untagged)]
+pub enum ImportMultiTimestamp {
+    /// The string "now".
+    Now(String),
```
tcharding, line 279: If we know what string it is we don't need `String`.

client/src/client_sync/v17/mod.rs
```diff
@@ -246,3 +247,40 @@ pub enum SetBanCommand {
...
+    /// The UNIX epoch time.
+    Time(u64),
```
tcharding, line 281:
```suggestion
    /// The UNIX timestamp.
    Time(u64),
```

'epoch' is the start - time 0

client/src/client_sync/v17/mod.rs
```diff
@@ -246,3 +247,40 @@ pub enum SetBanCommand {
...
+}
+
+impl Default for ImportMultiTimestamp {
+    fn default() -> Self { ImportMultiTimestamp::Now("now".to_string()) }
+}
```
tcharding, line 286: Don't add this please. Its just unnecessary code to maintain.

tcharding (commented):
Everything looks good except I'd like to see the test improved upon please. Its great that you are testing so thoroughly its just a bit unclear by reading the test _exactly_ what it is testing.

Since we are primarily interested in proving the shape of the returned type perhaps structure the test, and include code comments, around this idea.

E.g, Do some set up and comment what is expected then assert that we get what is expected.

Also please comment if things are version specific, ATM it looks like descriptors are supported all the way back to v17 but the docs don't support that, or is it just undocumented?

Thanks

GideonBature:
> Everything looks good except I'd like to see the test improved upon please. Its great that you are testing so thoroughly its just a bit unclear by reading the test _exactly_ what it is testing.
> 
> Since we are primarily interested in proving the shape of the returned type perhaps structure the test, and include code comments, around this idea.
> 
> E.g, Do some set up and comment what is expected then assert that we get what is expected.
> 
> Also please comment if things are version specific, ATM it looks like descriptors are supported all the way back to v17 but the docs don't support that, or is it just undocumented?
> 
> Thanks

Yes, descriptors is not mentioned in v17, yet the test still passes, prolly will exclude the test using descriptors for v17.

client/src/client_sync/v17/mod.rs
GideonBature: yes, it is required by v17 - v29. Why I made it Option is because for v18 - v29 scriptPubkey or descriptor is required but not both.

client/src/client_sync/v17/mod.rs
GideonBature: Noted!

GideonBature:
Implements suggestions and also, add more cases to test, which ultimately lead to a silent error where just any random script wouldn't work for v17, best an address or descriptor be used. While everything works for v18 onwards.

GideonBature:
v17 is failing when using a script. I want to look into why it's failing, I will come back to it.

client/src/client_sync/v17/mod.rs
tcharding: Needs rename to `script_pubkey`.

tcharding (approved):
ACK 283c53ef9caeb90ad6f8d293097ea13f4dda0325

tcharding:
Good effort man, that test is nice.

client/src/client_sync/v17/mod.rs
GideonBature: will open a patch PR for the renaming.

GideonBature:
Thank you. :).

PR #264, opened 2025-06-21
fix all clippy warnings for verify

0xZaddyy:
This PR resolves all Clippy warnings in the verify crate by applying idiomatic Rust fixes and performing necessary cleanup.

Additionally, I updated the `cargo-minimal.lock` and `cargo-recent.lock` files to reflect the inclusion of the verify crate in the workspace.


tcharding:
If you want to add `verify` to the workspace unfortunately that means you'll have to look in to how `run_task.sh` works.

I would suggest it is easier if you just fix the lint warnings and leave adding `verify` to the workspace to be done in a separate PR. Yes this means your changes aren't tested in CI but I can check them locally. I can do the 'add to workspace' PR if you don't want to read that script.

https://github.com/rust-bitcoin/rust-bitcoin-maintainer-tools/blob/master/ci/run_task.sh

PR #267, opened 2025-06-23
Add missing return fields in v20 and v21

jamillambert:
Some of the types in v20 and v21 had missing return fields.

For each of the following RPCs that were missing a return field in v20 or v21 check for further changes up to v29 and implement them:

-  `createmultisig`
- `getblockchaininfo`
- `getmempoolinfo`
- `getnetworkinfo`
- `getpeerinfo`

Add all of the missing fields update the models, into function and reexports accordingly. Redefinitions of the types and into functions are copies of the most recent version of the RPC that had fields added.

Draft because on top of #255, first 3 patches. This PR is the last 5 patches only.


jamillambert:
Rebased. Didn't even get a merge conflict in the reexports. Score!

Issue #236, opened 2025-06-11
`verify` and `integration_test`  are not linted in CI


tcharding:
Both these crates have a bunch of `clippy` warnings, it would be nice to clear them then enforce it in CI. They are skipped right now because they are excluded from the workspace.

0xZaddyy:
oh now i understand, Thank you sir.

PR #268, opened 2025-06-24
Lint non-workspace crates in CI

tcharding:
Do a bunch of preparation and then add commands to the already present lint job in CI to lint `verify` and `integration_tests`.

Close: #236

### DIFF
--- a/justfile
+++ b/justfile
@@ -3,6 +3,9 @@ set export
 REPO_DIR := `git rev-parse --show-toplevel`
 
 alias ulf := update-lock-files
+alias l := lint
+alias li := lint-integration-tests
+alias lv := lint-verify
 
 default:
   @just --list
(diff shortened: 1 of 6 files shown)


justfile
```diff
@@ -18,6 +21,14 @@ check:
 # Lint everything.
 lint:
```
jamillambert, line 22:
```suggestion
lint: lint-verify lint-integration-tests
```

The other tests need to be listed as dependencies. It doesn't run as it currently is and I had to look up why: If you list them below it's run in the shell so would need to be `just lint-verify`.

0xZaddyy:
Looking at this now, I clearly see the mistakes I made earlier and I really appreciate how clean and well structured your commits are it's a great reference for improving mine.

tcharding:
Thanks @0xZaddyy, glad you found it useful. Stick at it man, its a long path.

PR #269, opened 2025-06-24
Implement importpubkey method and test

GideonBature:
The JSON-RPC method `importpubkey` does not return anything. We want to test this to catch any changes in behavior in future Core versions.


tcharding (approved):
ACK 1430723d36f4531132de0abd761e01ebb8c0a566

PR #271, opened 2025-06-25
Add new return fields from later versions to v17 types

jamillambert:
Some of the methods have new return fields in later versions. Add the fields to the existing types in v17 as options and update the errors, models and into functions.

Patches:
- Remove an option from a non optional field.
- Add fields to `decodescript`, update the models, errors and into functions.
- Reorder the list in `logging` alphabetically, no other changes.
- Add fields to `logging`, remove redefinitions for later versions and reexport v17 for all versions.
- Add field to `scriptpubkey`.
- Add fields to `gettransaction`, add the new returned type `LastProcessedBlock`, update the models, errors and into functions.


jamillambert:
This is a deviation from what was done before, adding the fields to v17 rather than redefining the whole type for a later version. It is significantly easier, in particular if there is just a single return field added in a later version.

types/src/v17/control.rs
```diff
@@ -52,25 +52,34 @@ pub struct Locked {
...
-    pub bench: bool,
-    pub zmq: bool,
-    pub db: bool,
-    pub rpc: bool,
-    pub estimatefee: bool,
     pub addrman: bool,
-    pub selectcoins: bool,
-    pub reindex: bool,
+    pub bench: bool,
+    pub blockstorage: Option<bool>, // v23 and later only
```
tcharding, line 57: TL;DR At this stage I don't want to make changes like this because it invalidates the whole design of the crate. 

I accidentally did this same thing once but I came to the conclusion that doing so totally invalidates the whole design of the crate. The _whole point_ of `v17::Foo` is that one knows that this is the exact shape of a `Foo` returned by Bitcoin Core v17. If one wants to not think about the version and check all the optional fields then one can first convert to a `model`  type and do so. The central claim of this repo is that these use cases are different and keeping them separate will lead to better software. This was an intuitive leap and only time will tell if I was correct.

jamillambert:
Done separately for every version as suggested in #272.

PR #272, opened 2025-06-26
Add missing return fields in v22 to v29

jamillambert:
Some of the types in v22 to v29 had missing return fields. All missing fields that are shown by adding `#[serde(deny_unknown_fields)]` to the structs in all versions have now been added.

Minor reorganisation of the code, in separate commits which describe what was done.

Add all of the missing fields update the models, error, into function and reexports accordingly. Redefinitions of the types, error and into functions are modified copies of the most recent version of the RPC that had fields added.


jamillambert:
Sorry this was a big one. It's the last of it's kind though, there should be no more missing fields for any version and we can then add the `#[serde(deny_unknown_fields)]` to everything to close issue #241.

PR #273, opened 2025-06-27
Ensure missing return fields from types structs are tested for

### LINKED ISSUE #241: Missing fields in types structs not tested

jamillambert:
If there are missing return fields in the types structs the integration tests still pass.

All of the missing fields have been added in #243, #251, #255, #267 and #272.

The solution suggested in issue #241 is adding `#[serde(deny_unknown_fields)]` to each struct which makes the test fail if there are fields returned that are not in the struct.

- First two patches: do a small code reorganisation to make sure the third patch that uses a search and replace catches all cases.
- Third patch: add `#[serde(deny_unknown_fields)]` to all structs in `types` that derive `Serialize` and therefore use serde. Using a search and replace.

Closes #241

### DIFF
--- a/types/src/lib.rs
+++ b/types/src/lib.rs
@@ -180,6 +180,7 @@ pub fn compact_size_decode(slice: &mut &[u8]) -> u64 {
 /// backwards compatible so we only provide it not a v0.17 specific type. The `mtype::ScriptPubkey`
 /// mirrors this design (but with concrete `rust-bitcoin` types).
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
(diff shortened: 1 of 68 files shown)


tcharding (approved):
ACK 16ae504cb238a4c18571bc9dd9215939a8347d6e

PR #276, opened 2025-06-28
Implement importwallet method and test

GideonBature:
The JSON-RPC method `importwallet` does not return anything. We want to test this to catch any changes in behavior in future Core versions.


tcharding (approved):
ACK c62b4d23ab0372bb64c99b70382708c0476837df

PR #278, opened 2025-06-30
Implement lockunspent method and test

GideonBature:
The JSON-RPC method `lockunspent` does return a type. We want to test this to catch any changes in behavior in future Core versions.


integration_test/tests/wallet.rs
```diff
@@ -502,6 +502,15 @@ fn wallet__load_wallet__modelled() {
     create_load_unload_wallet();
 }
 
+#[test]
+fn wallet__lock_unspent() {
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    node.fund_wallet();
+
+    let json: LockUnspent = node.client.lock_unspent(false).expect("lockunspent");
+    assert!(json.0);
```
tcharding, line 511: We could test locking and also unlocking?

types/src/v17/wallet/mod.rs
```diff
@@ -928,6 +928,24 @@ impl LoadWallet {
...
+/// > 1. unlock                  (boolean, required) Whether to unlock (true) or lock (false) the specified transactions
+/// > 2. transactions            (json array, optional, default=[]) The transaction outputs and within each, the txid (string) vout (numeric).
+/// >   [
+/// >   {                   (json object)
+///   > "txid": "hex",    (string, required) The transaction id
+///   > "vout": n,        (numeric, required) The output number
+/// >   },
+/// >   ...
+///]
+/// > 3. persistent              (boolean, optional, default=false) Whether to write/erase this lock in the wallet database, or keep the change in memory only. Ignored for unlocking.
```
tcharding, line 945: Lets not document all the optional args, just the required ones.

integration_test/tests/wallet.rs
GideonBature: Yes, that would even be better.

tcharding (approved):
ACK 045b25e2eda189a032bbadee4e24115d5cbb464e

PR #283, opened 2025-07-04
Implement sethdseed method and test

GideonBature:
The JSON-RPC method `sethdseed` does not return anything. We want to test this to catch any changes in behavior in future Core versions.


jamillambert (approved):
ACK ee82283a64cb730ab30962cf1006ba6750463999

integration_test/tests/wallet.rs
```diff
@@ -568,6 +568,28 @@ fn wallet__set_wallet_flag() {
...
 }
 
+#[test]
+fn wallet__set_hd_seed() {
+    let node = match () {
+        #[cfg(feature = "v22_and_below")]
+        () => Node::with_wallet(Wallet::Default, &[]),
+        #[cfg(not(feature = "v22_and_below"))]
+        () => {
+            let node = Node::with_wallet(Wallet::None, &["-deprecatedrpc=create_bdb"]);
```
tcharding, line 578: What are you adding this flag for? I see you've done this a bunch of times and I never got round to asking, my bad. I got the test to pass without it.

When you force push perhaps add 
```rust
            // sethdseed is only used for non-descriptor wallets which are the default in v23 onwards.
```

because I had to go digging to find out why we need a legacy wallet.

integration_test/tests/wallet.rs
```diff
@@ -568,6 +568,28 @@ fn wallet__set_wallet_flag() {
...
+            node.client.create_legacy_wallet("wallet_name").expect("createlegacywallet");
+            node
+        }
+    };
+
+    #[cfg(feature = "v20_and_below")]
+    {
+        let mining_addr = node.client.new_address().expect("newaddress");
+        node.client.generate_to_address(101, &mining_addr).expect("generatetoaddress");
+    }
```
tcharding, line 588: I rekon we can just use `fund_wallet` without the feature gate.

integration_test/tests/wallet.rs
GideonBature: I added `-deprecatedrpc=create_bdb` because the test calls `sethdseed`, which is only available on legacy wallets. From v23 onwards, Bitcoin Core uses descriptor wallets by default, and BDB wallet creation is deprecated unless explicitly re-enabled via this flag. Without it, I get the following error:
>
> ```
> createlegacywallet: JsonRpc(Rpc(RpcError { code: -4, message: "BDB wallet creation is deprecated and will be removed in a future release. In this release it can be re-enabled temporarily with the -deprecatedrpc=create_bdb setting.", data: None }))
> ```

GideonBature: I added `-deprecatedrpc=create_bdb` because the test calls `sethdseed`, which is only available on legacy wallets. From v23 onwards, Bitcoin Core uses descriptor wallets by default, and BDB wallet creation is deprecated unless explicitly re-enabled via this flag. Without it, I get the error below:
>
> ```
> createlegacywallet: JsonRpc(Rpc(RpcError { code: -4, message: "BDB wallet creation is deprecated and will be removed in a future release. In this release it can be re-enabled temporarily with the -deprecatedrpc=create_bdb setting.", data: None }))
> ```

integration_test/tests/wallet.rs
GideonBature: That's right, I just confirmed. Thank you.

integration_test/tests/wallet.rs
GideonBature: That is why for v23 onwards, some of the rpc methods are only supported on legacy wallet, that is why I use legacy wallet, and to create_legacy_wallet, some of the methods also required passing in that flag to create a BDB wallet for them to work.

GideonBature:
Removed the feature gating and used `fund_wallet` method as suggested.

integration_test/tests/wallet.rs
tcharding: Thanks for the explanation. Weirdly I swear the test to pass yesterday without the flag. Perhaps we could add a `Legacy` variant to `Wallet` then use `with_wallet(Wallet::Legacy)` to hide all these details? Want to either have a crack at that or write up an issue explaining the problem? Just holla if you don't want to and I can write an issue.

PR #286, opened 2025-07-09
Implement walletlock method and test

GideonBature:
The JSON-RPC method `walletlock` does not return anything. We want to test this to catch any changes in behavior in future Core versions.


tcharding (approved):
ACK 22acb266f61e495b5ae648a8a5b854c1c3f07147

PR #287, opened 2025-07-10
Implement walletpassphrase method and test

GideonBature:
The JSON-RPC method `walletpassphrase` does not return anything. We want to test this to catch any changes in behavior in future Core versions.

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -622,6 +622,17 @@ fn wallet__wallet_lock() {
     let _: () = node.client.wallet_lock().expect("walletlock");
 }
 
+#[test]
+fn wallet__wallet_passphrase() {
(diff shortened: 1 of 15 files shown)


integration_test/tests/wallet.rs
```diff
@@ -622,6 +622,17 @@ fn wallet__wallet_lock() {
     let _: () = node.client.wallet_lock().expect("walletlock");
 }
 
+#[test]
+fn wallet__wallet_passphrase_and_lock() {
```
jamillambert, line 626: Is there a reason this is called `wallet_passphrase_and_lock` and not just `wallet_passphrase`?

tcharding (approved):
ACK 86b06c5c2c5f48e1255f00acab32c6f8aacac12f

PR #289, opened 2025-07-10
Address all `TODO` in v21

jamillambert:
Go through all the `TODO` in the v21 types table. Add all the missing RPCs including models when required and tests.

Add a new macro `create_wallet_with_descriptors` that creates a descriptor wallet for use with v21 and v22 where the default is a legacy wallet. This is required for the `importdescriptors` test.

Move the `wallet` module to a subdirectory since there are multiple new RPCs in v21.

### DIFF
--- a/client/src/client_sync/v21/wallet.rs
+++ b/client/src/client_sync/v21/wallet.rs
@@ -30,6 +30,21 @@ macro_rules! impl_client_v21__create_wallet_with_descriptors {
     };
 }
 
+/// Implements Bitcoin Core JSON-RPC API method `importdescriptors`
+#[macro_export]
+macro_rules! impl_client_v21__import_descriptors {
+    () => {
+        impl Client {
+            pub fn import_descriptors(
+                &self,
+                requests: &[ImportDescriptorsRequest],
+            ) -> Result<ImportDescriptors> {
+                self.call("importdescriptors", &[into_json(requests)?])
+            }
+        }
+    };
+}
+
 /// Implements Bitcoin Core JSON-RPC API method `unloadwallet`
 #[macro_export]
 macro_rules! impl_client_v21__unload_wallet {
--- a/client/src/client_sync/v21/wallet.rs
+++ b/client/src/client_sync/v21/wallet.rs
@@ -45,6 +45,18 @@ macro_rules! impl_client_v21__import_descriptors {
     };
 }
 
+/// Implements Bitcoin Core JSON-RPC API method `send`.
+#[macro_export]
+macro_rules! impl_client_v21__send {
+    () => {
+        impl Client {
+            pub fn send(&self, outputs: &BTreeMap<String, f64>) -> Result<Send> {
+                self.call("send", &[into_json(outputs)?])
+            }
+        }
+    };
+}
+
 /// Implements Bitcoin Core JSON-RPC API method `unloadwallet`
 #[macro_export]
 macro_rules! impl_client_v21__unload_wallet {
--- a/client/src/client_sync/v21/wallet.rs
+++ b/client/src/client_sync/v21/wallet.rs
@@ -9,6 +9,27 @@
 //!
 //! See or use the `define_jsonrpc_minreq_client!` macro to define a `Client`.
 
+/// Implements Bitcoin Core JSON-RPC API method `createwallet` with descriptors=true (descriptor wallet)
+#[macro_export]
+macro_rules! impl_client_v21__create_wallet_with_descriptors {
+    () => {
+        impl Client {
+            pub fn create_wallet_with_descriptors(&self, wallet: &str) -> Result<CreateWallet> {
+                let args = [
+                    wallet.into(),
+                    false.into(),            // disable_private_keys
+                    false.into(),            // blank
+                    serde_json::Value::Null, // passphrase
+                    false.into(),            // avoid_reuse
+                    true.into(),             // descriptors=true
+                    serde_json::Value::Null, // load_on_startup
+                ];
+                self.call("createwallet", &args)
+            }
+        }
+    };
+}
+
 /// Implements Bitcoin Core JSON-RPC API method `unloadwallet`
 #[macro_export]
 macro_rules! impl_client_v21__unload_wallet {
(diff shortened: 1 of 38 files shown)


integration_test/tests/generating.rs
```diff
@@ -8,6 +8,39 @@ use integration_test::{Node, NodeExt as _, Wallet};
...
+        model.unwrap();
+    }
+
+    #[cfg(not(feature = "v24_and_below"))]
+    {
+        // Check with `submit = false` so that `hex` is returned. v25 and later only.
+        let json: GenerateBlock = node.client.generate_block(&mining_addr.to_string(), &transactions, false).expect("generateblock");
+        let model: Result<mtype::GenerateBlock, GenerateBlockError> = json.into_model();
+        model.unwrap();
+    }
```
tcharding, line 41: Can do as a follow up. I rekon the feature guards could just guard the `generate_block` call. The other two lines are identical.

types/src/v21/util.rs
```diff
@@ -0,0 +1,36 @@
...
+/// > {                               (json object)
+/// >   "name" : {                    (json object) The name of the index
+/// >     "synced" : true|false,      (boolean) Whether the index is synced or not
+/// >     "best_block_height" : n     (numeric) The block height to which the index is synced
+/// >   },
+/// >   ...
+/// > }
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[serde(deny_unknown_fields)]
+pub struct GetIndexInfo(pub std::collections::BTreeMap<String, GetIndexInfoName>);
```
tcharding, line 26: nit: I'd prefer to import `BTreeMap`.

tcharding (commented):
Only got to reviewing half mate, I'll come back. Gotta do some kids stuff now.

types/src/v21/util.rs
```diff
@@ -0,0 +1,36 @@
...
+/// > Arguments:
+/// > 1. index_name    (string, optional) Filter results for an index with a specific name.
+/// >
+/// > Result:
+/// > {                               (json object)
+/// >   "name" : {                    (json object) The name of the index
+/// >     "synced" : true|false,      (boolean) Whether the index is synced or not
+/// >     "best_block_height" : n     (numeric) The block height to which the index is synced
+/// >   },
+/// >   ...
```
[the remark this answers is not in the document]
jamillambert: I removed the optional args and result section from the docs of all the added RPCs in this PR.

client/src/client_sync/v21/wallet.rs
```diff
@@ -9,6 +9,66 @@
...
+                    serde_json::Value::Null, // passphrase
+                    false.into(),            // avoid_reuse
+                    true.into(),             // descriptors=true
+                    serde_json::Value::Null, // load_on_startup
+                ];
+                self.call("createwallet", &args)
+            }
+        }
+    };
+}
```
tcharding, line 31: I think this macro should be done like the v23 one i.e., include function `create_wallet` as well as the with desrciptor function. Then in mod.rs of v23 and v24 only call this macro not the v17 one and from v23 onwards only call the v23 one.

tcharding: And I"d comment liberally about args to the RPC method changing and that we only include args we actually use in the integration tests - or some such thing.

integration_test/tests/wallet.rs
```diff
@@ -310,6 +310,26 @@ fn wallet__import_address() {
...
+    let address = node.client.new_address().expect("failed to get new address");
+    let descriptor = format!("addr({})", address);
+
+    let request = ImportDescriptorsRequest {
+        desc: descriptor,
+        timestamp: serde_json::Value::String("now".to_string()),
+    };
+
+    let _: ImportDescriptors = node.client.import_descriptors(&[request]).expect("importdescriptors");
+}
```
tcharding, line 331: From what I just read and assuming I understand the RPC method, this is going to just return 'success=true' because of use of "now". If we are going to do such a basic test (and that is perfectly ok) we don't need to have the `timestamp` arg at all - just insert it in the client method that does the RPC call.

Another, more thorough test idea (although I'm not 100% its necessary) would be to

1. get the current time
2. get an address and send some coins to it
3. get the descriptor
4. mine 100 blocks
5. scan for that descriptor using the time from (1)

Personally I'd just remove the "now" and comment that this just a basic test and likely good enough for now. There are other tests we can write that add more value and are better uses of your time.

The takeaway here is we (all of us) should try to understand _exactly_ what we are testing and test as simply as possible to get that coverage.

integration_test/tests/wallet.rs
```diff
@@ -546,6 +585,23 @@ fn wallet__unload_wallet() {
...
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    node.fund_wallet();
+    let address = node.client.new_address().expect("failed to create new address");
+
+    let mut outputs = BTreeMap::new();
+    outputs.insert(address.to_string(), 0.001);
+
+    let json: Send = node.client.send(&outputs).expect("send");
+    let model: Result<mtype::Send, SendError> = json.into_model();
+    model.unwrap();
```
tcharding, line 602: Cool. This works (I debug printed the result of the last unwrap but I'm baffled as to _why_ it works. Can you explain? The docs say 'data' is required but I don't understand why it works without it and what 'data' even is?

tcharding (approved):
ACK a150076d74721ac341f4b6198df4ae52addfffdf

tcharding:
A lot of work here man, well done.

client/src/client_sync/v21/wallet.rs
tcharding: Addedd #292

integration_test/tests/wallet.rs
jamillambert: `data` is not required, even though it doesn't say so in the docs. From what I read `It is for arbitrary hex-encoded data in a transaction, typically as an [OP_RETURN]  output e.g. { "data": "deadbeef" }`.

integration_test/tests/wallet.rs
jamillambert: I went with the more thorough option of changing the test in #293

integration_test/tests/wallet.rs
tcharding: ah, thanks.

Issue #294, opened 2025-07-15
Testing RPCs that require an external signer


jamillambert:
`enumeratesigners` and `walletdisplayaddress` both require an external signer to work. They have been left untested until a solution for this has been implemented, e.g. a dummy signer script.

jamillambert:
`enumeratesigners` tested in #346. Changing title to state `walletdisplayaddress` to make finding it easier.

jamillambert:
Closing since all development has now shifted to forgejo.

PR #295, opened 2025-07-15
Address all TODO in v22

jamillambert:
Go through all the TODO in the v22 types table. Add all the missing RPCs. 

Redefine `listdescriptors` in v25 due to added return field.

`enumeratesigners` and `walletdisplayaddress` RPCs require an external signer and have been left untested because of this. Opened #294.

### DIFF
--- a/types/src/v25/wallet.rs
+++ b/types/src/v25/wallet.rs
@@ -77,7 +77,7 @@ pub struct DescriptorInfo {
     pub internal: Option<bool>,
     /// Defined only for ranged descriptors.
     pub range: Option<[u64; 2]>,
-    /// The next index to generate addresses from; defined only for ranged descriptors.
+    /// Same as `next_index` field. Kept for compatibility reason.
     pub next: Option<u64>,
     /// The next index to generate addresses from; defined only for ranged descriptors.
     pub next_index: Option<u64>,
--- a/types/src/v25/wallet.rs
+++ b/types/src/v25/wallet.rs
@@ -50,6 +50,38 @@ impl CreateWallet {
     /// Returns the created wallet name.
     pub fn name(self) -> String { self.into_model().name }
 }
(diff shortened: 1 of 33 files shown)


types/src/v25/wallet.rs
```diff
@@ -50,6 +50,38 @@ impl CreateWallet {
...
+    pub descriptor: String,
+    /// The creation time of the descriptor.
+    pub timestamp: u64,
+    /// Activeness flag.
+    pub active: bool,
+    /// Whether this is an internal or external descriptor; defined only for active descriptors.
+    pub internal: Option<bool>,
+    /// Defined only for ranged descriptors.
+    pub range: Option<[u64; 2]>,
+    /// The next index to generate addresses from; defined only for ranged descriptors.
```
tcharding, line 80:
```suggestion
    /// Same as `next_index` field. Kept for compatibility reason.
```

types/src/v22/wallet/error.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[cfg(feature = "std")]
+impl std::error::Error for WalletDisplayAddressError {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use WalletDisplayAddressError::*;
+        match *self {
+            Address(ref e) => Some(e),
+        }
+    }
+}
```
tcharding, line 33: We don't need all this man, we can just return the `address::ParseError` directly since there is only one error. I have a vague feeling you added a single field enum recently and I missed observing this, correct?

jamillambert: I have removed the error enum. I'll have a look for the other one.

tcharding:
Man I was super confused because there is a docs change in the _final_ patch instead of the one before - I thought I had a `jj` fetching problem. Merging anyways - YOLO.

PR #300, opened 2025-07-16
Remove `GenerateBlockError` in v21-24

jamillambert:
This only has one field and the error can be returned directly.

From v25 onwards there are 2 distinct types and the error enum has been left.


tcharding (approved):
ACK 925b5becee9451e1c80a501972839bdfe0e0afa9

PR #301, opened 2025-07-16
Address all TODO in v23

jamillambert:
Go through all the TODO in the v23 types table. Add all the missing RPCs.

- Add getblockfrompeer macro and test
- Add restorewallet struct, macro and test
- Add newkeypool macro and test
- Add getdeploymentinfo struct, model and test
- Run the formatter

### DIFF
--- a/client/src/client_sync/v23/blockchain.rs
+++ b/client/src/client_sync/v23/blockchain.rs
@@ -25,6 +25,18 @@ macro_rules! impl_client_v23__get_block_from_peer {
     };
 }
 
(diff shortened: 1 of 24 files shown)


jamillambert:
`GetDeploymentInfo` has a lot of return fields, and only `pub hash: BlockHash` that is modelled, all the sub structs are also in the model since they are a different type, even though all the fields are identical to the ones in `types`. I wasn't sure if this is correct or if there is a better way, also not sure about the error for it due to the multi layered into_model functions that don't really do anything.

tcharding (approved):
ACK 828454a340635cd63810f57a1433dc9a28363c7b

PR #71, opened 2025-02-20
Fix to download the correct binary for Apple silicon Macs

alpinevm:
Previously, using the (node) crate with the `from_downloaded()` trait would fail on apple silicon with `bad CPU type in executable`

Part of the update includes some code-signing, this is necessary for unsigned binaries (like bitcoind) to run on Arm Macs:
https://github.com/bitcoin/bitcoin/pull/26694#issuecomment-1348887186

PR #309, opened 2025-07-24
Only codesign bitcoind if necessary

alpinevm:
Extension of [PR #71](https://github.com/rust-bitcoin/corepc/pull/71), which introduced initial support for codesigning in the `node` crate. As of version 29, Bitcoin Core has [begun codesigning](https://github.com/bitcoin/bitcoin/pull/31407) the `bitcoind` binary for macOS users. This broke the current logic of the node crate here because we're currently signing every time. This results in the build step of this crate panicking with:

```
  --- stderr
..../target/debug/build/corepc-node-ab8800f75cc98512/out/bitcoin/bitcoin-29.0/bin/bitcoind: is already signed
```

To fix this, I've added some simple validation code that checks if the binary is already signed and subsequently skips the codesign step.

Verified on macOS Sequoia 15.5 (24F74) with M3 silicon using:

```
cargo test --features 29_0,download
cargo test --features 28_0,download
```

to ensure backward compatibility.

### DIFF
--- a/node/build.rs
+++ b/node/build.rs
@@ -160,17 +160,26 @@ mod download {
             #[cfg(all(target_os = "macos", target_arch = "aarch64"))]
             {
                 use std::process::Command;
-                let status = Command::new("codesign")
-                    .arg("-s")
-                    .arg("-")
-                    .arg(existing_filename.to_str().unwrap())
+
+                let signing_status = Command::new("codesign")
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK dc4809ee032b378097c37e73e2929ceb9ee62c54

PR #311, opened 2025-07-28
Address all TODO in v25

jamillambert:
Go through all the TODO in the v25 types table. Add all the missing RPCs.

One missing RPC in v25, `scanblocks` which has 3 different return shapes, one each  for `abort`, `start` and `status`. Create a struct for each one.

`status` returns null if there is no ongoing scan, use an Option in the client macro to deal with this.


tcharding (approved):
ACK f6753625b1990182bd202c36c8cdd3069265e108

PR #317, opened 2025-08-01
Add support for run-bitcoind.sh start all

jamillambert:
On more than one occasion I have tried to start all nodes by passing `run all` or `start all` instead of just `all`.

Handle this the same as passing `all` as the argument to the script.


tcharding (approved):
ACK 25c8bf6f8ad7d1214db5b90d33d0b20504afe8f2

tcharding:
Nice, this bothered me a bunch too.

PR #323, opened 2025-08-05
Fix `getaddressinfo`

jamillambert:
The test was marked as TODO and the fields in the structs do not match what is returned by the RPC.

In v20 the return field `labels` changes from a vector of json objects that was modelled as `GetAddressInfoLabel` to a `Vec<String>`. To allow for this make the modelled type a `Vec<String>`, this means for v17-v19 the `name` field is modelled and the  `purpose` field of `GetAddressInfoLabel` is not.

- Fix all of the fields in v17.
- Redefine the types for changes in v18, v29, v22 and v28. 
- Update the model for to allow for the changes in return types.
- Update the test using different `Address` types to check the inner structs contain expected values.


types/src/model/wallet.rs
```diff
@@ -140,18 +148,20 @@ pub struct GetAddressInfo {
...
-    pub label: String,
+    pub label: Option<String>,
     /// The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT).
     pub timestamp: Option<u32>,
     /// The HD keypath if the key is HD and available.
     pub hd_key_path: Option<bip32::DerivationPath>,
     /// The Hash160 of the HD seed.
     pub hd_seed_id: Option<hash160::Hash>,
+    /// The fingerprint of the master key.
+    pub hd_master_fingerprint: Option<String>,
```
tcharding, line 162: The docs for this field are misleading, its only 4 bytes and is abstracted in `rust-bbitcoin` as `bip32::Fingerprint`.

```bash
t18 getaddressinfo $(bt18 getnewaddress)
{
  "address": "2Mv8t4zo6LZNV7C6h8Yw4BPxDGmRhN2QgRc",
  "scriptPubKey": "a9141fb5411701d330094d66b47429fbd1e60cc5383087",
  "ismine": true,
  "solvable": true,
  "desc": "sh(wpkh([8b864bde/0'/0'/1']02183110dabe2b4fc707584013ecc609e341c379328ab1c5e15f5324fa188e1efa))#qymhf8xd",
  "iswatchonly": false,
  "isscript": true,
  "iswitness": false,
  "script": "witness_v0_keyhash",
  "hex": "00142046f772984248fcce2cc899e9c423e1c12f7451",
  "pubkey": "02183110dabe2b4fc707584013ecc609e341c379328ab1c5e15f5324fa188e1efa",
  "embedded": {
    "isscript": false,
    "iswitness": true,
    "witness_version": 0,
    "witness_program": "2046f772984248fcce2cc899e9c423e1c12f7451",
    "pubkey": "02183110dabe2b4fc707584013ecc609e341c379328ab1c5e15f5324fa188e1efa",
    "address": "bcrt1qypr0wu5cgfy0en3vezv7n3pru8qj7az3kr7cv0",
    "scriptPubKey": "00142046f772984248fcce2cc899e9c423e1c12f7451"
  },
  "label": "",
  "ischange": false,
  "timestamp": 1754432922,
  "hdkeypath": "m/0'/0'/1'",
  "hdseedid": "33b74e94e3adec8c7c8f1158098c89846ca29273",
  "hdmasterfingerprint": "8b864bde",
  "labels": [
    {
      "name": "",
      "purpose": "receive"
    }
  ]
}
```

types/src/v17/wallet/into.rs
```diff
@@ -169,14 +169,18 @@ impl GetAddressInfo {
...
 
         Ok(model::GetAddressInfo {
             address,
             script_pubkey,
             is_mine: self.is_mine,
             is_watch_only: self.is_watch_only,
-            is_script: self.is_script,
+            solvable: None,          //v18 and above only.
+            descriptor: None,        //v18 and above only.
+            parent_descriptor: None, //v22 and above only.
```
tcharding, line 181: Typically code comments don't need to enforce too much style but since I'm the don in this crate I can get away with being even more anal than normal, can we have a space after the `//` please. Feel free to slap me :)

tcharding: Here is a one-liner to fix it codebase wide (from Claude). I tested it works.

```bash
find . -name "*.rs" -type f -exec sed -i 's|//v|// v|g' {} +
```

tcharding (commented):
Reviewed fc0a4867746e9902ecec304f49543be24c80be91

types/src/model/wallet.rs
jamillambert: Changed to `bip32::Fingerprint` and added the error variant.

jamillambert:
Suggested changes made. Also tidied up the imports in v18 `types` `wallet`.

tcharding (approved):
ACK 6cc26f076268d5d4a0350dd248ddd4d84d073e91

PR #326, opened 2025-08-08
Test finalizepsbt

jamillambert:
Fix the test for `finalizepsbt`.

The field `psbt` is optional for all versions, the help does not specifically say `optional` until v23, before that it says `if not extracted`.

Make `psbt` an `Option`.


tcharding (approved):
ACK a4d688e88caf2fa8d9a9638f7a1d75701748f9c0

PR #330, opened 2025-08-13
Test `listaddressgroupings`

jamillambert:
`listaddressgroupings` was untested.

- Add a test for it that returns a non empty list. But now with this test there is an error because "when Serde deserializes from a JSON array (a sequence), it uses positional fields and requires the sequence length to match the number of struct fields exactly" (AI quote). Which causes an error when the optional field `label` is not returned.

- Change the struct to an enum with 2 fields, one for 2 returned items and one for 3.

No changes to the RPC return fields up to v29.

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -484,6 +484,24 @@ fn wallet__keypool_refill() {
     let _: () = node.client.key_pool_refill().expect("keypoolrefill");
 }
 
+#[test]
+fn wallet__list_address_groupings__modelled() {
+    let node = Node::with_wallet(Wallet::Default, &[]);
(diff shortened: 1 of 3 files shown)


tcharding (approved):
ACK e76ab0de9771a0f2cdf4b83976bcb40e8acd38e0

tcharding:
Nice, you worked out the JSON shape, that stumped me.

jamillambert:
> Nice, you worked out the JSON shape, that stumped me.

Yeah it was a strange one. I assumed it would work like the others and using an Option would take care of the field being there or not. I had no idea Serde worked like that with a JSON array. AI first tried to write this massive deserialization function just for that one case, and I had to argue with it for a bit.

PR #331, opened 2025-08-13
Test `listlabels`

jamillambert:
`listlabels` was untested and had a model with no bitcoin types.

Add a test and remove the model.

Update verify, and types.


tcharding (dismissed):
ACK ebaa961e85bb031e68d2787a8d038d0f22e51929

integration_test/tests/wallet.rs
```diff
@@ -484,6 +484,17 @@ fn wallet__keypool_refill() {
...
 
+#[test]
+fn wallet__list_labels__modelled() {
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    let label = "list-label-test";
+    let _ = node.client.new_address_with_label(label).expect("newaddress");
+
+    let json: ListLabels = node.client.list_labels().expect("listlabels");
+
+    assert!(json.0.contains(&label.to_string()));
```
tcharding, line 495: I'm going to ack/merge this but in case it needs a rebase this line is a bit off. The `label` var is of type `&str`. This line allocates memory, creates an `String`, then immediately dereferences it to get an `&str`. So we can just pass in `label`.

tcharding: Ha I merged in order of PR number, got all the way to here!

jamillambert: > The `label` var is of type `&str`. This line allocates memory, creates an `String`, then immediately dereferences it to get an `&str`. So we can just pass in `label`.

So... I had to change it to `any()` with a closure. With `assert!(json.0.contains(label));` I get the error ```expected `&String`, found `&str` ``` which is why I changed it to the above originally. I didn't know that you could compare a `String` to `&str` with the `==` in the closure but not with `contains()`.

jamillambert:
I'm glad I had to fix this, the `&label.to_string()` did annoy me a little, now I know why it was needed and how not to do it.

Also first rebase with jj, had to do it twice, first time messed up the commands but going back with `jj op restore` worked a charm. Was easy in the end, and it supports using vscode as the merge editor.

i.e. I made the change and rebased to fix merge conflicts.

tcharding:
Mad, you are winning.

integration_test/tests/wallet.rs
tcharding: Bizzare, I thought I tested my suggestion yesterday but I guess I did something wrong (probably ran the `test17` command from crate root, thereby not _actually_ building the test).

And thanks, TIL about using `any` - its literally in the docs of `contains`!

tcharding (approved):
ACK 5a756eced52c804f63841f32fd3cdcf26807ad64

PR #334, opened 2025-08-19
Fix `getnodeaddresses` test

### LINKED ISSUE #320: `getnodeaddresses` returns an empty list
It is not possible to test the return shape of `getnodeaddresses` with the current test setup since an empty list is returned. 

Related to the issue https://github.com/bitcoin/bitcoin/issues/16990

jamillambert:
The return shape of the RPC is not tested since an empty list is returned.

- Add the hidden RPC `addpeeraddress` in v21 in a new module called `hidden`. 

- In the `getnodeaddresses` test use the hidden RPC `addpeeraddress` for v21 and later to add an address to the address manager table. Assert that the returned `GetNodeAddresses` contains the added peer address and port.

Closes #320


tcharding (approved):
ACK c9cf9406ab355964ed1924b31f1bb7fc2a8a1a0f

PR #338, opened 2025-08-20
Test `listreceivedbyaddress`

jamillambert:
The RPC was untested and missing a change in the return type in v18. 
`involves_watch_only` is always optional, the v17-v23 help says "only returned...", from v24 it specifically says optional.

- Update the struct in v18, make the field an Option and add a test.
- Run the formatter.


tcharding:
acked and merged 4 in a row no conflicts, top work bro.

PR #340, opened 2025-08-21
Test `walletcreatefundedpsbt` and `walletprocesspsbt`

jamillambert:
`walletcreatefundedpsbt` and `walletprocesspsbt` were implemented in v17 but untested. `walletcreatefundedpsbt` has  no return changes up to v29, `walletprocesspsbt` changes once in v26.

- Fix `walletcreatefundedpsbt` macro:
    The client macro passed an `Amount` to the RPC which parses to the
    wrong format resulting in 0 BTC for that address.
    
    The `inputs` struct has private fields and no constructor.
    
    Add a constructor for `WalletCreateFundedPsbtInput`.
    
    Convert the `Amount` in the `outputs` argument to an f64 in BTC before calling
    the RPC.
- Add a test for `walletcreatefundedpsbt` and update the types table.
- Update the `walletprocesspsbt` struct in v26, test and update the types table.
-  Run the formatter.

### DIFF
--- a/types/src/v26/mod.rs
+++ b/types/src/v26/mod.rs
@@ -271,8 +271,7 @@ pub use self::{
     },
     wallet::{
         CreateWallet, GetBalances, GetBalancesError, GetTransaction, GetTransactionError,
-        LastProcessedBlock, LastProcessedBlockError, LoadWallet, UnloadWallet,
-        WalletProcessPsbt,
+        LastProcessedBlock, LastProcessedBlockError, LoadWallet, UnloadWallet, WalletProcessPsbt,
     },
 };
 #[doc(inline)]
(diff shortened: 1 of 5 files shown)


client/src/client_sync/v17/wallet.rs
```diff
@@ -756,7 +765,11 @@ macro_rules! impl_client_v17__wallet_process_psbt {
     () => {
         impl Client {
             pub fn wallet_process_psbt(&self, psbt: &bitcoin::Psbt) -> Result<WalletProcessPsbt> {
-                self.call("walletprocesspsbt", &[into_json(psbt)?])
+                // Core expects the PSBT as a base64 string argument (same representation
+                // used by `finalizepsbt`). Serializing the struct with `into_json` produced
+                // an object which Core rejected ("Expected type string, got object").
```
tcharding, line 770: This is a known issue, I ran into it myself already. PSBT `serde` serialization does not use base64. I thought we had a tracking issue in `rust-bitcoin` but I couldn't find it just now.

This one kind of covers it: https://github.com/rust-bitcoin/rust-bitcoin/issues/765

PR #343, opened 2025-08-26
Test `getzmqnotifications`

jamillambert:
The RPC struct exists but there is no client macro or test. The return changes in v18 and then stays the same up to v29.

- Add an updated struct for v18. Add the client macro, test and update the types table.
- Run the formatter.


integration_test/tests/zmq.rs
```diff
@@ -0,0 +1,22 @@
...
+
+#[test]
+#[cfg(not(feature = "v17"))]
+fn zmq__get_zmq_notifications__modelled() {
+    // Start node with a ZMQ notification enabled so we have at least one entry.
+    // Using hashblock as it is lightweight.
+    let node = Node::with_wallet(Wallet::Default, &["-zmqpubhashblock=tcp://127.0.0.1:29000"]);
+
+    let list: Vec<GetZmqNotifications> = node.client.get_zmq_notifications().expect("getzmqnotifications");
+    let zmq_notification = &list[0];
```
tcharding, line 19: This will panic if the list is empty. While that does prove that its not, and since its a test, panic is ok for fail. It would be nicer to use `first` and call `unwrap` or `expect`.

jamillambert: I don't understand the difference. If I use `.first().unwrap()` it will panic on an empty array as well, so why add the extra two function calls when an empty array is a test fail anyway?

I get that if you want to handle an empty array then you need to use `first`, otherwise it just panics. But you are not suggesting doing anything with the `None` case for `first` are you?

Or is the reason just to use `expect` so that I can write the error message?

tcharding: FTR the array access on an empty list will look something like this

```bash
thread 'main' panicked at src/main.rs:4:21:
index out of bounds: the len is 0 but the index is 0
```

So it doesn't _really_ matter since that will give the line of the test that fails but my reasoning is as follows:

1. It saves the reader having to wonder _why_ the access is ok and what it means if it panics
2. Using `expect` with a message acts as both a code comment and makes debugging (slightly) easier

In general, any time I see an array access I immediately check the surrounding code to see if/why its ok, which takes time/effort. An overriding principle I like is to write code that minimises these sort of mental overhead on the reader since code is written once and read 1000's of times. More jovially, I like to think that code I write is going to be read by someone quietly having a cup of tea and it should be an easy enjoyable experience for them.

tcharding: The obvious problem with this mentality is that it is based on _some_ assumption of the readers ability and IQ - I've not worked out how to avoid this dilemma.

tcharding: cc @apoelstra in case you are interested in this armchair discussion. (He is longer in the tooth, when it comes to programming, than me.)

PR #345, opened 2025-08-27
Test and update `getwalletinfo`

jamillambert:
`getwalletinfo` was implemented for v17 only and untested. There are return changes in v18, v19, v21, v23 and v26.  

- Add the updated structs in all versions, update the model and add new  `into_model()`.  
- -Add a test with version specific sections to test the new sub-structs.  
- Update the types tables and reexports.

Draft because on top of #344, this PR is the last patch only.


jamillambert:
NB. the force push from today was from `jj` when pushing a commit for the next PR. Not sure why it did it, there are no changes.

jamillambert:
Rebased on master, no changes.

PR #346, opened 2025-08-27
Test `enumeratesigners`

jamillambert:
`enumeratesigners` is implemented but not tested. Also the v22 help is not correct, it gives incorrect field names which were implemented in the struct. They are correct in v23 and onwards help.

- Fix the v22 struct names.
- Add a test and update the types table.

### DIFF
--- a/integration_test/tests/signer.rs
+++ b/integration_test/tests/signer.rs
@@ -1,0 +1,35 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Tests for methods found under the `== Signer ==` section of the API docs.
+
+#![allow(non_snake_case)] // Test names intentionally use double underscore.
+#![allow(unused_imports)] // Because of feature gated tests.
+
+use integration_test::{Node, NodeExt as _, Wallet};
+use node::{mtype, Input, Output};
+use node::vtype::*;             // All the version specific types.
+
+#[test]
+#[cfg(unix)]
+#[cfg(not(feature = "v21_and_below"))]
+fn signer__enumerate_signers() {
+    use std::os::unix::fs::PermissionsExt;
+
+    let script_path = integration_test::random_tmp_file();
+    // `script_body` is minimal JSON array expected by `enumeratesigners` RPC: an array
+    // of signer objects with at least a fingerprint and name. Using a hard-coded
+    // dummy signer (fingerprint "deadbeef").
+    let script_body = "#!/bin/sh\necho '[{\"fingerprint\":\"deadbeef\",\"name\":\"TestSigner\"}]'\n";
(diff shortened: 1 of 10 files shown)


integration_test/tests/signer.rs
```diff
@@ -0,0 +1,31 @@
...
+	let script_path = integration_test::random_tmp_file();
+	let script_body = "#!/bin/sh\necho '[{\"fingerprint\":\"deadbeef\",\"name\":\"TestSigner\"}]'\n";
+	std::fs::write(&script_path, script_body).expect("write signer script");
+    
+	#[cfg(unix)]
+	{
+		use std::os::unix::fs::PermissionsExt;
+		std::fs::set_permissions(&script_path, std::fs::Permissions::from_mode(0o755))
+			.expect("chmod");
+	}
```
tcharding, line 24: A few comments:

- This using tabs, can we use spaces please (tabs vs spaces is a age old religious war - I'm on team spaces).
-  Line 18 has trailing whitespace
- Perhaps add a code comment saying "Script needs to be executable" or some such thing
- Perhaps a code comment on the `script_body` - I have no idea why the script is as it is?
- Its not clear to me what happens if this test is run on non-unix machine. I don't actually care about non-unix machines though, maybe put the `cfg` on the whole test to make it clear we only support UNIX for this test.

integration_test/tests/signer.rs
```diff
@@ -0,0 +1,31 @@
...
+
+	let signer_arg = format!("-signer={}", script_path.to_str().unwrap());
+	let node = Node::with_wallet(Wallet::None, &[&signer_arg]);
+	let json: EnumerateSigners = node.client.enumerate_signers().expect("enumeratesigners");
+
+    assert_eq!(json.signers[0].fingerprint, "deadbeef");
```
tcharding, line 30: As in other comment, better to use `first` than the array access. In idiomatic Rust array access is only used if there is a good reason to do so.

integration_test/tests/signer.rs
jamillambert: > * Its not clear to me what happens if this test is run on non-unix machine. I don't actually care about non-unix machines though, maybe put the `cfg` on the whole test to make it clear we only support UNIX for this test.

I feel like the tests are mainly for CI to check that any code changes or changes to the RPCs are ok, so it's not that important it works on other systems. On file systems like windows the test should run fine since you don't need to set the execute permission, but I moved the `cfg(unix)` to the top because there is no guarantee that it will work on non unix/windows systems.

integration_test/tests/signer.rs
jamillambert: I have changed it to first and expect.

tcharding:
Nice, the test is great now.

PR #347, opened 2025-08-28
Test and update `listsinceblock`

jamillambert:
The RPC was implemented for v17 and untested, and the struct is not exactly what is returned by the RPC in v17 since some of the fields are optional or undocumented.

- In preparation for changes (Code move only): 
   - Move v25 wallet to subdirectory
   - Move `wallet.rs` to `./wallet/mod.rs`
   - Split out the `into_model()` into an `into` module. 

- Fix the struct for v17. Update the struct, model, into_model, and error for changes in v20, v23, v24 and v28. Add a test and update the types tables.

Draft because on top of #345, this PR is the last two patches.


types/src/v17/wallet/mod.rs
```diff
@@ -21,8 +21,8 @@ pub use self::error::*;
 //
 // The following structs are very similar but have slightly different fields and docs.
 // - GetTransaction
-// - ListSinceBlockTransaction
-// - ListTransactionsItem
+// - TransactionItem
+// - TransactionItem
```
tcharding, line 25: Duplicate, typo?

PR #352, opened 2025-09-08
Fix verify

jamillambert:
#351 described the problem

- During the updating of the RPCs some of them changed from having a
model to not or vice versa. There is one typo. There is also a duplicate entry for `enumeratesigners` which was added to `signer` and not removed from `wallet`.
   - Fix the tags in `verify` to match the correct RPC method definitions.
 - There were false positives in verify when the RPC name was a substring
of another identifier.
   - Change the regex to match whole words.
- `verify` does not fail when there are detected problems, it prints
`Correct ✓` and continues.
   - Change the tool so that it prints `Incorrect ✗` if there are
inconsistencies.
   - Run all tests to the end and then if there were any errors, return a
non-zero exit code.


tcharding:
Nice one fella.

PR #353, opened 2025-09-09
Improve `types` struct and enum rustdocs

jamillambert:
The rustdocs for the RPCs in types in inconsistent. In particular the sub structs and enums don't always mention which RPC they are part of. This makes it harder to understand which bit goes with which.

Use the same language throughout so that when scanning through the code it is easier to follow.

Make all the rustdocs consistent and mention which RPC they are part of.


tcharding:
Mad. Did you do this manually or were you able to get an LLM to do it?

PR #356, opened 2025-09-10
Update `just fmt` to format test and verify

jamillambert:
Running `just fmt` only formats modules in the workspace, and not `integration_test` and `verify`.

- Add a line for both of these modules so that they are formatted when running `just fmt`.
- Format `integration_test`.
- Format `verfiy`.


tcharding (approved):
ACK b6f827bc8ffcf5debb2d843dda531af87a74aad6

tcharding:
Needs rebase, I merged #358 first because I thought it would be easier to rebase this one. Tip, when you rebase just reset the index and then run the formatter again. That way you should keep the commit log and patch series structure.

PR #358, opened 2025-09-10
Update verify to check for model vs no model

### LINKED ISSUE #144: Verify modelled vs no model matches
After #142 merges add logic to `verify` to check that `Method::new_modelled` matches 'version + model' in 'Returns' section and that `Method::new_no_model` matches with 'version'.

jamillambert:
Verify does not correctly check that the RPCs state if they are modelled or not consistently between `verify` and the `types` table.
Which meant that a lot are different between the two.

- Go through all of them and make them both match what is done for that RPC.
- Update `verify` to check that modelled vs no model matches.

Closes #144


tcharding (approved):
ACK 42fc2c4dbd09291aae95e170c9c02e2119f45501

tcharding:
Nice one man.

Issue #361, opened 2025-09-11
`just update-lock-files` hangs


tcharding:
Probably because of #147. We shouldn't use all features but probably just cd into each crate and build as appropriate. 

FTR to update the lock files I built `node` using  `cargo check --features=v29_0`

tcharding:
Oh interesting, well that is good at least, if its just something in my environment.

tcharding:
ooph, no clue then.

PR #362, opened 2025-09-11
Fix build system hanging due to conflicting version features

0xZaddyy:
This PR addresses #361 
The `just update-lock-files` command was hanging because it used `--all-features` which enabled mutually exclusive version features in the node crate simultaneously (29_0, 28_2, 28_1, etc.), causing Cargo to get stuck resolving impossible dependencies.

  **Changes:**
  - Build each crate individually instead of using workspace `--all-features`  
  - Use `features=29_0` for node crate (latest supported version)
  - Handle excluded crates (verify) with `cd` commands
  - Apply same pattern to all build commands (build, check, lint, docsrs)

### DIFF
--- a/justfile
+++ b/justfile
@@ -12,17 +12,23 @@ default:
 
 # Cargo build everything.
 build:
-<<<<<<< HEAD
   cargo build --workspace --all-targets --all-features
+  cargo build -p corepc-client --all-targets --all-features
+  cargo build -p corepc-types --all-targets --all-features
+  cargo build -p jsonrpc --all-targets --all-features
+  cargo build -p corepc-node --all-targets --features=latest
+  cargo build --manifest-path verify/Cargo.toml --all-targets
 
 # Cargo check everything.
 check:
-  cargo check --workspace --all-targets --all-features
+  cargo check -p corepc-client --all-targets --all-features
+  cargo check -p corepc-types --all-targets --all-features
+  cargo check -p jsonrpc --all-targets --all-features
+  cargo check -p corepc-node --all-targets --features=latest
+  cargo check --manifest-path verify/Cargo.toml --all-targets
 
 # Lint everything.
 lint: lint-verify lint-integration-tests
-  cargo +$(cat ./nightly-version) clippy --workspace --all-targets --all-features -- --deny warnings
-=======
   cargo build -p corepc-client --all-targets --all-features
   cargo build -p corepc-types --all-targets --all-features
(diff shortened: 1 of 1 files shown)


tcharding:
Thanks for the contribution!

contrib/update-lock-files.sh
```diff
@@ -6,6 +6,13 @@ set -euo pipefail
...
     cp --force "$file" Cargo.lock
-    cargo check --all-features
+    # Build each crate individually to avoid conflicting features
+    (cd client && cargo check --all-features)
+    (cd types && cargo check --all-features)
+    (cd jsonrpc && cargo check --all-features)
+    (cd node && cargo check --features=29_0)
+    # verify crate is excluded from workspace, handle separately
+    (cd verify && cargo check)
+    # Skip integration_test as it cannot use --all-features
```
tcharding, line 16: `integration_test` can be checked the same as `node` (and for the same reason).

What about using lists of crate names and then a for loop, might be cleaner? I don't see the reason to have `verify` separate.

justfile
```diff
@@ -12,15 +12,27 @@ default:
...
+  (cd verify && cargo build --all-targets)
 
 # Cargo check everything.
 check:
-  cargo check --workspace --all-targets --all-features
+  cargo check -p corepc-client --all-targets --all-features
+  cargo check -p corepc-types --all-targets --all-features
+  cargo check -p jsonrpc --all-targets --all-features
+  cargo check -p corepc-node --all-targets --features=29_0
+  (cd verify && cargo check --all-targets)
```
tcharding, line 27: We could use `cargo check --manifest-patch verify/Cargo.toml` instead to save cd'ing into `verify`.

tcharding:
I like the `latest` feature idea.

jamillambert (commented):
Looks good, but can you change the order of the two commits so that the latest feature is added to the `Cargo.toml` file first and then in the second commit you can use `latest` in the changes to the scripts. Otherwise you set it to 29.0 then immediately change it to latest.

0xZaddyy:
Hey @jamillambert good now?

tcharding:
You still have mereg conflicts in patch 1 mate.

jamillambert (commented):
@0xZaddyy Are you happy for me to take this over and make the last few changes?

tcharding:
Its been a while now, feel free to pick this up @jamillambert if you feel like it.

jamillambert:
Closed in favor of #464

PR #364, opened 2025-09-12
Type alias `SignRawTransaction`

### LINKED ISSUE #357: `signrawtransactionwithkey` has no type or model with the same name
The RPC `signrawtransactionwithkey` replaced `signrawtransaction` in v18 and above. It still uses the same Type name of `SignRawTransaction` which causes an issue with `verify` and when manually searching the code for the type or model.

I would suggest renaming it everwhere to `SignRawTransactionWithKey`, similarly with the test and macro name. And keeping a duplicate of the v17 ones with the current name, it's the only version that has both RPCs

EDIT: I fixed verify and there is no issue with this RPC since it only checks the line in verify vs the types table, but I still prefer the solution above to how it is now.

jamillambert:
Both `signrawtransactionwithkey` and `signrawtransactionwithwallet` use the type `SignRawTransaction` in types and model. But the policy is to have a type for all RPCs with the same name, this makes everything uniform and easier to use without having to search for the correct type.

- Add aliases for these two RPC methods and a note in the docs explaining why.
- Use the new aliases in the tests.
- Run the formatter. Reordering reexports only.

Closes #357


tcharding (approved):
ACK 5b3fbcd455ed31c656c6241132161a919cfec113

PR #368, opened 2025-09-26
Change script_buf and address methods to public

benthecarman:


tcharding:
Thanks man. I very nearly roasted you for empty PR description and git log because we get so many LLM generated PRs in `rust-bitcoin` these days. Then I saw your username :)

I created, and would prefer to fix, #369.

PR #374, opened 2025-10-02
Remove `serde-deny-unknown-fields` from `model` types

### LINKED ISSUE #373: `serde-deny-unknown-fields` pointless in `model`
The deny unknown fields attribute was added so that in testing we catch mistakes in what Core returns and what we expect to get. In `model` we derive `serde` traits so that users of the lib can persist the types if they want to, the `model` types are never used directly with Core. It does not make sense to have the deny unknown fields attribute in `model`.

100% my mistake for not noticing this during review when we added the attribute, and then again later when we feature gated the attribute.

jamillambert:
The attribute was added to catch mistakes in what core returns compared to what is in the structs. This has no meaning in `model` since they are internal conversions of the version specific types.

Remove the attribute from all `model` types.

Closes #373

### DIFF
--- a/types/src/model/blockchain.rs
+++ b/types/src/model/blockchain.rs
@@ -19,7 +19,6 @@ use crate::ScriptPubkey;
 
 /// Models the result of JSON-RPC method `dumptxoutset`.
 #[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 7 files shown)


tcharding (approved):
ACK 899b08ed6ae09c4abec4b9affe78c7dc9217ecfb

PR #376, opened 2025-10-02
feat(types): add new types for inputs/outputs and options

storopoli:
These types are needed to finalize the migration of [`bitcoind-async-client`](https://github.com/alpenlabs/bitcoind-async-client) to use `corepc-types`.

They are currently being used by Alpen Labs and Payjoin Dev Kit.

I put them in `v29` for now, but let me know if we need to put them in previous versions.

### DIFF
--- a/types/src/v29/mod.rs
+++ b/types/src/v29/mod.rs
@@ -255,6 +255,7 @@ mod blockchain;
 mod mining;
 mod raw_transactions;
 mod util;
+mod wallet;
 
 #[doc(inline)]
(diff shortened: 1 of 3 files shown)


storopoli:
Ok closing this because my main use case was to implement in an async client and get rid of my custom types.

tcharding:
These types should stay in your client.

tcharding:
> I think they should go in `/client/src/client_sync/vXX/mod.rs`

He isn't using our client.

PR #377, opened 2025-10-06
Fix clippy `doc list item without indentation`

jamillambert:
Clippy gives `error: doc list item without indentation`.

Add a list item with a description on where to find the missing argument information to fix the error.


jamillambert:
NB. This came up when making changes to `rust-bitcoin-maintainer-tools` to add support for `extra_lints.sh`.

PR #378, opened 2025-10-06
Bump versions to 0.10.0

tcharding:
In preparation for release add a changelog entry, bump the version, and update the lockfiles.

This is for all three released crates. FTR because they depend on each other release of `types` triggers release of the others too (if I'm not mistaken).


jamillambert (approved):
ACK 2e71a27e7a1664715b4045f7cf5da187a9c8c515

PR #387, opened 2025-10-15
Add support for Bitcoin Core 30.0

jamillambert:
Add all the necessary files and changes to support Core v30.

RPCs that were dropped in v30 have been removed.

New and changed RPCs between v29 and v30 have not been implemented in this PR and are TODO in `types`.

Patches are:
- Remove duplicate methods in verify.
- Add all the required files, modules and features to support v30. `verify/src/method/v30.rs` is a copy of v29 with the label changed to v30. `types/src/v30/mod.rs` is a copy of v29 with the exports of the new v29 types moved to a reexport section under v29, no other changes.
- Edit the files copied from v29 in a separate patch to make reviewing the changes easier. Feature gate the tests for RPCs that were removed or changed in v30. The new and changed RPCs are left as TODO in the types table for a follow up PR or patch.


integration_test/Cargo.toml
```diff
@@ -14,7 +14,8 @@ edition = "2021"
 download = ["node/download"]
 
 # Enable the same feature in `node` and the version feature here.
-# All minor releases of the latest three versions.
+# All minor releases of the latest four versions.
```
tcharding: Novel idea, increase this number so as not to have to make changes to the 4th oldest release. Are you planning on dropping this back to 3 or do you think we leave it as is for now?

tcharding:
F**k you learn fast. Really nice separation of concerns in this PR. By that I mean what to change and what to leave for later.

integration_test/Cargo.toml
jamillambert: > Its a policy thing, if we keep this and 27.x comes out we have to add support for it. 

The comment below says `# Only the latest minor version for older versions.` which would suggest we have to anyway. I guess we don't want this for the older versions and should change the wording.

PR #388, opened 2025-10-16
Add `waitforblock`, `waitforblockheight` and `waitfornewblock`

jamillambert:
`waitforblock`, `waitforblockheight` and `waitfornewblock` were added to the Core API docs in v30 under blockchain. The RPCs have been there since before v17.

Add the RPCs to v17, and add to the types table in v30.

- The hidden module in v21 had a copy-paste error and stated it was for the `generating` section. 
  - Update the documentation to state that the methods are `hidden`.
- `blockchain__get_tx_out_proof__modelled` is identical to `blockchain__verify_tx_out_proof__modelled`, and the method `gettxoutproof` is tested above in `blockchain__get_tx_out_proof`. 
  - Remove the redundant test.
- The model part of the `verify_tx_out_proof` test was in it's own function that the test function called. Other tests all check the model in the same test. 
  - Combine them into one function.
  - Place it in the correct location.
  - Remove associated import.
- Add `waitforblock`, `waitforblockheight` and `waitfornewblock` methods to v17. They are hidden until v30 with no changes.

### DIFF
--- a/client/src/client_sync/v17/hidden.rs
+++ b/client/src/client_sync/v17/hidden.rs
@@ -1,0 +1,46 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Macros for implementing JSON-RPC methods on a client.
+//!
+//! Specifically this is `== Hidden ==` methods that are not listed in the
+//! API docs of Bitcoin Core `v0.17`.
+//!
+//! All macros require `Client` to be in scope.
+//!
+//! See or use the `define_jsonrpc_minreq_client!` macro to define a `Client`.
+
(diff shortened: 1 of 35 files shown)


jamillambert:
Converting to draft to fix CI test fail that does not happen locally

jamillambert:
The three RPCs have the same return shape. I made 3 seperate types since they only have a block hash and height in them it made no sense to have a seperate internal type that only has those 2 values in it. I considered having one type e.g. `WaitForBlock` and the other two use that, but then the type name doens't match the RPC name.

tcharding:
Needs rebase please mate.

Issue #393, opened 2025-10-21
bitreq: Add json support


tcharding:
Add back in the `json-using-serde` feature from `minreq`. Required so that `jsonrcp` can depend on `bitreq`.

PR #398, opened 2025-10-26
bitreq: Re-implement `json-using-serde` feature

tcharding:
This feature was removed when we gutted `minreq` after forking it. Turns out we need it to be able to use `bitreq` in `jsonrpc`.

Re-implement the feature by copying code from `minreq`.

Close: #393

### DIFF
--- a/bitreq/src/response.rs
+++ b/bitreq/src/response.rs
@@ -127,6 +127,41 @@ impl Response {
     /// # fn main() -> Result<(), Box<dyn std::error::Error>> { Ok(()) }
     /// ```
     pub fn into_bytes(self) -> Vec<u8> { self.body }
+
+    /// Converts JSON body to a `struct` using Serde.
+    ///
+    /// # Errors
+    ///
+    /// Returns
+    /// [`SerdeJsonError`](enum.Error.html#variant.SerdeJsonError) if
+    /// Serde runs into a problem, or
+    /// [`InvalidUtf8InBody`](enum.Error.html#variant.InvalidUtf8InBody)
+    /// if the body is not UTF-8.
+    ///
+    /// # Example
+    /// In case compiler cannot figure out return type you might need to declare it explicitly:
+    ///
+    /// ```no_run
+    /// use serde_json::Value;
+    ///
+    /// # fn main() -> Result<(), bitreq::Error> {
+    /// # let url_to_json_resource = "http://example.org/resource.json";
+    /// // Value could be any type that implements Deserialize!
+    /// let user = bitreq::get(url_to_json_resource).send()?.json::<Value>()?;
+    /// println!("User name is '{}'", user["name"]);
+    /// # Ok(())
+    /// # }
+    /// ```
+    #[cfg(feature = "json-using-serde")]
+    pub fn json<'a, T>(&'a self) -> Result<T, Error>
+    where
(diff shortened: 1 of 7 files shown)


bitreq/src/response.rs
```diff
@@ -127,6 +127,45 @@ impl Response {
...
+    /// [`InvalidUtf8InBody`](enum.Error.html#variant.InvalidUtf8InBody)
+    /// if the body is not UTF-8.
+    ///
+    /// # Example
+    /// In case compiler cannot figure out return type you might need to declare it explicitly:
+    ///
+    /// ```no_run
+    /// use serde_json::Value;
+    ///
+    /// # fn main() -> Result<(), minreq::Error> {
```
jamillambert:
```suggestion
    /// # fn main() -> Result<(), bitreq::Error> {
```

bitreq/tests/main.rs
```diff
@@ -14,6 +14,21 @@ fn test_https() {
...
+    const JSON_SRC: &str = r#"{
+        "str": "Json test",
+        "num": 42
+    }"#;
+
+    setup();
+    let original_json: serde_json::Value = serde_json::from_str(JSON_SRC).unwrap();
+    let response = bitreq::post(url("/echo")).with_json(&original_json).unwrap().send().unwrap();
+    let actual_json: serde_json::Value = response.json().unwrap();
+    assert_eq!(&actual_json, &original_json);
```
jamillambert:
```suggestion
    assert_eq!(actual_json, original_json);
```

bitreq/src/response.rs
```diff
@@ -127,6 +127,45 @@ impl Response {
...
+    where
+        T: serde::de::Deserialize<'a>,
+    {
+        let str = match self.as_str() {
+            Ok(str) => str,
+            Err(_) => return Err(Error::InvalidUtf8InResponse),
+        };
+        match serde_json::from_str(str) {
+            Ok(json) => Ok(json),
+            Err(err) => Err(Error::SerdeJsonError(err)),
```
jamillambert:
```suggestion
        serde_json::from_str(self.as_str()?)
            .map_err(Error::SerdeJsonError)
```

The existing code returns the incorrect error variant. The docs say `InvalidUtf8InBody` which is what `as_str()` returns.

bitreq/src/response.rs
```diff
@@ -127,6 +127,45 @@ impl Response {
...
+    /// # Example
+    /// In case compiler cannot figure out return type you might need to declare it explicitly:
+    ///
+    /// ```no_run
+    /// use serde_json::Value;
+    ///
+    /// # fn main() -> Result<(), minreq::Error> {
+    /// # let url_to_json_resource = "http://example.org/resource.json";
+    /// // Value could be any type that implements Deserialize!
+    /// let user = minreq::get(url_to_json_resource).send()?.json::<Value>()?;
```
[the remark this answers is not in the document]
jamillambert: Why are these not caught in CI? I get an error locally. Ah, is it because of #400 not being merged yet?

bitreq/src/response.rs
tcharding: Half right I think. Your suggestion is correct and better but AFAICT it doesn't change the logic. I used the suggestion.

PR #399, opened 2025-10-26
jsonrpc: Depend on `bitreq`

tcharding:
We just forked `minreq` into this repository to create an HTTP crate that is specifically designed to be used by Bitcoin projects. As such we called it `bitreq`.
    
Depend on the local `bitreq` crate instead of `minreq`.


jamillambert:
Could address #406 as part of this PR. The two cases left I can see are:
`bitreq/src/lib.rs:1` `//! # Minreq`
and `MINREQ_TIMEOUT`

PR #405, opened 2025-10-28
Bump MSRV to 1.75.0

### LINKED ISSUE #401: Update MSRV
The decision at the Rust Bitcoin Summit was to have an `MSRV` in `corepc` of `1.75.0`. 

`rust-bitcoin` is currently at `1.74.0` but there is a change in `1.75.0` that is relevant to `corepc` mentioned at the summit, I think it is:  
- [Stabilize async fn and return-position impl Trait in traits.](https://github.com/rust-lang/rust/pull/115822/)

Opening the issue to confirm this is correct before doing it.

jamillambert:
A consensus was reached at the 2025 Rust Bitcoin Summit to use an MSRV of 1.75.0 for corepc and all its crates.

Closes #401


tcharding (approved):
ACK a7d68bfd61efef5c39f4429fc87b22c17df79819

PR #409, opened 2025-11-03
Implement v30 changes to `getmininginfo` and `getmempoolinfo` RPCs

jamillambert:
Redefine the `getmininginfo` and `getmempoolinfo` types in v30 for the changes and remove TODO from table.

Update the exports, into functions and test feature gates.

PR #410, opened 2025-11-03
Implement v30 changes to `decodepsbt` RPC

jamillambert:
`decodepsbt` adds new musig fields in v30. These cannot be currently modelled with `rust-bitcoin` types and this PR does not change the model type. (https://github.com/rust-bitcoin/rust-bitcoin/issues/4207).

The error types from v17 were not reexported and instead imported by referencing the v17 module in the import statement.

- Tidy up the error exports and imports to reexport from v17 and then import from `super` instead of `crate::v17`.
- In order to make the review easier first copy the previous version of `decodepsbt` and all the related sub types and errors from v24 to v30. This patch copies to code and updates the exports only.
- Add the musig fields in a separate patch to make the changes from v24 clear.

### DIFF
--- a/types/src/v30/mod.rs
+++ b/types/src/v30/mod.rs
@@ -128,7 +128,7 @@
 //! | converttopsbt                      | version + model |                                        |
 //! | createpsbt                         | version + model |                                        |
 //! | createrawtransaction               | version + model |                                        |
-//! | decodepsbt                         | version + model | TODO                                   |
+//! | decodepsbt                         | version + model | Musig not modelled: not in rust-bitcoin|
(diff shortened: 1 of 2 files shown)


jamillambert:
This is a Draft to get feedback on the way it has been done, and the merge conflicts which will come after #409 goes in. 
Specifically:
- All of the error types have been redefined even though the model and into functions have not changed. This was to make it simpler in the future when the model will be updated and the errors also need to also be expanded to include the musig fields.
- Not updating the model at all. The current model is just `psbt: Psbt`  and `fee: Option<Amount>`. Adding extra fields didn't make sense.

PR #412, opened 2025-11-07
Implement RPC changes in v30

### LINKED ISSUE #384: Support Bitcoin Core v30.0
The changes in terms of RPC calls and return types don't seem to be that large. From running a script to compare it to v29:

EDIT: updated below based on further testing

Added (Already there since v17 but hidden #333):

- Blockchain
  - waitforblock
  - waitforblockheight
  - waitfornewblock

Removed:
- addmultisigaddress
- dumpprivkey
- dumpwallet
- getunconfirmedbalance
- importaddress
- importmulti
- importprivkey
- importpubkey
- importwallet
- newkeypool
- sethdseed
- upgradewallet

Changed:
- Blockchain
  - getmempoolinfo
- Mining
  - getmininginfo
- Rawtransactions
  - decodepsbt
- Wallet
  - getwalletinfo
  - migratewallet
  - listwalletdir
  - settxfee

jamillambert:
There are four remaining RPCs that have changes in v30 that need to be implemented:

- `settxfee` is deprecated in v30 and removed in v31. Feature gate the test to v30 and below and for v30 add the required `-deprecatedrpc=settxfee` flag. Remove TODO from the types table.
- In v30 it is no longer possible to create a legacy wallet. This means that `migratewallet` cannot be tested using only v30. There are no changes to the RPC in v30 so the v29 test is still valid. Add a comment to the types table and test.
- `getwalletinfo` had some return fields removed in v30 and a new flags field was added. Update the model to make the removed fields options and add the new flags. Update all the into functions. Redefine the type, error and into function for v30. Remove the TODO from the types table. Remove the `v29_and_below` feature gate from the test.
- `listwalletdir` has a new `warnings` field was added in v30. Redefine the type for v30, there is no model. Remove the TODO from the types table. Remove the `v29_and_below` feature gate from the test.

Together with #387, #388, #409 and #410 Closes #384

### DIFF
--- a/integration_test/tests/wallet.rs
+++ b/integration_test/tests/wallet.rs
@@ -814,7 +814,7 @@ fn wallet__list_unspent__modelled() {
 }
 
 #[test]
-#[cfg(all(feature = "v29_and_below", not(feature = "v17")))]
+#[cfg(not(feature = "v17"))]
 fn wallet__list_wallet_dir() {
     let wallet_name = "test-wallet";
     let node = Node::with_wallet(Wallet::None, &[]);
--- a/types/src/v30/mod.rs
+++ b/types/src/v30/mod.rs
@@ -206,7 +206,7 @@
 //! | listsinceblock                     | version + model |                                        |
 //! | listtransactions                   | version + model |                                        |
 //! | listunspent                        | version + model |                                        |
-//! | listwalletdir                      | version         | TODO                                   |
+//! | listwalletdir                      | version         |                                        |
 //! | listwallets                        | version + model |                                        |
 //! | loadwallet                         | version + model |                                        |
 //! | lockunspent                        | version         |                                        |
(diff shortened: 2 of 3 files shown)


tcharding (approved):
ACK 2343534f22a63e5b65a8dc493bfb3d4c88a6f75c

tcharding:
Nice work man. This PR highlights an interesting thing about the version specific types (we really need a shorter name for those) and the model types. As the library progresses to support new versions of Core users using the model types face breakage when they upgrade but users of the version specific types (vs-types?) do not. I.e., the vs-types should be stable across releases except for bug fixes but the model types are perma-unstable. Interesting, I never thought of it like this before.

tcharding:
Interesting idea, doesn't help with when Core adds new fields though. We could have `non_exhaustive` but I think that both these 'solutions' make the crate to annoying to use. Better just to be perma-unstable I rekon. Lets wait and see if any user ever asks us wen-1.0

PR #420, opened 2025-11-24
verify: Add a check that re-exports in `types` are complete

### LINKED ISSUE #137: Test that re-exported `types` auxiliary types are correct

jamillambert:
Add a new check to `verify` that goes through every public struct or enum defined in the specified version in `types` and makes sure that it is exported for that version. This includes auxiliary types and error enums.

It then checks that all auxiliary types that are a field of one of the re-exported types from a previous version are also re-exported.

The first patch adds all the missing re-exports.

The second patch adds the test to `verify`.

Closes #137

### DIFF
--- a/verify/Cargo.toml
+++ b/verify/Cargo.toml
@@ -7,3 +7,5 @@ edition = "2021"
 anyhow = "1.0.93"
 clap = { version = "4.5.23", features = ["cargo"] }
 regex = "1"
+syn = { version = "2.0", default-features = false, features = ["full", "parsing"] }
+walkdir = "2.5"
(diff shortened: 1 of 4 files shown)


tcharding (approved):
tACK 0bbbfbb2855d06043042fb33fe924bc13d575bf9

Sick.

tcharding:
This is mad, id you enjoy hacking it up?

PR #421, opened 2025-11-25
bitreq: Remove unused tokio features

jamillambert:
There are tokio features that are not used in the crate, or only used in an example.

Update the example so that it doesn't require any additional tokio features that the rest of the crate doesn't use.
 
Remove the unused tokio features. Addresses one of the points in #402.

Update the lock files.


tcharding (approved):
ACK 61518d53b674a3dca1796b9e64fd46ef1727cec8

PR #430, opened 2025-12-16
fix(types): use f64 for GetChainTxStats tx_rate field

deadmanoz:
Bitcoin Core has always returned `txrate` as a double since the RPC was added in v0.15 (PR #9733).

The previous type (`i64`) causes deserialisation to fail. Here's a minimal example:

```rust
use corepc_client::client_sync::v28::Client;
use corepc_client::client_sync::Auth;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let auth = Auth::UserPass("bitcoin".to_string(), "bitcoin".to_string());
    let client = Client::new_with_auth("http://127.0.0.1:8332", auth)?;
    let stats = client.get_chain_tx_stats()?;
    println!("tx_rate: {:?}", stats.tx_rate);
    Ok(())
}
```

```
Error: JsonRpc(Json(Error("invalid type: floating point `5.086614477523837`, expected i64", line: 1, column: 273)))
```

### DIFF
--- a/integration_test/tests/blockchain.rs
+++ b/integration_test/tests/blockchain.rs
@@ -218,11 +218,15 @@ fn blockchain__get_chain_tips__modelled() {
 
 #[test]
 fn blockchain__get_chain_tx_stats__modelled() {
-    let node = Node::with_wallet(Wallet::None, &[]);
+    let node = Node::with_wallet(Wallet::Default, &[]);
+    node.fund_wallet();
+    let (_address, _tx) = node.create_mined_transaction();
 
     let json: GetChainTxStats = node.client.get_chain_tx_stats().expect("getchaintxstats");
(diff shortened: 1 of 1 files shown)


deadmanoz:
Updated existing test.

Switced to a `Default` wallet, and fund/mine so we get transactions to report meaningful data for assertion, e.g.

```rust
GetChainTxStats { 
  time: 1766036241,
  tx_count: 104,
  window_final_block_hash: 5d1f1967e8c133164a25a51f9bb125e79c8d2395baaabde8c3767b0220442a66,
  window_final_block_height: Some(102), 
  window_block_count: 101, 
  window_tx_count: Some(102), 
  window_interval: Some(17), 
  tx_rate: Some(6.0)
}
```

PR #438, opened 2025-12-29
Fix fuzz

### LINKED ISSUE #437: fuzz is broken
The daily fuzz action fails because it's not in the workspace. When you add it to the workspace it doesn't build.

jamillambert:
The fuzz module does not build and the fuzz action fails in CI.

- Fix the incorrect `jsonrpc` dependency path and add the missing `cfg(jsonrpc_fuzz)` to `unexpected_cfgs`.
- Update the rustc version in the action to 1.75 to fix the fuzz job error.
- Update the lockfiles.
- Add `fuzz` to the workspace.

Closes #437


tcharding (approved):
ACK 154c994ae8ec4d59e4199316915eb53486ff6ad3

PR #435, opened 2025-12-22
add hidden RPC `getorphantxs`

0xB10C:
> getorphantxs ( verbosity )
> Shows transactions in the tx orphanage.

This implements the hidden `getorphantxs` RPC in the last commit. Depends on #432 and #433. Part of #333. 

Since the `expiry` and `entry` fields were dropped for v30 (this is an experimental RPC after all), there is a `into.rs` implementation for both v29 and v30.

PR #439, opened 2025-12-30
Add hidden RPC `getrawaddrman`

deadmanoz:
Add support for the hidden `getrawaddrman` RPC that returns raw address manager table entries (added v26). This RPC is marked as experimental and hidden from the API docs by default.

The `mapped_as` and `source_mapped_as` fields are `Option<u32>` following the established pattern for version-specific fields (no `into.rs` required because these are additions rather than removals or changed types). These fields were added in Bitcoin Core v28 and will deserialise as `None` for earlier versions or when `-asmap` is not configured.

Note: will have conflicts with https://github.com/rust-bitcoin/corepc/pull/435 and https://github.com/rust-bitcoin/corepc/pull/433, happy to rebase after either merges


deadmanoz:
Ok so I learned that only the public RPCs should be added to the rustdoc table in the `vX/mod.rs`!

types/src/v26/hidden.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Addresses in the "new" table (potential peers discovered but not yet connected to).
+    pub new: BTreeMap<String, RawAddrmanEntry>,
+    /// Addresses in the "tried" table (peers successfully connected to in the past).
+    pub tried: BTreeMap<String, RawAddrmanEntry>,
+}
+
+/// An entry in the address manager table. Part of `getrawaddrman`.
+///
+/// The key in the parent map is formatted as "bucket/position" indicating the
+/// location in the relevant address manager table.
```
0xB10C: This key is a bit weird (not in your implementation, but in the Bitcoin Core implement). IIRC, it was done this way, as the key should be seen as Bitcoin Core implementation specific. However, RPC users might want to consume and act on it. I guess probably best to write a parser for it in downstream apps and not here.

deadmanoz:
> Looks good to me!
> 
> https://github.com/0xB10C/addrman-observer still uses a custom rust-bitcoincore-rpc `getrawaddrman` implementation. Would be nice to replace that with this PR! I might do this to test this PR.

I was going to propagate the change once all good here, but go for it!

types/src/v26/hidden.rs
```diff
@@ -0,0 +1,60 @@
...
+///
+/// Field order matches Bitcoin Core's RPC response definition.
+#[derive(Clone, Debug, PartialEq, Deserialize, Serialize)]
+#[cfg_attr(feature = "serde-deny-unknown-fields", serde(deny_unknown_fields))]
+pub struct RawAddrmanEntry {
```
tcharding:
```suggestion
pub struct RawAddrManEntry {
```

And same everywhere else that rawaddrman is converted to a Rust type.

tcharding:
Everything else looks good. Reviewed: bafede838c449cfe33c2d9f995a11853c726a017

tcharding:
Heads up, we are going to cut a release shortly. Do you want to get this in it?

tcharding (approved):
ACK a7a878511cab28f4c3fb68a923095857961173c3

tcharding:
Thanks man!

Will wait in case @jamillambert wants to take a look before merging.

Issue #429, opened 2025-12-16
Type discrepancy between corepc-types and Bitcoin Core


deadmanoz:
I'm not sure if the types in `corepc-types` are supposed to faithfully match what Bitcoin Core RPC returns, but I noticed a discrepancy with `tx_rate` in `GetChainTxStats`. Bitcoin Core returns `txrate` as a double value (and has since the RPC was introduced in v0.15 [PR #9733](https://github.com/bitcoin/bitcoin/pull/9733)):

```
{
  "txrate": 5.153020888909642
}
```

The current types use `i64` in the JSON structs and `u32` in the model,  yet `f64` should be used if matching Bitcoin Core's output type is the goal?

I understand that this is a breaking change to the interface that people might already be using so wanted to raise it as an issue first before doing anything further.

And, in general, is the intent for the types to always match Bitcoin Core? 

I found this as I was looking to extend some of the RPC extraction functionality being built out in [peer-observer](https://github.com/peer-observer/peer-observer)

tcharding:
Thanks for the issue! `corepc-types` definitely should match what Core returns, you found a bug! I'll happily review any PRs you raise.

deadmanoz:
Ok I have found a few other type inconsistencies tangential to my peer-observer related work.

Me and my mate Claude hacked together an audit tool to uncover any similar inconsistencies between Bitcoin Core's RPC types and `corepc`'s Rust types across all supported versions (v17-v30).

| Status | Issue | Versions | PR |
|--------|-------|----------|-----|
| [ ] | `getchaintxstats.txrate` should be `f64` | v17, v19 | #430 |
| [ ] | `getmininginfo.networkhashps` should be `f64` | v17, v28-v30 | |
| [ ] | `decodepsbt.fee` should be `f64` | v17, v23, v24, v30 | #431 |
| [ ] | `decodepsbt.psbt_version` should be `u64` | v23, v24, v30 | |

The first three are the same pattern - Bitcoin Core returns `double` but `corepc` uses integer types, causing deserialisation failures. The last one is a width mismatch (`u32` vs `u64`) - unlikely to cause issues in practice but worth fixing for correctness.

I am intending to open separate PRs to fix each?

tcharding:
@deadmanoz legendary, thanks for the help man!

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/429

PR #440, opened 2025-12-30
fix(types): use f64 for GetMiningInfo networkhashps field

deadmanoz:
As per raised in https://github.com/rust-bitcoin/corepc/issues/429, addressing type discrepancies between corepc-types and Bitcoin Core.

Bitcoin Core has always returned `networkhashps` as a double since `getmininginfo` was introduced. The previous type (`i64`) causes deserialisation to fail. Here's a minimal example:

```rust
use corepc_client::client_sync::v28::Client;
use corepc_client::client_sync::Auth;

fn main() -> Result<(), Box<dyn std::error::Error>> {
    let auth = Auth::UserPass("bitcoin".to_string(), "bitcoin".to_string());
    let client = Client::new_with_auth("http://127.0.0.1:8332", auth)?;
    let info = client.get_mining_info()?;
    println!("networkhashps: {}", info.network_hash_ps);
    Ok(())
}
```

```
Error: JsonRpc(Json(Error("invalid type: floating point `1.055282156351532e+21`, expected i64", line: 1, column: 179)))
```

This change:
- Changes the `networkhashps` field type from `i64` to `f64`
- Modifies existing integration test to verify correct deserialisation (mines some blocks to yield a `networkhashps` `> 0.0`) and remove redundant comments

### DIFF
--- a/integration_test/tests/mining.rs
+++ b/integration_test/tests/mining.rs
@@ -39,19 +39,21 @@ fn mining__get_block_template__modelled() {
 #[test]
 fn mining__get_mining_info() {
     let node = Node::with_wallet(Wallet::Default, &[]);
+    node.fund_wallet();
 
     let json: GetMiningInfo = node.client.get_mining_info().expect("rpc");
 
     // Up to v28 (i.e., not 29_0) there is no error converting into model.
     #[cfg(feature = "v28_and_below")]
-    let _: mtype::GetMiningInfo = json.into_model();
+    let model: mtype::GetMiningInfo = json.into_model();
 
-    // v29 onwards
     #[cfg(not(feature = "v28_and_below"))]
-    {
-        let model: Result<mtype::GetMiningInfo, GetMiningInfoError> = json.into_model();
-        model.unwrap();
-    }
+    let model: mtype::GetMiningInfo = {
+        let result: Result<mtype::GetMiningInfo, GetMiningInfoError> = json.into_model();
+        result.unwrap()
+    };
+
+    assert!(model.network_hash_ps > 0.0);
 }
 
 #[test]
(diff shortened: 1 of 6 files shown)


tcharding:
Took me a minute to work out why the original test did not catch this bug. And the reason seems to be that when networkhashps is zero, Core returns `0` and `serde` so is willing to deserialize `0` as either a float or an int.

Adding `node.fund_wallet();` to the original test causes the test to fail as expected.

integration_test/tests/mining.rs
```diff
@@ -39,19 +39,17 @@ fn mining__get_block_template__modelled() {
...
-    // Up to v28 (i.e., not 29_0) there is no error converting into model.
     #[cfg(feature = "v28_and_below")]
-    let _: mtype::GetMiningInfo = json.into_model();
-
-    // v29 onwards
+    let model: mtype::GetMiningInfo = json.into_model();
+    // v29 added fields (bits, target) that can fail to parse.
     #[cfg(not(feature = "v28_and_below"))]
-    {
-        let model: Result<mtype::GetMiningInfo, GetMiningInfoError> = json.into_model();
```
tcharding: We want to keep this line with the explicit `Result` type. The reason we do this is to ensure that we exported all the error types correctly. In the past before we did this we forgot a bunch of them.

tcharding: FTR most of the 'weird' stuff here is for a specific reason, not always obvious. Feel free to ask about anything that looks odd. Also if anything is non-uniform them we probably made a mistake - it happens.

deadmanoz: Noted, thanks for the info

PR #443, opened 2025-12-31
fix!(types): ListUnspentItem's amount should be Amount

### LINKED ISSUE #436: corepc-types: `ListUnspentItem`'s `amount` should be an `Amount` instead of `SignedAmount`
The `amount` field in `ListUnspentItem` (returned by the `listunspent` RPC) is typed as `SignedAmount`, but UTXO amounts are always non-negative per the Bitcoin protocol.

https://github.com/rust-bitcoin/corepc/blob/87c72e7ecffa07a6850ffc8fc606ea5ff9f0f7f3/types/src/model/wallet.rs#L684-L686

Bitcoin Core's own RPC documentation does not indicate that this field can be signed/negative.
See https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/listunspent/:

Image

In contrast, signed amounts are appropriate for other RPCs (e.g., `listtransactions`) where negative values indicate outflows/sends/fees.
See https://bitcoincore.org/en/doc/29.0.0/rpc/wallet/listtransactions/:

Image

storopoli:
Closes #436

### DIFF
--- a/types/src/model/wallet.rs
+++ b/types/src/model/wallet.rs
@@ -682,8 +682,7 @@ pub struct ListUnspentItem {
     /// The script key.
     pub script_pubkey: ScriptBuf,
     /// The transaction amount.
-    #[serde(default, with = "bitcoin::amount::serde::as_btc")]
-    pub amount: SignedAmount,
+    pub amount: Amount,
     /// The number of confirmations.
     pub confirmations: u32, // Docs do not indicate what negative value means?
     /// The redeemScript if scriptPubKey is P2SH.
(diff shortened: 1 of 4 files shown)


types/src/model/wallet.rs
```diff
@@ -682,8 +682,7 @@ pub struct ListUnspentItem {
     /// The script key.
     pub script_pubkey: ScriptBuf,
     /// The transaction amount.
-    #[serde(default, with = "bitcoin::amount::serde::as_btc")]
```
tcharding: This change is correct but unrelated to the PR. Took me a minute to work out why it was being removed.

Take pity on us poor reviewers please, throw changes like this in a separate patch that says what it does and save me a few clock cycles. This old brain is hard to kick into gear sometimes.

tcharding: Oh my apologies, my comment was definitely wrong. I _think_ my mistake was I thought model types didn't implement `serde` traits and that was why the attribute was being removed. I was clearly wrong on a few levels. Please excuse me.

tcharding (approved):
ACK 9541cad02ed3aac213c3a7b1d0cb529af960316b

Issue #447, opened 2026-01-06
Tests failing due to Bitcoin Core v30 binaries missing


0xB10C:
As part of https://bitcoincore.org/en/2026/01/05/wallet-migration-bug/, the Bitcoin Core v30 and v30.1 binaries have been removed. This will cause tests that use these binaries through the `node` package to fail:  

```
error: failed to run custom build command for `corepc-node v0.11.0 (https://github.com/0xb10c/corepc?rev=022a23a81e859a5e6f0d8b3774e02f68b2f8a44b#022a23a8)`

Caused by:
  process didn't exit successfully: `/home/runner/work/peer-observer/peer-observer/target/debug/build/corepc-node-d88456b27e32f4f2/build-script-build` (exit status: 101)
  --- stdout
  download_filename: bitcoin-30.0-x86_64-linux-gnu.tar.gz
  filename:bitcoin-30.0-x86_64-linux-gnu.tar.gz version:30.0 hash:00964ae375084113b1162f2f493b9372421608af23539766e315a3cb0ee54248

  --- stderr

  thread 'main' (6540) panicked at /home/runner/.cargo/git/checkouts/corepc-0b6594124fe197fa/022a23a/node/build.rs:96:21:
  assertion `left == right` failed: url https://bitcoincore.org/bin/bitcoin-core-30.0/bitcoin-30.0-x86_64-linux-gnu.tar.gz didn't return 200
    left: 404
   right: 200
```

Opening this issue to track progress on this. Since v30.2 is being released soon, it's likely best to just wait until that's done and then switch to using it.

jamillambert:
> The real issue is that they weren't being cached. Please for christ's sake do not hammer bitcoincore.org trying to download the same files on every CI run.

Files are now cached in #449

TheBlueMatt:
Maybe someone should write a standard GitHub action that incorporates appropriate caching that we can push as a standard?

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/447

PR #448, opened 2026-01-07
Make async fetching actually async

### LINKED ISSUE #441: `bitreq`: Use `tokio::net` types instead of `spawn_blocking` in `AsyncConnection`
Currently, `AsyncConnection` is just a wrapper around the blocking `Connection` and utilizes `spawn_blocking` to spawn a thread per request. This is very heavy-handed. It would be preferable if `AsyncConnection` would be actually async, probably using `tokio::net::TcpSocket` etc.

TheBlueMatt:
Prevously, "async" HTTP calls were actually just spawning blocking
    calls which were run in a background task and waited on. This is
    obviously rather un-scalable (not to mention breaks with
    single-threaded `tokio` executors) and can hang attempts to shut
    down the `tokio` reactor.

    Instead, here, we switch to doing actual async reads.

    In order to avoid changing the API (so that we can ship a point
    release), we still have to support returning a `ResponseLazy` from
    `send_lazy_async`, which we obviously can't make async since
    `ResponseLazy`'s read methods are sync. Instead, in that case we
    pre-load the full response and store the bytes in `ResponseLazy`,
    allowing it to read them as needed.

    Asking claude to copy `ResponseLazy` to an `AsyncResponseLazy`
    which reads from an actual async stream is left as an exercise for
    the reader.

Fixes #441

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -185,6 +185,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
+name = "bytes"
+version = "1.11.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b35204fbdc0b3f4446b89fc1ac2cf84a8a68971995d0bf2e925ec7cd960f9cb3"
+
+[[package]]
 name = "bzip2"
 version = "0.4.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK aab48de940c0766d73c753d316dc07186cf1fe27

My review is quite shallow because I don't know much at all about async Rust.

tcharding:
Would be nice to get a review from someone who knows async but since this crate is so new and I blindly vibe coded all the async stuff anyways I'm happy to just run ACK, merge, YOLO policy on this crate for now. @TheBlueMatt?

tcharding:
I'm going to merge this. The CI fails are unrelated (#447).

tcharding (approved):
ACK 30a7d380ec409c3e17e6e7ebd2f02950cfcafb13

PR #449, opened 2026-01-08
Cache bitcoind for CI

jamillambert:
Cache bitcoind so that it is not downloaded every CI run.

- Change the integration CI job matrix to be the version only and move the `,download` to the run command since the `,` causes an issue when used in the cache key.
- Set `BITCOIND_SKIP_DOWNLOAD` to 1 for `docs`, `docsrs` and `lint` so that bitcoind is not downloaded when it is not needed. Since these jobs are run with a feature matrix that included `download`.
- Apply the changes suggested by clippy to clear the lint errors.
- Cache the bitcoind binary between CI runs, keyed on OS, arch and version.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -242,5 +242,12 @@ jobs:
         uses: actions/checkout@v4
       - name: "Select toolchain"
         uses: dtolnay/rust-toolchain@stable
+      - name: "Cache downloaded bitcoind"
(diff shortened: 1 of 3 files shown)


tcharding (approved):
ACK 3cc51e2cd545a48f2d2903f1d469720cd38bfcd3

tcharding:
Legend.

PR #450, opened 2026-01-10
[bitreq] Support connection reuse and clean up `Connection`

TheBlueMatt:
I went into this thinking it would be easy and claude would be able to do it. I was very, very wrong. The `Connection` stuff was an absolute mess and basically had to be rewritten, but then actually properly supporting pipelining requests was more annoying that I thought. Still, it works now, and I even tested it in BDK's rust-esplora-client inside of ldk-node, which hasn't yet broken. Probably best to review `connection.rs` from scratch as of the third-to-last commit and as of the end and not bother looking at the earlier intermediate commits in `connection.rs`.

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -222,7 +222,10 @@ struct AsyncConnectionState {
     /// can be sent on this socket and a new connection must be made. Thus, in order to limit the
     /// connection to sending N new requests, this may be set to [`usize::MAX`] - N.
     next_request_id: AtomicUsize,
-    /// The ID of the next request which is readable from the socket.
+    /// The ID of the next request which is readable from the socket. If we're pipelining this may
+    /// be a few behind [`Self::next_request_id`]. If this is [`usize::MAX`], the socket is in an
+    /// indeterminate state and no further reading is allowed. Any pending requests must either be
+    /// retried or failed.
     readable_request_id: AtomicUsize,
     /// The time at which we should stop sending new requests over this socket and should instead
     /// connect again.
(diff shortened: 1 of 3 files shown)


tcharding:
ACK 30a7d380ec409c3e17e6e7ebd2f02950cfcafb13

tcharding (approved):

bitreq/src/connection.rs
```diff
@@ -312,35 +340,167 @@ impl AsyncConnection {
...
+                while conn.readable_request_id.load(Ordering::Acquire) < request_id {
+                    // There's a race where we can finish writing but see a context switch between
+                    // dropping the write lock and getting to the waiter that can lead to waiters being
+                    // registered out of order. Thus, if we're not actually ready to read, wake another
+                    // waiter and see if we're ready when we get the semaphore back.
+                    read.take();
+                    tokio::task::yield_now().await;
+                    read = Some(conn.read.lock().await);
+                }
+                let mut read = read.take().unwrap();
```
tnull: nit: Seems fine for now, but this `unwrap` looks like it could easily become reachable if there's some refactor that doesn't have the full context. Not sure if worth changing now that the PR already landed, but I'd personally prefer avoid introducing such (~any really) `unwrap`s.

tnull (commented):
Again, didn't spent a huge amount of time on review here. Generally looks good (so post-merge ACK from my side), but I have some questions/nits that might or might not be worth addressing in a follow-up.

(cc @TheBlueMatt)

bitreq/src/connection.rs
TheBlueMatt: Not really sure how else to write this given the loop above. I mean we can `debug_assert` that its there and then re-lock if we have to but that code would be really incredibly wrong.

PR #452, opened 2026-01-12
[bitreq] Fix unbounded proxy / body reads

tnull:
Asked Claude to fix a few minor issues that might allow a malicious counterparty to send us unbounded data.

(cc @TheBlueMatt)

### DIFF
--- a/bitreq/src/error.rs
+++ b/bitreq/src/error.rs
@@ -58,12 +58,16 @@ pub enum Error {
     HttpsFeatureNotEnabled,
     /// The provided proxy information was not properly formatted. See
     /// [Proxy::new](crate::Proxy::new) for the valid format.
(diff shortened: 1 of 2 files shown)


TheBlueMatt (commented):
thanks

bitreq/src/request.rs
```diff
@@ -247,6 +249,23 @@ impl Request {
         self
     }
 
+    /// Sets the maximum size of the response body this request will
+    /// accept.
+    ///
+    /// If this limit is passed, the request will close the connection
+    /// and return an [Error::BodyOverflow] error.
```
TheBlueMatt: Oops, worth noting this does not apply to lazily-loaded requests (or maybe we should make it?)

TheBlueMatt:
Will do in the future!

PR #454, opened 2026-01-13
Fix bitreq doc example

### LINKED ISSUE #434: CI failure on bitreq docs
There is an unrelated CI failure in a PR that looks like it is a rustdoc issue in `bitreq`, possibly from a bad response from `example.com`:

```
---- bitreq/src/lib.rs - (line 73) stdout ----
Test executable failed (exit status: 101).

stderr:

thread 'main' (8704) panicked at bitreq/src/lib.rs:8:1:
assertion `left == right` failed
  left: 200
 right: 403
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace



failures:
    bitreq/src/lib.rs - (line 73)

test result: FAILED. 12 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.40s
```

I reran the test and it passed on the third time, but now on a second PR it fails for stable and nightly with the same error message.

jamillambert:
The example in the docs cause a random CI failure.

Change it so that it doesn't rely on `example.com` anymore. Add a timeout to avoid hanging.

Closes #434

### DIFF
--- a/bitreq/src/lib.rs
+++ b/bitreq/src/lib.rs
@@ -65,21 +65,35 @@
 //!
 //! ## Get
 //!
-//! This is a simple example of sending a GET request and printing out
-//! the response's body, status code, and reason phrase. The `?` are
+//! This is a simple example of sending a GET request and checking the
+//! response's body, status code, and reason phrase. The `?` are
 //! needed because the server could return invalid UTF-8 in the body,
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 3c0c4cefb0071efa2009fce8eab000d33c48acb9

PR #463, opened 2026-01-16
[bitreq] Add default size limits for headers, status line, and body

TheBlueMatt:
I realize we also don't have default size limits for the total response headers or status line, which is quite a large footgun. Instead, here we adopt the default response header limit from chrome and set a status line limit of 1/4 of that. Also adds a 1GiB default body response just to have *some* limit even if its incredibly high and we still document that folks should really set it.

IMO we should cut this as 0.3.1.


tcharding:
> nit: Could be defined as consts e.g. DEFAULT_MAX_HEADER_SIZE to make it easier to understand what the numbers are when scanning over the code.

I elected to merge anyway and just leave the magic numbers in there.

PR #465, opened 2026-01-19
Release tracking PR: `bitreq v0.3.1`

tcharding:
In preparation for releasing the changes in #463 (add default size limits for headers, status line, and body).

Point release because if you squint 463 was a bug fix.

Bump the version number, add a changelog entry, and update the lock files.


TheBlueMatt:
I noticed that and figured it was on purpose, but its obviously more correct - isnt minimal supposed to be the lowest ones?

PR #466, opened 2026-01-20
Set `all-features = true ` in `docs.rs` builds

tnull:
Previously, the docs have been generated without the `async` feature, so `Client` for example never showed up. Here we add the feature to the doc list.


jamillambert (approved):
ACK 0c37601e3a3a98e0a3e9bb5b584b71daa026fd1b

oleonardolima (approved):
utACK 0c37601e3a3a98e0a3e9bb5b584b71daa026fd1b

PR #469, opened 2026-01-21
add: support for v30.2 and integration test it

0xB10C:
v30.0 and v30.1 currently aren't covered by the ingegration tests and their binaries aren't avalaible for download anymore due to a wallet migration bug. See https://github.com/rust-bitcoin/corepc/issues/447

We can however use v30.2 in the integration tests. Also, make it avaliable as feature.


0xB10C:
Set 30_2 as `latest`.

PR #471, opened 2026-01-23
Add lock! macro to handle mutex poisoning gracefully

crodas:
Mutex::lock().unwrap() panics if the mutex is poisoned (when a thread panicked while holding the lock). In practice, this is rarely the desired behavior - the data may still be valid and propagating panics across threads is usually unhelpful.

This adds a lock! macro that recovers from poisoning via unwrap_or_else(|e| e.into_inner()), providing a concise and consistent way to acquire locks throughout the codebase.

An alternative would be using parking_lot::Mutex which has no poisoning semantics, but this avoids adding an external dependency.

### DIFF
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -66,7 +66,7 @@ impl Client {
 
         // Try to get cached connection
         let conn_opt = {
-            let state = self.r#async.lock().unwrap();
+            let state = lock!(self.r#async);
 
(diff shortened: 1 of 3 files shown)


TheBlueMatt (commented):
I'd kinda rather make sure that none of the operations within any of our locks can possibly panic. I believe we're close but you identified one bug where we're holding a mutex for too long.

tcharding:

Issue #474, opened 2026-01-23
Type discrepancies between corepc and v30 blockchain module.


natobritto:
Hi! I manually inspected the types for every call in blockchain module compared to v30. I found some errors. Here is a report of what I found:

- [x] dumptxoutset
- [x] getbestblockhash
- [ ] getblock: The RPC has 4 distinct verbose return types, only 2 of which are implemented.
- [x] getblockchaininfo
- [x] getblockcount
- [x] getblockfilter
- [x] getblockfrompeer
- [x] getblockhash
- [x] getblockheader
- [x] getblockstats
- [x] getchainstates
- [x] getchaintips
- [x] getchaintxstats
- [ ] getdeploymentinfo: The getdeploymentinfo RPC blockhash argument is optional. If omitted, it queries the blockchain tip.
- [ ] getdescriptoractivity: Bitcoin Core v30 marks the blockhashes and scanobjects arguments of getdescriptoractivity RPC as required, instead of failing with a JSON parsing error when omitted.
- [x] getdifficulty
- [x] getmempoolancestors
- [x] getmempooldescendants
- [x] getmempoolentry
- [x] getmempoolinfo
- [ ] getrawmempool: When getrawmempool RPC is called with optional verbose argument set to false and subsequent mempool_sequence set to true, a distinct type is returned from v21 onward to v30.
- [x] gettxoutproof
- [ ] gettxoutsetinfo: The RPC has missing fields which are enabled only when coinstatsindex option is called or muhash hash type is used.
- [x] gettxspendingprevout
- [x] importmempool
- [x] preciousblock
- [x] pruneblockchain
- [x] savemempool
- [ ] scanblocks: From v26 onward, the return type of scanblocks RPC has an argument called completed.
- [x] verifychain
- [x] verifytxoutproof
- [ ] estimaterawfee: The conf_target argument only takes in values from 1 to 1008 inclusive. All values outside these ranges return an error.
- [x] waitforblock
- [x] waitforblockheight
- [x] waitfornewblock

PR #476, opened 2026-01-23
[client] Validate `conf_target` argument value range in `estimaterawfee` RPC

satsfy:
The `conf_target` argument only takes in values from 1 to 1008 inclusive. All values outside these ranges return an error. This commit adds validation for it in all versions from v17 to v30 with tests.

Partially fixes issue #474.


jamillambert (dismissed):
ACK d0a8016aed903654e4033d974a684f0bce31ffd8

tcharding (approved):
ACK 433d03760a4e8ecaaa1be0690e202fda3f9684a9

PR #478, opened 2026-01-23
Implement `GetRawMempoolSequence` return type for `getrawmempool` RPC

satsfy:
When `getrawmempool` RPC is called with optional verbose argument set to `false` and subsequent `mempool_sequence` set to `true`, a distinct type is returned from v21 onward to v30. This PR implements this return type with corresponding tests.

Partially fixes issue #474.


jamillambert (dismissed):
ACK 49281789d45ca965d0dfcf3ba301aba7d88c4f9e

PR #485, opened 2026-01-27
Add `getblock` verbosity 2/3 support for v29 onward

### LINKED ISSUE #474: Type discrepancies between corepc and v30 blockchain module.

satsfy:
This PR:
- Adds full support for `getblock` RPC with verbosity levels 2 and 3 for v29 and later.
- Add coinbase transaction support in `RawTransactionInput` type, to be fully compliant with the RPC specification.
- Introduces versioned RPC types and model conversions, including transaction fees and prevout data, extends raw transaction input handling to correctly support coinbase inputs, and updates the client API with verbosity levels.
- Adds integration tests to validate new models conversions.
- Removes FIXME comment regarding weight units on getblock verbosity 0 and 1, the value really is weight units (WU).

This partially fixes #474.

Notably missing are the changes in implementation from v17 to v28 inclusive.


jamillambert (commented):
I would prefer that `GetBlockVerboseTwoTransaction` and `GetBlockVerboseThreeTransaction` matched the RPC output. There are already so many structs and fields I think adding another 6 fields to these structs is preferable to changing the way that they need to be accessed.

Everything else looks good. I would wait for @tcharding to review before making the above change as he may disagree.

Reviewed abc433cc9e0acb3acd909d267edd930ea1a9df35

PR #486, opened 2026-01-28
[bitreq] Check utf-8 while deserializing JSON body

tankyleo:
```
    [bitreq] Check utf-8 while deserializing JSON body

    While deserializing, `serde_json::from_slice` validates utf-8 as
    needed. So instead of making two passes on the response body, one
    to validate utf-8, and another to deserialize the object, we can
    let `serde_json::from_slice` check utf-8 as needed during
    deserialization.

    `Response::json` now returns `Error::SerdeJsonError` instead of
    `Error::InvalidUtf8InBody` if invalid utf-8 bytes are found during
    deserialization.
```
and
```
    [bitreq] Serialize `Request` body into `Vec<u8>`

    We avoid creating a `String` only to immediately convert it back to its
    inner `Vec<u8>`.

    This also mirrors the `serde_json::from_slice` call made when parsing a
    `Response` body as JSON.
```

### DIFF
--- a/bitreq/src/request.rs
+++ b/bitreq/src/request.rs
@@ -190,7 +190,7 @@ impl Request {
     pub fn with_json<T: serde::ser::Serialize>(mut self, body: &T) -> Result<Request, Error> {
         self.headers
             .insert("Content-Type".to_string(), "application/json; charset=UTF-8".to_string());
-        match serde_json::to_string(&body) {
+        match serde_json::to_vec(&body) {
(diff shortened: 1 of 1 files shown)


tankyleo:
This came up while me and @tnull were working on migrating to bitreq in https://github.com/lightningdevkit/vss-client/pull/56

TheBlueMatt (approved):
Mmm, fair enough.

ACK 2c3bbf21a47132b4292e13093e8f6068db679dff

PR #488, opened 2026-01-29
Handle async pipelined request cancellation correctly

### LINKED ISSUE #487: `[bitreq]` pipelining cancelation safety
our pipelined requests aren't cancel-safe. we'll need to track cancellation somehow and fail the connection.

TheBlueMatt:
It turns out if we get a pipelined request and it is cancelled (by dropping the `Future` reading the response), we'll end up infinite-looping all future requests on the connection waiting for their turn to read that will never come.

Instead, we have to manually detect `Future` drops and retry later requests.

Fixes #487

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -227,11 +227,32 @@ struct AsyncConnectionState {
     /// indeterminate state and no further reading is allowed. Any pending requests must either be
     /// retried or failed.
     readable_request_id: AtomicUsize,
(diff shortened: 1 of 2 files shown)


tcharding (approved):
ACK 5c2b0890ee721fd5624646b19edb7ae08a186308 - did not review. Need one ack from a maintainer to run the merge scritp. Invited @tnull to be a maintainer so this doesn't happen again.

PR #489, opened 2026-02-03
ci: use ubuntu-slim for label & prepare ci tasks

0xB10C:
The `ubuntu-slim` 1vCPU runners were recently introduced and seem to be the right fit here.

https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/

https://docs.github.com/en/actions/reference/runners/github-hosted-runners#single-cpu-runners

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -5,7 +5,7 @@ on:
 
 jobs:
   label-pr:
-    runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


jamillambert (approved):
ACK f1866420c1c6659e7c09d2853fc4f7975a0b3913

Looks like these two jobs are a good fit. Not sure it will make any real difference though since they are so fast.

PR #498, opened 2026-02-10
bitreq: Fix issues with non-`std` builds and add `#[no_std]` attribute

### LINKED ISSUE #492: bitreq: Does not implement no-std
There is no

```rust
#![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
```

And we use types from `std` with no feature gating ...

mpbagot:
When bitreq doesn't have the std feature enabled, it currently fails to correctly disable the use of std-specific features. In some cases, std is used where core would suffice. Importantly, types that have no use without std remain public from the crate.

Introduce a gated no_std attribute to the crate.
Fix feature gates in url module.
Gate out Response and Request for no-std.

Closes #492

### DIFF
--- a/bitreq/src/lib.rs
+++ b/bitreq/src/lib.rs
@@ -242,6 +242,7 @@
 //! If the timeout is set with `with_timeout`, the environment
 //! variable will be ignored.
 
+#![cfg_attr(all(not(feature = "std"), not(test)), no_std)]
 #![deny(missing_docs)]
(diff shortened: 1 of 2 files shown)


mpbagot:
That makes sense. I assumed that you wanted to just keep everything that was currently there.
I fixed the format issue from the CI and I've changed the PR to just gate things at the lib.rs level + fix the issues in the url module.

PR #500, opened 2026-02-11
[bitreq] Make `Url::append_query_pairs` and `preserve_fragment_from` public methods

tnull:
We previously considered `append_query_pairs` and
`preserve_fragment_from` only internal methods and hence made them only `pub` when fuzzing.

Unfortunately it turns out that they are very useful and we need them for downstream migration, so here we make them `pub` for everybody.

### DIFF
--- a/bitreq/fuzz/src/url_parse.rs
+++ b/bitreq/fuzz/src/url_parse.rs
@@ -86,7 +86,7 @@ pub fn do_test(data: &[u8]) {
             {
                 let mut url_clone = bitreq_url.clone();
                 // Use the input itself as both key and value to exercise encoding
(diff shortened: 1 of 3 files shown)


TheBlueMatt (commented):
ACK 0736073abfa5554711cb9394611f7f12d87424ba

tcharding (dismissed):
ACK 0736073abfa5554711cb9394611f7f12d87424ba

tcharding:
Ah damn, I've done the release without merging this first - fail!

tcharding (approved):
ACK cb8d981ae9894ed9f7c4dfa553146ef173763e7a

Issue #473, opened 2026-01-23
feat(bitreq): API for custom root certificates at runtime


DanGould:
## Problem

We’re migrating https://github.com/payjoin/rust-payjoin/issues/1288 from reqwest to bitreq to reduce our dependency footprint.

A current blocker is payjoin’s `_manual-tls` feature, which requires adding custom root certificates at runtime. This is necessary for development and testing scenarios that rely on self-signed certificates.

Current reqwest implementation

In payjoin ([payjoin/src/core/io.rs](https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/core/io.rs)):
```rs
#[cfg(feature = "_manual-tls")]
pub async fn fetch_ohttp_keys_with_cert(
    ohttp_relay: impl IntoUrl,
    payjoin_directory: impl IntoUrl,
    cert_der: Vec<u8>,
) -> Result<OhttpKeys, Error> {
    let client = Client::builder()
        .use_rustls_tls()
        .add_root_certificate(
            reqwest::tls::Certificate::from_der(&cert_der)?
        )
        .proxy(proxy)
        .http1_only()
        .build()?;

    // ...
}
```


In payjoin-cli ([src/app/mod.rs](https://github.com/payjoin/rust-payjoin/blob/45689ba2cbd6d8bcde7974987b280e34a9f514b6/payjoin-cli/src/app/mod.rs#L75-L87)):

```rs
#[cfg(feature = "_manual-tls")]
fn http_agent_builder(
    root_cert_path: Option<&std::path::PathBuf>,
) -> Result<reqwest::ClientBuilder> {
    let mut builder = reqwest::ClientBuilder::new()
        .use_rustls_tls()
        .http1_only();

    if let Some(root_cert_path) = root_cert_path {
        let cert_der = std::fs::read(root_cert_path)?;
        builder = builder.add_root_certificate(
            reqwest::tls::Certificate::from_der(cert_der.as_slice())?
        );
    }

    Ok(builder)
}
```

## Desired Solution
Add certificate configuration to Client (or a new ClientBuilder), matching reqwest's architecture where TLS settings belong at the client/connection level rather than per-request       

```rs                                                                                                                                               
 // Proposed API - Option A: ClientBuilder pattern (matches reqwest)                                                                                                                       
 let cert_der = std::fs::read("path/to/cert.der")?;                                                                                                                                        
 let client = bitreq::Client::builder()                                                                                                                                                    
[...]

PR #502, opened 2026-02-11
[bitreq] Bitreq client builder

ton-anywhere:
# Summary

This PR addresses issue #473 by creating a Client Builder able to receive custom root certificates at runtime.
The ClientBuilder requires the same features as Client (async-https-rustls).

# Changes

- Add Client Builder pattern.
- Creates Certificates wrapper module
- Append certificate using with_root_certificate builder method.
- Include example.

### DIFF
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -153,9 +153,9 @@ impl ClientBuilder {
     /// Creates a new `ClientBuilder` with default settings.
     ///
     /// Default configuration:
-    /// * `capacity` - 1 (single connection)
+    /// * `capacity` - 10 (single connection)
     /// * `root_certificates` - None (uses system certificates)
-    pub fn new() -> Self { Self { capacity: 1, client_config: None } }
+    pub fn new() -> Self { Self { capacity: 10, client_config: None } }
 
     /// Sets the maximum number of connections to keep in the pool.
     ///


tcharding:
Holla if you want/need any review mate. Otherwise, for me at least, while there are WIP commits I'll assume you are happily working on it still.

APonce911:
Thanks @tcharding. My idea with this draft was to give a bit of visibility of where we're at and to guarantee that the solution is in the right path. No formal code review is needed yet.

Thanks for the tip about the CI, so far it's ok not to run.

bitreq/src/connection/rustls_stream.rs
```diff
@@ -106,6 +156,33 @@ pub(super) async fn wrap_async_stream(
     Ok(AsyncHttpStream::Secured(Box::new(tls)))
 }
 
+#[cfg(all(feature = "rustls", feature = "tokio-rustls"))]
+pub(super) async fn wrap_async_stream_with_configs(
```
APonce911: same as _wrap_async_stream_ but with additional parameter _custom_client_config_

bitreq/src/connection.rs
```diff
@@ -298,6 +300,42 @@ impl AsyncConnection {
         }))))
     }
 
+    pub(crate) async fn new_with_configs(
```
APonce911: same as _new_ but with _client_config_ parameter

APonce911:
@tcharding hey mate, can you please run the CI once? It's ready for review.

bitreq/src/connection.rs
TheBlueMatt: Rather than having a new method, just convert the existing (crate-private) method to taking an `Option<config>`?

bitreq/src/client.rs
```diff
@@ -39,6 +39,50 @@ struct ClientImpl<T> {
...
+pub struct TlsConfig {
+    pub custom_certificate: Vec<u8>, // DER-encoded cert
+}
+
+impl ClientBuilder {
+    pub fn new() -> Self {
+        Self { capacity: 1, client_config: None }
+    }
+
+    pub fn with_root_certificate<T: Into<Vec<u8>>>(mut self, certificate: T) -> Self {
```
TheBlueMatt: Its pretty awkward that we don't error here but then just ignore certs silently. IMO we kinda need to actually make this fallible, store a root cert store in `ClientConfig`, and pass that around instead. We'll have to have an abstract root cert store in `rustls_stream` (or somewhere), but that should be pretty straightforward.

TheBlueMatt (commented):
Obv needs docs, but at a high level looks good to me.

tcharding:
Just an FYI, you can use `just` to run build commands in this repo mate.

https://github.com/casey/just

bitreq/src/connection/certificates.rs
```diff
@@ -0,0 +1,64 @@
...
+        let result = if let Some(certificate) = certificate {
+            certificates.append_certificate(certificate)
+        } else {
+            Ok(certificates)
+        };
+        result
+    }
+
+    #[cfg(feature = "rustls")]
+    pub(crate) fn append_certificate(mut self, certificate: &Vec<u8>) -> Result<Self, Error> {
```
DanGould: Same here, tell me the fmt

bitreq/src/connection/certificates.rs
```diff
@@ -0,0 +1,64 @@
...
+
+use crate::Error;
+
+#[derive(Clone)]
+pub(crate) struct Certificates {
+    pub(crate) inner: RootCertStore,
+}
+
+impl Certificates {
+    pub(crate) fn new(certificate: Option<&Vec<u8>>) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
APonce911: @DanGould Great catch. Yes, you're correct. We must receive a DER encoded certificate. 
I've updated the variable to cert_der to give more clarity. About the doc, I've written on the Client class since it's public. 
https://github.com/rust-bitcoin/corepc/pull/502/changes#diff-21f18a8053f4d7bce98cbe0c84adc7e943caf49c0a245be26096be8ce461ae64R100  
What do you think?

PR #505, opened 2026-02-12
Create an async corepc-client

jamillambert:
This PR adds an async corepc-client with the RPCs used by BDK.

The patches are structured to make it easier to see what was changed from sync to async. First the sync version is copied and renamed, then in a separate patch it is changed to async.

### DIFF
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -22,7 +22,7 @@ allowed_duplicates = ["base64"]
 
 [features]
 # Enable this feature to get a blocking JSON-RPC client.
-client-sync = ["jsonrpc"]
+client-sync = ["jsonrpc", "jsonrpc/bitreq_http"]
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false, features = ["std", "serde"] }
@@ -31,6 +31,6 @@ serde = { version = "1.0.103", default-features = false, features = [ "derive", 
 serde_json = { version = "1.0.117" }
 types = { package = "corepc-types", version = "0.15.0", path = "../types", default-features = false, features = ["std"] }
 
-jsonrpc = { version = "0.20.0", path = "../jsonrpc", features = ["bitreq_http"], optional = true }
+jsonrpc = { version = "0.20.0", path = "../jsonrpc", default-features = false, optional = true }
 
 [dev-dependencies]
--- a/client/Cargo.toml
+++ b/client/Cargo.toml
@@ -23,6 +23,8 @@ allowed_duplicates = ["base64"]
 [features]
 # Enable this feature to get a blocking JSON-RPC client.
 client-sync = ["jsonrpc", "jsonrpc/bitreq_http"]
+# Enable this feature to get an async JSON-RPC client.
+client-async = ["jsonrpc", "jsonrpc/bitreq_http_async", "jsonrpc/client_async"]
 
 [dependencies]
 bitcoin = { version = "0.32.0", default-features = false, features = ["std", "serde"] }
--- a/client/src/client_async/error.rs
+++ b/client/src/client_async/error.rs
@@ -1,0 +1,112 @@
+// SPDX-License-Identifier: CC0-1.0
+
+use std::{error, fmt, io};
+
+use bitcoin::hex;
+
+/// The error type for errors produced in this library.
+#[derive(Debug)]
+pub enum Error {
+    JsonRpc(jsonrpc::error::Error),
+    HexToArray(hex::HexToArrayError),
+    HexToBytes(hex::HexToBytesError),
+    Json(serde_json::error::Error),
+    BitcoinSerialization(bitcoin::consensus::encode::FromHexError),
+    Io(io::Error),
+    InvalidCookieFile,
+    /// The JSON result had an unexpected structure.
+    UnexpectedStructure,
+    /// The daemon returned an error string.
+    Returned(String),
+    /// The server version did not match what was expected.
+    ServerVersion(UnexpectedServerVersionError),
+    /// Missing user/password.
+    MissingUserPassword,
+}
+
+impl From<jsonrpc::error::Error> for Error {
+    fn from(e: jsonrpc::error::Error) -> Error { Error::JsonRpc(e) }
+}
+
+impl From<hex::HexToArrayError> for Error {
+    fn from(e: hex::HexToArrayError) -> Self { Self::HexToArray(e) }
+}
+
+impl From<hex::HexToBytesError> for Error {
+    fn from(e: hex::HexToBytesError) -> Self { Self::HexToBytes(e) }
+}
+
+impl From<serde_json::error::Error> for Error {
+    fn from(e: serde_json::error::Error) -> Error { Error::Json(e) }
+}
+
+impl From<bitcoin::consensus::encode::FromHexError> for Error {
+    fn from(e: bitcoin::consensus::encode::FromHexError) -> Error { Error::BitcoinSerialization(e) }
+}
+
+impl From<io::Error> for Error {
+    fn from(e: io::Error) -> Error { Error::Io(e) }
+}
+
+impl fmt::Display for Error {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        use Error::*;
+
+        match *self {
+            JsonRpc(ref e) => write!(f, "JSON-RPC error: {}", e),
+            HexToArray(ref e) => write!(f, "hex to array decode error: {}", e),
+            HexToBytes(ref e) => write!(f, "hex to bytes decode error: {}", e),
+            Json(ref e) => write!(f, "JSON error: {}", e),
+            BitcoinSerialization(ref e) => write!(f, "Bitcoin serialization error: {}", e),
+            Io(ref e) => write!(f, "I/O error: {}", e),
+            InvalidCookieFile => write!(f, "invalid cookie file"),
+            UnexpectedStructure => write!(f, "the JSON result had an unexpected structure"),
+            Returned(ref s) => write!(f, "the daemon returned an error string: {}", s),
+            ServerVersion(ref e) => write!(f, "server version: {}", e),
+            MissingUserPassword => write!(f, "missing user and/or password"),
+        }
+    }
+}
+
+impl error::Error for Error {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        use Error::*;
+
+        match *self {
+            JsonRpc(ref e) => Some(e),
+            HexToArray(ref e) => Some(e),
+            HexToBytes(ref e) => Some(e),
+            Json(ref e) => Some(e),
+            BitcoinSerialization(ref e) => Some(e),
+            Io(ref e) => Some(e),
+            ServerVersion(ref e) => Some(e),
+            InvalidCookieFile | UnexpectedStructure | Returned(_) | MissingUserPassword => None,
+        }
+    }
+}
+
+/// Error returned when RPC client expects a different version than bitcoind reports.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct UnexpectedServerVersionError {
+    /// Version from server.
+    pub got: usize,
+    /// Expected server version.
+    pub expected: Vec<usize>,
+}
+
+impl fmt::Display for UnexpectedServerVersionError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        let mut expected = String::new();
+        for version in &self.expected {
+            let v = format!(" {} ", version);
+            expected.push_str(&v);
+        }
+        write!(f, "unexpected bitcoind version, got: {} expected one of: {}", self.got, expected)
+    }
+}
+
+impl error::Error for UnexpectedServerVersionError {}
+
+impl From<UnexpectedServerVersionError> for Error {
+    fn from(e: UnexpectedServerVersionError) -> Self { Self::ServerVersion(e) }
+}
--- a/client/src/client_async/error.rs
+++ b/client/src/client_async/error.rs
@@ -1,23 +1,30 @@
 // SPDX-License-Identifier: CC0-1.0
 
+//! Async client error types.
+//!
+//! Async RPC methods return per-method typed errors (for example, [`GetBlockError`])
+//! so callers can match method-specific `Model` conversion failures directly.
+//!
+//! Every method error exposes inherent helpers (`is_not_found_error`, `is_transport_error`,
+//! `rpc_code`, `as_client_error`) for common JSON-RPC and transport inspection.
+
+use core::convert::Infallible;
 use std::{error, fmt, io};
 
-use bitcoin::hex;
-
-/// The error type for errors produced in this library.
+/// The general error type for the async client.
+///
+/// This covers connecting, authenticating, and performing a JSON-RPC call. Each RPC method returns
+/// its own error type (e.g. [`GetBlockError`]) which wraps this type in its `Rpc` variant.
 #[derive(Debug)]
 pub enum Error {
+    /// A JSON-RPC error occurred (transport error or the node returned an error).
     JsonRpc(jsonrpc::error::Error),
-    HexToArray(hex::HexToArrayError),
-    HexToBytes(hex::HexToBytesError),
+    /// Serializing an argument or deserializing the response failed.
     Json(serde_json::error::Error),
-    BitcoinSerialization(bitcoin::consensus::encode::FromHexError),
+    /// An I/O error occurred (e.g. reading the cookie file).
     Io(io::Error),
+    /// The cookie file was invalid.
     InvalidCookieFile,
-    /// The JSON result had an unexpected structure.
-    UnexpectedStructure,
-    /// The daemon returned an error string.
-    Returned(String),
     /// The server version did not match what was expected.
     ServerVersion(UnexpectedServerVersionError),
     /// Missing user/password.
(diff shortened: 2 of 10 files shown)


jsonrpc/src/client_async.rs
```diff
@@ -17,29 +19,34 @@ use serde_json::Value;
...
+    pub(crate) transport: Box<dyn AsyncTransport>,
     nonce: atomic::AtomicUsize,
 }
 
-impl Client {
+impl AsyncClient {
     /// Creates a new client with the given transport.
-    pub fn with_transport<T: Transport>(transport: T) -> Client {
-        Client { transport: Box::new(transport), nonce: atomic::AtomicUsize::new(1) }
+    pub fn with_transport<T: AsyncTransport>(transport: T) -> AsyncClient {
```
tcharding: My suggestion would be to not rename these types. That way the diff is smaller. I'd rename `client.rs` to `client_sync.rs`. You'll need to remove the public re-exports from `lib.rs` and that maybe fix up some paths to always use `client_sync::` or `client_async::` as appropriate.

tcharding:
@apoelstra whats your take man? Would you be willing to have `jsonrpc` take changes to be able to use `bitreq` in an `async` manner? It obviously smashes the dep tree if `--all-features` are used.

tcharding:
Holla at me if/when you would like some review mate.

client/src/client_async/bdk_client.rs
```diff
@@ -0,0 +1,107 @@
...
+const VERSION_WITH_TARGET_FIELD: usize = 290000;
+
+impl Client {
+    async fn bdk_server_version(&self) -> Result<usize> {
+        let info: serde_json::Value = self.call("getnetworkinfo", &[]).await?;
+        let version = info
+            .get("version")
+            .and_then(serde_json::Value::as_u64)
+            .ok_or(Error::UnexpectedStructure)?;
+        usize::try_from(version).map_err(|_| Error::UnexpectedStructure)
```
tcharding: Unfortunately we cannot do this. This is one of the main reasons behind certain design choices in this repo. While it seems sane I heard but cannot remember right now that there may be reasons why a node operator disables the `getnetworkinfo` method and it shouldn't be relied upon. Sorry for the poor explanation, it was a while ago now and I just committed to memory that we can't rely on it.

I'll comment further near where this is called.

tcharding:
> Rewrote the async client:
> 
> * Removed all the version folders where the RPC macros were defined.
> * Removed macroization from client_async module.
> * Created `bdk_client` module that implements the the required RPC methods in it using the shared client methods in mod.rs.

After writing a massive post I realised this solution is not going to work (see bottom for explanation). I think we should inline the `bdk` module and just document that its designed explicitly with that project in mind.

How about throwing this at the top:

```rust
//! Async JSON-RPC client designed explicitly to support BDK.
//!
//! ## Project decisions
//!
//! * Support Core versions 25 to 30.
```

>   * Supports v25 to v30 by checking the server version and then using the required vtype version before converting into the model type.

Troublesome, commented already on the code.

>   * The bdk_client RPC methods all return the modelled types. This made the error handling difficult since all of the error types for converting into the modelled type are version specific types. I am not sure how to handle this ergonomically. Currently they all return the same Error type defined in the async client.

Another solution is to have an error type for each function. Then it can have a variant (assuming its an enum) for the exact error returned by `into_model`.

>   * There are a few RPCs that return a single field, e.g. `getblock` returns a `Block` and `getblockcount` a `u32`. Would it be better to return e.g. `Result<u32>` instead of `Result<GetBlockCount>` ?

Agreed that a wrapper type is not useful to returen (eg `GetBlockCount`). At first blush I think you are correct and returning the inner type is better (its a `u64` in this case). After upgrade to `bitcoin v0.33` we may have better concrete types to return if needed but your argument stands.

> * Rewrote the tests to use the new bdk_client rpc calls.

Nice.


> @tcharding thoughts? in particular the error handling of converting to the modelled type. Returning a modelled type vs the inner field e.g. `Block` or `u32`. And having a bdk_client module that only implements the RPCs, the idea was to be able to add another one for e.g. LDK with their required RPCs and arguments but sharing the `new`, `call` etc. methods.

I don't think separation by project module is not going to work because the modules will conflict with each other. And if we feature gate the features will not be additive. We could try some macro stuff but I think it defeats the purpose which is to write a simple clean client that others can fork if they need to.

jamillambert:
> > @apoelstra whats your take man? Would you be willing to have `jsonrpc` take changes to be able to use `bitreq` in an `async` manner? It obviously smashes the dep tree if `--all-features` are used.
> 
> Would this then tie all users of the crate to tokio? Why can't it be executor-agnostic?

The changes to `jsonrpc` to make it support async is executor-agnostic. It is only the use of `bitreq` and async that requires `tokio` in the current state of this PR.

jamillambert:
Renaming the `jsonrpc` sync client feature was a breaking change. I have reverted it so that the changes in enabling async are all additive. 

I used AI to search through all the dependents of `jsonrpc` on `crates.io` none of them forward features to it and `async` is off by default. So adding feature gated async support should not require any published downstream to depend on `tokio` or any other new dependency, if they update to the new version with `async` support. Unless of course they enable the async feature. Making `bitreq` executor-agnostic is a separate issue I will leave to the related issue.

jamillambert:
Reviving this PR.

I rebased on master now that jsonrpc has async support. The commit sequence is recreated to be simpler. Still a work in progress.

jamillambert:
@tcharding I think this is ready for preliminary review.

The last patch is a new idea to make the async client universal by not having any inherent RPC methods. Users instead add an extension trait to implement the RPCs that they need in their own codebase. The RPCs for BDK have already been done in this crate and can be used as a template for others. Users don't need to fork the client and can instead depend on the published version which is better for any fixes or security updates. It can also stay in this repo.

The docs will need to be improved, I'm leaving it for now due to this possible change in structure.

The CI failure for v31 needs the related PR and then I'll see if any of the async client tests need to be updated.

client/Cargo.toml
```diff
@@ -31,6 +33,6 @@ serde = { version = "1.0.103", default-features = false, features = [ "derive",
 serde_json = { version = "1.0.117" }
 types = { package = "corepc-types", version = "0.14.0", path = "../types", default-features = false, features = ["std"] }
 
-jsonrpc = { version = "0.20.0", path = "../jsonrpc", features = ["bitreq_http"], optional = true }
+jsonrpc = { version = "0.20.0", path = "../jsonrpc", optional = true }
```
tcharding: Perhaps we should turn off default features too, they look unused.

client/src/client_async/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+        let json: NetworkVersion = self.call("getnetworkinfo", &[]).await?;
+        Ok(json.version)
+    }
+
+    /// Checks that the JSON-RPC endpoint is for a `bitcoind` instance with the expected version.
+    ///
+    /// # Parameters
+    ///
+    /// `expected_versions`: A vector of expected server versions e.g., `[230100, 230200]`.
+    pub async fn check_expected_server_version(&self, expected_versions: &[usize]) -> Result<()> {
```
tcharding: This is an integration-ish method, I rekon remove it.

client/src/client_async/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+                got: server_version,
+                expected: expected_versions.to_vec(),
+            }));
+        }
+        Ok(())
+    }
+}
+
+/// Shorthand for converting a variable into a `serde_json::Value`.
+pub fn into_json<T>(val: T) -> Result<serde_json::Value>
```
tcharding: Can de-dup this.

client/src/client_async/mod.rs
```diff
@@ -0,0 +1,171 @@
...
+where
+    T: serde::ser::Serialize,
+{
+    Ok(serde_json::to_value(val)?)
+}
+
+/// Helper to log an RPC response.
+fn log_response(method: &str, resp: &Result<jsonrpc::Response>) {
```
tcharding: And de-dup this?

client/src/client_async/error.rs
```diff
@@ -0,0 +1,41 @@
...
+    GetBlockHeaderError, GetBlockHeaderVerboseError, GetBlockVerboseOneError,
+    GetRawTransactionVerboseError,
+};
+use types::v19::GetBlockFilterError;
+use types::v29::{
+    GetBlockHeaderVerboseError as GetBlockHeaderVerboseErrorV29,
+    GetBlockVerboseOneError as GetBlockVerboseOneErrorV29,
+};
+
+pub use crate::error::{Error, UnexpectedServerVersionError};
```
tcharding: For a production client all this error stuff could be better I rekon.

client/src/client_async/rpcs.rs
```diff
@@ -0,0 +1,121 @@
...
+use super::{into_json, Client, Result};
+use crate::types::model::{GetBlockFilter, GetBlockHeaderVerbose, GetBlockVerboseOne};
+
+/// Bitcoin Core RPC methods (v25 to v30).
+///
+/// This trait exposes the Bitcoin Core RPC methods available on [`Client`]. Downstream users
+/// can define their own extension traits with additional methods without risk of
+/// name collision with the methods defined here.
+// `async fn` in traits produces non-`Send` futures by default; we suppress this lint because
+// `BitcoinRpcs` is intended only for use with concrete types (never `dyn BitcoinRpcs`).
```
tcharding: We can just put the subset of RPC methods we want directly on the `Client` type. Downstream can still define extension traits and name collisions can still be avoided by calling their trait methods using trait syntax or whatever its called.

Does BDK use the verbose methods? At first blush I don't like them, bit ugly for a 'bare bones client'.

tcharding: I rekon we add an example that adds an extension trait and a single method, then that can be used as documentation and to link LLMs too.

jamillambert: The reason for this change was to separate Client from the BDK RPCs. Otherwise we have a BDK client which can be modified for other projects, rather than the way it is here where it is a universal client with a trait premade for BDK.

So I still prefer to have no RPCs at all in the Client, and keep the full set of BDK RPCs in the prewritten trait.

I also think that having extra RPCs in this trait is a good idea rather than just one example and force everyone to write their own. If we put in all that BDK need it works out of the box for them, and if a few can be added to fill the needs of LDK and possibly there is only a few changes needed to this set to support multiple users. I understand the concern that supporting multiple projects will cause conflicts in the RPC methods if they are all together. But if we can support multiple projects without conflicts why not? i.e. The policy is we don't support adding downstream RPC requirements to the Client you need to add the trait to your own codebase, but on an individual basis we can consider if only a small change is required to support whole project then why not.

EDIT: AI says LDK only need `getblockchaininfo` and `gettxout` added to the trait so that is supports them too.

client/Cargo.toml
jamillambert: Done

client/src/client_async/mod.rs
jamillambert: Done

client/src/client_async/mod.rs
jamillambert: Done

client/src/client_async/mod.rs
jamillambert: Done

client/src/client_async/error.rs
jamillambert: I made the errors simpler and better.

client/src/client_async/rpcs.rs
tcharding: I agree with all that about supporting bdk and ldk. I rekon we should do so as methods on the client still though not a trait. We can chat about it on the phone tomorrow though.

tcharding:
Were you planning on doing the `corepc-regtest-client` rename as part of this or afterwards? At a cursory look everything looks good.

jamillambert:
> Were you planning on doing the `corepc-regtest-client` rename as part of this or afterwards? At a cursory look everything looks good.

I was planning on doing the rename afterwards since it also requires changing the current testing client.

tcharding:
Sweet. Whats left to undraft this then?

jamillambert:
> Sweet. Whats left to undraft this then?

I got it to work locally on BDK. It did require breaking changes due to being async, but I think this is the right choice. This testing resulted in the last patch to make the use in BDK easier, and probably other downstream too.

Undrafted.

client/src/client_async/mod.rs
```diff
@@ -0,0 +1,176 @@
...
+use std::path::PathBuf;
+
+pub use crate::client_async::error::{
+    Error, GetBestBlockHashError, GetBlockCountError, GetBlockError, GetBlockFilterError,
+    GetBlockHashError, GetBlockHeaderError, GetBlockHeaderVerboseError, GetBlockVerboseError,
+    GetBlockchainInfoError, GetRawMempoolError, GetRawTransactionError, GetTxOutError,
+    RpcMethodErrorExt, ServerVersionError, UnexpectedServerVersionError,
+};
+pub use crate::client_async::rpcs::RpcApi;
+pub(crate) use crate::{into_json, log_response};
```
tcharding: This import is unusual, why the `pub(crate)`?

integration_test/tests/bdk_client.rs
```diff
@@ -0,0 +1,180 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Tests for the async client.
+
+#![cfg(feature = "v30_and_below")]
+#![cfg(not(feature = "v24_and_below"))]
+#![allow(non_snake_case)] // Test names intentionally use double underscore.
```
tcharding: I think we would be better off without this.

integration_test/tests/bdk_client.rs
```diff
@@ -0,0 +1,180 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Tests for the async client.
+
+#![cfg(feature = "v30_and_below")]
+#![cfg(not(feature = "v24_and_below"))]
```
tcharding: Perhaps add a comment stating why we only test versions 25-30

integration_test/tests/bdk_client.rs
```diff
@@ -0,0 +1,180 @@
...
+async fn async__get_best_block_hash__modelled() {
+    let node = BitcoinD::with_wallet(Wallet::None, &[]);
+    let client = async_client_for(&node);
+
+    let model: Result<bitcoin::BlockHash, GetBestBlockHashError> =
+        client.get_best_block_hash().await;
+    let model = model.unwrap();
+    let expected = node.client.best_block_hash().expect("best_block_hash");
+    assert_eq!(model, expected);
+}
```
tcharding: What if we do this for the first test and for subsequent tests the same but with the comments?

```rust
async fn get_best_block_hash() {
    let node = BitcoinD::with_wallet(Wallet::None, &[]);
    let client = async_client_for(&node);

    // Tests that the async-client has this function.
    let got  = client.get_best_block_hash().await.unwrap();
    // Grabs the block hash using the sync client which we know works.
    let want = node.client.best_block_hash().expect("best_block_hash");

    // Tests tat the async client returned the same block hash as the async client.
    assert_eq!(got, want);
}
```

The reason is that we are expecting BDK devs to look at this test file. We are just testing the functionality and how usage of the client might look. I'm undecided about the explicit error stuff, leaning towards not testing it here and just making sure we get it right. There are not that many errors anyways. Open to your thoughts though?

tcharding:
Everything else looks good. The error stuff is pretty funky but I don't have a better solution right now. We can change it later if we want to.

tcharding:
Hey mate, I had a bit of a play because I wanted to get rid of the `Raw<Box>` stuff, in the end I did not get very far but a bunch of other little cleanups fell out: https://github.com/rust-bitcoin/corepc/pull/664

client/src/client_async/error.rs
```diff
@@ -0,0 +1,596 @@
...
+}
+
+/// Shared inspection helpers implemented by every async method-specific RPC error type.
+///
+/// The RPC methods continue to return their typed errors, and callers can opt into common handling
+/// without repetitive enum matching. Implementors only need to provide [`as_client_error`]; the
+/// remaining helpers have default implementations derived from it.
+///
+/// [`as_client_error`]: RpcMethodErrorExt::as_client_error
+pub trait RpcMethodErrorExt {
```
tcharding: Bro the extension patter is specifically for bypassing the orphan rule. It makes no sense to use it within a single crate.

jamillambert: Changed to inherent methods.

integration_test/tests/bdk_client.rs
jamillambert: I changed the tests to your example and assumed it was a typo and you meant without the comments for later tests?

integration_test/tests/bdk_client.rs
jamillambert: Comment added and bumped to v31

integration_test/tests/bdk_client.rs
jamillambert: I changed all the test names and removed the allow.

client/src/client_async/mod.rs
jamillambert: Done as part of taking all your suggested changes from #664

jamillambert:
> Hey mate, I had a bit of a play because I wanted to get rid of the `Raw<Box>` stuff, in the end I did not get very far but a bunch of other little cleanups fell out: #664

Thanks, I included all of your cleanups and redid the error handling.

tcharding (approved):
ACK cd189401707680d8f078818994f3f15ea7b2830c

tcharding:
Mad! Thanks for sticking with this mate. Sorry if my review the other day was a bit prickly. Nice to see this go in. Lets do the split into two crates and release this fella.

PR #506, opened 2026-02-12
WIP: Create an async corepc-client (adding jsonrpc support into the corepc-client crate)

jamillambert:
Not a merge candidate. Proof of concept only.

This is PR option 2 of 2, both add an async corepc-client but in a slightly different ways. **This version adds the required part of `jsonrpc` into `corepc-client`** the other version adds async support to jsonrpc (#505).

The patches are structured to make it easier to see what was changed from sync to async. First the sync version is copied and renamed, then in a separate patch it is changed to async.

The jsonrpc part that was added to corepc-client could instead be a separate crate, or if the rest of jsonrpc is not required it could be the replacement for it.


tcharding:
#505 is a better approach IMO.

PR #509, opened 2026-02-16
Add 'all' recipe to justfile

ton-anywhere:
Include a new Just Recipe to run all other recipes.
Useful to run locally before opening the pull request for review.

to run:
`    just all
`


jamillambert (commented):
I agree it would be better to add `just sane`. It keeps it in line with `rust-bitcoin` and commands like `fmt` and `update-lock-files` change the code so are better run separately as required.

tcharding:

PR #514, opened 2026-02-18
Remove concrete types from types/src/vXY

### LINKED ISSUE #511: `vXY::wallet` has concrete bitcoin type
`v19::wallet` should have `String` not `Transaction`

```rust
    /// The decoded transaction (only present when `verbose` is passed). v19 and later only.
    pub decoded: Option<Transaction>,
```

And v20, v23, v24, v26, v28

jamillambert:
Concrete bitcoin types should only be used in the modelled type. 

Go through every module file in `types/src/vXY` remove the `bitcoin::x` imports and fix the issues that causes or put back if required for other reasons:

- Move `GetBlockchainInfo::into_model()` to the `into` module to be consistent with other types.
- Remove concrete bitcoin type from `GetOrphanTxs`: Use `String` instead of `Txid`, add the error type and fix `into_model()`.
- Move bitcoin imports to `into` module: Some bitcoin types are only used in the into module but are imported in the main module. Move the imports to make it clearer that no concrete bitcoin types are used until the conversion into the modeled type.
- Remove concrete bitcoin type from `GetTransaction`: Use `String` instead of `Transaction` in `GetTransaction` and fix `into_model()`.

Closes #511

### DIFF
--- a/types/src/v19/wallet/into.rs
+++ b/types/src/v19/wallet/into.rs
@@ -73,6 +73,10 @@ impl GetTransaction {
             .into_iter()
             .map(|d| d.into_model().map_err(E::Details))
             .collect::<Result<Vec<_>, _>>()?;
+        let decoded = self
+            .decoded
+            .map(|tx| encode::deserialize_hex::<Transaction>(&tx).map_err(E::Tx))
+            .transpose()?;
 
         Ok(model::GetTransaction {
             amount,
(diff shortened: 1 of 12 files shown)


tcharding (dismissed):
ACK 3c7bbf7acc3c9592381cb5d87e9856998c496789

jamillambert:
Rebased to remove merge conflicts. No changes, conflicts were due to adjacent line changes.

PR #517, opened 2026-02-24
Fix a few missed cases of "vXY and later only" in types

jamillambert:
#513 missed a few cases of "vXY and later only" comments that had different wording and didn't come up in the grep. Do a more general search and fix what is found.

One of the fields in types for `PeerInfo` was marked as "(deprecated in v0.21)" and set as an option in v22 and v26. The field is not in the RPC docs for either v22 or v26 and the tests pass with it being removed.

- Remove the deprecated `whitelisted` field from `PeerInfo`.
- Tidy up the docs.


tcharding (approved):
ACK c803543ca850a051d2c24369d1e6008b94122fb8

PR #518, opened 2026-02-25
[bitreq]: Add blocking `Client` with connection pooling

tnull:
Previously, we added a `Client` allowing for connection reuse behind the `async` feature. Here we do the same for the blocking, i.e., non-`async` path.

### DIFF
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -15,8 +15,17 @@ use crate::{Error, Request, Response};
 
 /// A client that caches connections for reuse.
 ///
-/// The client maintains a pool of up to `capacity` connections, evicting
-/// the least recently used connection when the cache is full.
+/// The client maintains a cache of up to `max_idle` connections, evicting the least
+/// recently used entry when the cache is full.
+///
+/// # Bound applies to cached entries, not live connections
+///
+/// `max_idle` bounds the number of connections held in the cache, not the number of
+/// connections the client may have open at any one time. Concurrent requests whose
+/// cached connection is absent (or currently checked out for another in-flight
+/// request) each open a fresh socket; any surplus streams simply fail to re-enter
+/// the cache on put-back once it is full. To bound concurrency — rather than idle
+/// reuse — a caller must arrange a separate semaphore on top.
 ///
 /// # Example
 ///
@@ -24,7 +33,7 @@ use crate::{Error, Request, Response};
 /// # async fn request() {
 /// use bitreq::{Client, RequestExt};
 ///
-/// let client = Client::new(10); // Cache up to 10 connections
+/// let client = Client::new(10); // Cache up to 10 idle connections
 /// let response = bitreq::get("https://example.com")
 ///     .send_async_with_client(&client)
 ///     .await;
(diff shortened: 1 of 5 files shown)


oleonardolima (commented):
cACK 94c62c39a9a84e3e917af98f4016e7732f3243bb

It looks good so far, though I won't be able to do a in-depth review and testing until early next week.

TheBlueMatt (commented):
Hmm, while the code is fine, its really annoying that we're diverging so much between the sync and async code, and even duplicating a decent amount of logic between the sync and sync-connection-reuse logic. Maybe
 * for both sync and async we pass a reference to the `Client` in so that the `Connection` can lookup a connection to use on redirect rather than moving that logic out to `client.rs`.
 * Similar for connection timeouts, I'm fine with it being in `Connection` or in the `Client` but let's pick one and stick to it.

luisschwab:
Any progress on this one?

bitreq/src/client.rs
```diff
@@ -115,3 +123,177 @@ impl RequestExt for Request {
...
+/// ```no_run
+/// # fn main() -> Result<(), bitreq::Error> {
+/// use bitreq::{Client, RequestExt};
+///
+/// let client = Client::new(10); // Cache up to 10 connections
+/// let response = bitreq::get("http://example.com")
+///     .send_with_client(&client)?;
+/// # Ok(()) }
+/// ```
+#[cfg(not(feature = "async"))]
```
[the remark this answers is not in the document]
tnull: Yeah, okay, though this ofc means we'll always have the `blocking` path/APIs now.

tnull:
> Hmm, while the code is fine, its really annoying that we're diverging so much between the sync and async code, and even duplicating a decent amount of logic between the sync and sync-connection-reuse logic. Maybe
> 
> * for both sync and async we pass a reference to the `Client` in so that the `Connection` can lookup a connection to use on redirect rather than moving that logic out to `client.rs`.
> * Similar for connection timeouts, I'm fine with it being in `Connection` or in the `Client` but let's pick one and stick to it.

Now pushed some updates / rewritten commits. Please let me know if that's more in-line with what you had in mind.

> Any progress on this one?

Yes, excuse the delay.

tcharding:
About to release the stack in #560, want this in? Although we don't want to hold it up too much because of the RUSTSEC stuff, right?

tnull:
> About to release the stack in #560, want this in? Although we don't want to hold it up too much because of the RUSTSEC stuff, right?

Yeah, would be good to ship this one soon, too, but we should probably not hold up the release for it, given it annoys people that their CI is red constantly.

luisschwab:
Yep this can go on the next release. Red CI jobs are annoying 😆

bitreq/src/client.rs
```diff
@@ -1,113 +1,232 @@
...
 }
 
-struct ClientImpl<T> {
-    connections: HashMap<ConnectionKey, Arc<T>>,
-    lru_order: VecDeque<ConnectionKey>,
-    capacity: usize,
+struct ClientState {
+    blocking_connections: HashMap<ConnectionKey, PoolEntry>,
+    #[cfg(feature = "async")]
+    async_connections: HashMap<ConnectionKey, Arc<AsyncConnection>>,
```
TheBlueMatt, line 69: It seems surprising that a single `Client` can do both async and sync pooling but also doesn't reuse the connections across them (we can't, but from an API perspective, it feels very surprising).

bitreq/src/client.rs
```diff
@@ -1,113 +1,232 @@
...
+#[derive(Clone, PartialEq, Eq)]
+enum LruKey {
+    Blocking(ConnectionKey),
+    #[cfg(feature = "async")]
+    Async(ConnectionKey),
+}
+
+pub(crate) struct PoolEntry {
+    pub(crate) stream: HttpStream,
+    pub(crate) expires_at: Instant,
```
TheBlueMatt, line 84: You didn't address my second comment - currently for async we do the retry/new stream/expiry handling all in Connection. It seems a bit easier to keep it that way (because if we go to use a connection and the connection has dropped, we similarly need to make a new connection just like for async/the timed-out case), but I'm certainly okay with moving it out to `Client`. We should, however, do the same thing for both sync and async.

bitreq/src/connection.rs
```diff
@@ -748,27 +733,165 @@ impl Connection {
...
+    /// redirect, a pooled connection to the new host is reused if available;
+    /// otherwise a fresh connection is created.
+    pub(crate) fn send_pooled(
+        self,
+        client: &crate::Client,
+        mut request: ParsedRequest,
+    ) -> Result<Response, Error> {
+        let mut connection = self;
+        loop {
+            let (response, recovered, req) = connection.send_and_buffer(request)?;
```
TheBlueMatt, line 784: We need to handle if the server dropped the connection between when we did the last request and we started this one like we do in async.

tcharding:

PR #519, opened 2026-02-25
Model `merkle_root` as `TxMerkleNode`

jamillambert:
Some of the modelled types used a `String` for `merkle_root`. But it has a rust-bitcoin type `TxMerkleNode`.

Change them all to `TxMerkleNode`, and update the `into_model()` functions.

### DIFF
--- a/types/src/v29/blockchain/into.rs
+++ b/types/src/v29/blockchain/into.rs
@@ -23,6 +23,7 @@ impl GetBlockVerboseOne {
             self.stripped_size.map(|size| crate::to_u32(size, "stripped_size")).transpose()?;
         let weight = Weight::from_wu(self.weight);
         let version = block::Version::from_consensus(self.version);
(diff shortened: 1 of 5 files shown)


types/src/v29/blockchain/into.rs
```diff
@@ -23,6 +23,7 @@ impl GetBlockVerboseOne {
             self.stripped_size.map(|size| crate::to_u32(size, "stripped_size")).transpose()?;
         let weight = Weight::from_wu(self.weight);
         let version = block::Version::from_consensus(self.version);
+        let merkle_root = self.merkle_root.parse::<TxMerkleNode>().map_err(E::Hash)?;
```
tcharding: This change is not 100% correct. It made me look if we do this elsewhere and we do. Created: #521

tcharding (approved):
ACK 162a2df3ba7a767763d28851141c80660b7cf1b3

PR #522, opened 2026-02-26
Fix types conversion error variants

### LINKED ISSUE #521: One error variant per field
Each error variant, in my memory at least, should be for an individual field. I.e in `v29::blockchain::GetDescriptorActivity` we [ab]use the `Hash` variant of `GetDescriptorActivityError` for `spend_txid` and `prevout_txid`. While this works it means (and one could argue 'who cares') that a user trying to debug the error will have no idea _which_ field failed to parse. The rest of the crate was written explicitly to try and minimise this sort of confusion.

I don't know how many instances of this there are. I rekon an LLM could find them by prompting it with 'search every `into.rs` file for lines that inlude setting a local variable with a statement that uses `map(E:Foo)` and `Foo` does not match the local variable name eg it should be `let this_var = ...map(E::ThisVar)`.

jamillambert:
Each error variant should be for an individual field. Use an LLM to go through all into.rs files to address the issue, spend longer than doing the job from scratch myself fixing all its errors.

- Remove a redefinition of an error in v30 that is unchanged from the v29 definition.
- Add new variants where one variant was used for more than one field. Rename existing variants and update error messages to match the field names.

Closes #521

### DIFF
--- a/types/src/v17/blockchain/error.rs
+++ b/types/src/v17/blockchain/error.rs
@@ -113,15 +113,14 @@ pub enum GetBlockHeaderError {
     /// Conversion of hex data to bytes failed.
     Hex(hex::HexToBytesError),
     /// Consensus decoding of bytes to header failed.
-    Consensus(encode::Error),
+    Header(encode::Error),
 }
 
 impl fmt::Display for GetBlockHeaderError {
     fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
         match *self {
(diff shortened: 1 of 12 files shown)


tcharding (dismissed):
ACK 43ea5f05b9c13d1aa8dc4634b78be5bf35007454

tcharding (approved):
ACK 9ca8e04aa685c0ec5f0ff3b55e8b2db2baf98091

PR #524, opened 2026-02-26
Use consistent capitalization for (Script)PubKey

### LINKED ISSUE #371: Use the same capitalisation as `rust-bitcoin` in identifiers that include 'pubkey'
We have `ScriptPubkey` but it looks like `rust-bitcoin` is going to use `ScriptPubKey` - we should use the same here. May not need changing until we upgrade `rust-bitcoin` version, I did not look. Wanted to flag the issue anyways.

https://github.com/rust-bitcoin/rust-bitcoin/discussions/4950

jamillambert:
The capitalization of key in pubkey and scriptpubkey is inconsistent. Core uses a capital K in e.g. scriptPubKey in gettxout, and a capital K is more common in rust-bitcoin.

Search the whole codebase for "Pubkey" and change them all to "PubKey". No changes made where the p is lowercase.

Closes #371

### DIFF
--- a/types/src/model/wallet.rs
+++ b/types/src/model/wallet.rs
@@ -171,10 +171,10 @@ pub struct GetAddressInfo {
 pub enum ScriptType {
     /// Non-standard output script type.
     NonStandard,
-    /// Pubkey output script.
-    Pubkey,
-    /// Pubkey hash output script.
-    PubkeyHash,
+    /// PubKey output script.
+    PubKey,
+    /// PubKey hash output script.
+    PubKeyHash,
     /// Script hash output script.
     ScriptHash,
     /// Multisig output script.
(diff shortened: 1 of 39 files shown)


types/src/model/wallet.rs
```diff
@@ -171,10 +171,10 @@ pub struct GetAddressInfo {
 pub enum ScriptType {
     /// Non-standard output script type.
     NonStandard,
-    /// Pubkey output script.
-    Pubkey,
-    /// Pubkey hash output script.
-    PubkeyHash,
+    /// PubKey output script.
```
tcharding: I'm surprised the linter doesnt' complain about this? If it starts doing so I'm inclined to think we should just use 'pubkey' and 'Pubkey' (capitalized) in docs and not add an attribute to shoosh the linter.

Issue #472, opened 2026-01-23
feat(bitreq): HTTPS Proxy Support for HTTP CONNECT


DanGould:
## Context

We’re working on migrating https://github.com/payjoin/rust-payjoin/issues/1288 from reqwest to bitreq to reduce our dependency footprint. One critical requirement is the ability to proxy HTTPS requests through an HTTP proxy using the CONNECT method.

In payjoin, we use this to protect user IP privacy when [fetching OHTTP keys from a payjoin directory](https://github.com/payjoin/rust-payjoin/blob/master/payjoin/src/core/io.rs):

```rs
// Current reqwest code (payjoin/src/core/io.rs)
let proxy = Proxy::all(ohttp_relay.into_url()?.as_str())?;
let client = Client::builder()
    .proxy(proxy)
    .http1_only()
    .build()?;
let res = client
    .get(ohttp_keys_url) // This is an HTTPS URL
    .timeout(Duration::from_secs(10))
    .header(ACCEPT, "application/ohttp-keys")
    .send()
    .await?;
```

The ohttp_keys_url is an HTTPS endpoint (e.g., https://directory.example.com/.well-known/ohttp-gateway), and we fetch it through an HTTP CONNECT proxy to avoid revealing the client’s IP to the directory. We need the inner authenticated TLS connection from client to https://directory.example.com to prevent the proxy from lying and serving the client the proxy's own private key. That's the whole point of the bootstrapping mechanism. The wrapped encrypted connection from client to proxy would be nice to have but not critical. Same with the connection from proxy to https://directory.example.com : best practice, but not critical for our use case.

We don't want to reveal to the proxy which key we got, so HTTPS helps.

## Desired Solution

We want to verify and document whether bitreq’s Proxy::new_http() supports proxying HTTPS destinations, not just HTTP.

The current bitreq docs state:

“Only HTTP CONNECT proxies are supported.”

It’s ambiguous whether this means:

The proxy protocol is HTTP (i.e., connecting to the proxy server over HTTP) — which is what we need, or

Only HTTP (non-HTTPS) destination requests are supported through the proxy.

If HTTPS destinations via HTTP CONNECT are already supported, it should be clarified in the documentation. If not, this would be a valuable feature to add, enabling us to upgrade.

## Alternatives Considered

Staying on reqwest (but this brings ~120 transitive dependencies).

## Additional Context

[...]

PR #533, opened 2026-03-20
bitreq: add SOCKS5 proxy support

FreeOnlineUser:
Add SOCKS5 proxy support (RFC 1928) alongside the existing HTTP CONNECT path.

## Motivation

ldk-node needs to route HTTP calls (RGS, scoring, LNURL-auth) through Tor's SOCKS proxy when TorConfig is enabled. The existing `proxy` feature only supports HTTP CONNECT, which doesn't handle `.onion` addresses. See lightningdevkit/ldk-node#834.

SOCKS5 and HTTP CONNECT serve different use cases and now coexist behind the same `proxy` feature flag:
- **SOCKS5**: Tor and general proxying with domain-based addressing (DNS resolved at the proxy).
- **HTTP CONNECT**: forward proxies for HTTPS tunneling. Kept for the payjoin use case raised in rust-bitcoin/corepc#472.

## Public API

- `Proxy::new_socks5(addr)`: unauthenticated SOCKS5. Accepts both `socks5://` and `socks5h://` URL schemes (identical behaviour, since destinations are always forwarded as ATYP 0x03 hostnames).
- `Proxy::new_socks5_with_credentials(addr, user, pass)`: RFC 1929 username/password auth at construction. Tor uses these credentials for circuit isolation.
- `Proxy::set_credentials(user, pass)`: mutate credentials on an existing `Proxy`. Lets a caller hold one `Proxy` and rotate credentials between connections to obtain fresh Tor circuits, without rebuilding from a URL.

## Implementation

- New `ProxyKind::Socks5` variant.
- Sync and async handshake paths share protocol logic via 5 extracted helpers (greeting, auth, connect framing). Only the I/O calls differ between sync and async.
- Connection layer branches on `proxy.is_socks5()` before the existing HTTP CONNECT path.

## Tests

18 SOCKS5-specific tests:
- 10 parsing / credential: host+port parsing, `socks5://` and `socks5h://` schemes, default port (1080), wrong-protocol rejection, `is_socks5` check, credential constructor, RFC 1929 length validation (constructor and setter), credential rotation.
- 8 mock handshake: success, `.onion` domain, server rejection, port encoding (big-endian), domain >255 bytes, RFC 1929 credential auth, credential rejection, no-auth fallback.

Unit tests, doctests, and clippy are all clean.

## Notes

- Production tested on a live Lightning node routing through Arti's SOCKS proxy.
- Related: rust-bitcoin/corepc#472 (original payjoin SOCKS5 ask), lightningdevkit/ldk-node#834 (downstream consumer).
- Read-timeout behaviour during the binary handshake matches the existing HTTP CONNECT path: both rely on the connect-time deadline and don't set per-read deadlines on the raw stream. Improving this for both paths is left as a follow-up.

---

Assisted by Joe (Claude Opus 4.7 under the hood).

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -13,8 +13,6 @@ use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 use std::time::Instant;
 
-#[cfg(all(feature = "async", feature = "proxy"))]
-use tokio::io::AsyncReadExt;
 #[cfg(feature = "async")]
 use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt, ReadHalf, WriteHalf};
 #[cfg(feature = "async")]
@@ -337,36 +335,8 @@ impl AsyncConnection {
         #[cfg(feature = "proxy")]
         match &params.proxy {
             Some(proxy) => {
-                // do proxy things
                 let mut tcp = Self::tcp_connect(&proxy.server, proxy.port).await?;
-
-                let proxy_request = proxy.connect(params.host, params.port);
-                tcp.write_all(proxy_request.as_bytes()).await?;
-                tcp.flush().await?;
-
-                // Max proxy response size to prevent unbounded memory allocation
-                const MAX_PROXY_RESPONSE_SIZE: usize = 16 * 1024;
-                let mut proxy_response = Vec::new();
-                let mut buf = [0; 256];
-
-                loop {
-                    let n = tcp.read(&mut buf).await?;
-                    if n == 0 {
-                        // EOF reached
-                        break;
-                    }
-                    proxy_response.extend_from_slice(&buf[..n]);
-                    if proxy_response.len() > MAX_PROXY_RESPONSE_SIZE {
-                        return Err(Error::ProxyConnect);
-                    }
-                    if n < buf.len() {
-                        // Partial read indicates end of response
-                        break;
-                    }
-                }
-
-                crate::Proxy::verify_response(&proxy_response)?;
-
+                proxy.handshake_async(&mut tcp, params.host, params.port).await?;
                 Ok(tcp)
             }
             None => Self::tcp_connect(params.host, params.port).await,
@@ -710,35 +680,8 @@ impl Connection {
         #[cfg(feature = "proxy")]
         match &params.proxy {
             Some(proxy) => {
-                // do proxy things
                 let mut tcp = Self::tcp_connect(&proxy.server, proxy.port, timeout_at)?;
-
-                write!(tcp, "{}", proxy.connect(params.host, params.port))?;
-                tcp.flush()?;
-
-                // Max proxy response size to prevent unbounded memory allocation
-                const MAX_PROXY_RESPONSE_SIZE: usize = 16 * 1024;
-                let mut proxy_response = Vec::new();
-                let mut buf = [0; 256];
-
-                loop {
-                    let n = tcp.read(&mut buf)?;
-                    if n == 0 {
-                        // EOF reached
-                        break;
-                    }
-                    proxy_response.extend_from_slice(&buf[..n]);
-                    if proxy_response.len() > MAX_PROXY_RESPONSE_SIZE {
-                        return Err(Error::ProxyConnect);
-                    }
-                    if n < buf.len() {
-                        // Partial read indicates end of response
-                        break;
-                    }
-                }
-
-                crate::Proxy::verify_response(&proxy_response)?;
-
+                proxy.handshake_sync(&mut tcp, params.host, params.port)?;
                 Ok(tcp)
             }
             None => Self::tcp_connect(params.host, params.port, timeout_at),
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -13,8 +13,6 @@ use std::sync::{Arc, Mutex};
 use std::task::{Context, Poll};
 use std::time::Instant;
 
-#[cfg(all(feature = "async", feature = "proxy"))]
-use tokio::io::AsyncReadExt;
 #[cfg(feature = "async")]
 use tokio::io::{AsyncRead, AsyncWrite, AsyncWriteExt, ReadHalf, WriteHalf};
 #[cfg(feature = "async")]
@@ -353,36 +351,8 @@ impl AsyncConnection {
         #[cfg(feature = "proxy")]
         match &params.proxy {
             Some(proxy) => {
-                // do proxy things
                 let mut tcp = Self::tcp_connect(&proxy.server, proxy.port).await?;
-
-                let proxy_request = proxy.connect(params.host, params.port);
-                tcp.write_all(proxy_request.as_bytes()).await?;
-                tcp.flush().await?;
-
-                // Max proxy response size to prevent unbounded memory allocation
-                const MAX_PROXY_RESPONSE_SIZE: usize = 16 * 1024;
-                let mut proxy_response = Vec::new();
-                let mut buf = [0; 256];
-
-                loop {
-                    let n = tcp.read(&mut buf).await?;
-                    if n == 0 {
-                        // EOF reached
-                        break;
-                    }
-                    proxy_response.extend_from_slice(&buf[..n]);
-                    if proxy_response.len() > MAX_PROXY_RESPONSE_SIZE {
-                        return Err(Error::ProxyConnect);
-                    }
-                    if n < buf.len() {
-                        // Partial read indicates end of response
-                        break;
-                    }
-                }
-
-                crate::Proxy::verify_response(&proxy_response)?;
-
+                proxy.handshake_async(&mut tcp, params.host, params.port).await?;
                 Ok(tcp)
             }
             None => Self::tcp_connect(params.host, params.port).await,
@@ -728,35 +698,8 @@ impl Connection {
         #[cfg(feature = "proxy")]
         match &params.proxy {
             Some(proxy) => {
-                // do proxy things
                 let mut tcp = Self::tcp_connect(&proxy.server, proxy.port, timeout_at)?;
-
-                write!(tcp, "{}", proxy.connect(params.host, params.port))?;
-                tcp.flush()?;
-
-                // Max proxy response size to prevent unbounded memory allocation
-                const MAX_PROXY_RESPONSE_SIZE: usize = 16 * 1024;
-                let mut proxy_response = Vec::new();
-                let mut buf = [0; 256];
-
-                loop {
-                    let n = tcp.read(&mut buf)?;
-                    if n == 0 {
-                        // EOF reached
-                        break;
-                    }
-                    proxy_response.extend_from_slice(&buf[..n]);
-                    if proxy_response.len() > MAX_PROXY_RESPONSE_SIZE {
-                        return Err(Error::ProxyConnect);
-                    }
-                    if n < buf.len() {
-                        // Partial read indicates end of response
-                        break;
-                    }
-                }
-
-                crate::Proxy::verify_response(&proxy_response)?;
-
+                proxy.handshake_sync(&mut tcp, params.host, params.port)?;
                 Ok(tcp)
             }
             None => Self::tcp_connect(params.host, params.port, timeout_at),
--- a/bitreq/src/proxy.rs
+++ b/bitreq/src/proxy.rs
@@ -2,18 +2,25 @@ use base64::engine::general_purpose::STANDARD;
 use base64::engine::Engine;
 
 use crate::error::Error;
+use crate::response::maybe_await;
 
-/// Kind of proxy connection (Basic, Digest, etc)
+/// Kind of proxy connection (Basic, Digest, SOCKS5, etc)
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub(crate) enum ProxyKind {
     Basic,
+    Socks5,
 }
 
-/// Proxy configuration. Only HTTP CONNECT proxies are supported (no SOCKS or
-/// HTTPS).
+/// Proxy configuration. Supports HTTP CONNECT proxies ([`Proxy::new_http`])
+/// and SOCKS5 proxies ([`Proxy::new_socks5`]).
 ///
-/// When credentials are provided, the Basic authentication type is used for
-/// Proxy-Authorization.
+/// SOCKS5 encodes targets per RFC 1928: IPv4 literals as ATYP 0x01, IPv6
+/// literals as ATYP 0x04, and anything else as a domain name (ATYP 0x03)
+/// with DNS resolution left to the proxy. The latter enables routing
+/// through Tor, including `.onion` addresses.
+///
+/// For HTTP CONNECT proxies, when credentials are provided, the Basic
+/// authentication type is used for Proxy-Authorization.
 #[derive(Clone, Debug, Hash, PartialEq, Eq)]
 pub struct Proxy {
     pub(crate) server: String,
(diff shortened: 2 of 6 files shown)


tnull (commented):
Took a first very high-level look.

bitreq/src/proxy.rs
```diff
@@ -86,6 +91,50 @@ impl Proxy {
...
+    /// The default port is 1080.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let proxy = bitreq::Proxy::new_socks5("127.0.0.1:9050").unwrap();
+    /// let request = bitreq::post("http://example.com").with_proxy(proxy);
+    /// ```
+    ///
+    pub fn new_socks5<S: AsRef<str>>(proxy: S) -> Result<Self, Error> {
```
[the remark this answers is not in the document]
FreeOnlineUser: Good call. I'll add username/password support (RFC 1929 sub-negotiation) to the SOCKS5 handshake. When credentials are set, the greeting advertises method 0x02 instead of 0x00, and Tor uses the unique credentials to isolate circuits. The Proxy API can accept optional credentials at construction or via a setter for rotation.

tnull:
@FreeOnlineUser any update on this? Is this ready to take out of draft?

daywalker90:
FYI: https://github.com/rust-bitcoin/bitcoin-payment-instructions needs `socks5h` support for remote DNS lookups.

FreeOnlineUser:
@tnull yes, ready. Apologies for the radio silence. Pulling threads together:

**HTTP CONNECT path** (your March 20 question on dropping it): kept it. payjoin in rust-bitcoin/corepc#472 was the original requester for proxy support and they use HTTP CONNECT. Happy to drop in a follow-up PR if you'd rather, but it's separable from this work and doesn't block SOCKS5 going green.

**Credential refresh interface** (your March 20 ask for rotating credentials freely): added `Proxy::set_credentials(user, pass)`. A caller can now hold one long-lived `Proxy` and rotate credentials per connection for Tor circuit isolation without rebuilding from a URL. Tested with a rotation case. Open to also adding a chained `with_credentials(self, ...)` builder if you'd prefer that shape too.

**`socks5h` scheme** (@daywalker90's hint about `bitcoin-payment-instructions` needing remote DNS): we already always do remote DNS at the proxy via ATYP 0x03 (DOMAIN). I've added `socks5h://` as an accepted URL alias to match curl convention. Both schemes behave identically and the doc explicitly states this.

**Hang risk on the binary handshake**: this PR's read-timeout behaviour matches the existing HTTP CONNECT path, which also doesn't set per-read deadlines on the raw stream during proxy negotiation. Improving this for both paths is left as a follow-up so we don't expand scope here.

The description has been refreshed to reflect everything currently shipped (RFC 1929 credentials, `socks5h://`, `set_credentials`, 18 SOCKS5-specific tests). Out of draft.

bitreq/src/connection.rs
```diff
@@ -712,8 +715,14 @@ impl Connection {
...
+            Some(proxy) if proxy.is_socks5() => {
+                // SOCKS5 proxy
+                let mut tcp = Self::tcp_connect(&proxy.server, proxy.port, timeout_at)?;
+                proxy.socks5_handshake_sync(&mut tcp, params.host, params.port)?;
+                Ok(tcp)
+            }
             Some(proxy) => {
-                // do proxy things
+                // HTTP CONNECT proxy
                 let mut tcp = Self::tcp_connect(&proxy.server, proxy.port, timeout_at)?;
```
TheBlueMatt, line 701: Mind refactoring this code so that it just always calls `handshake`... and doesn't care what type it is and that's abstracted by the proxy?

bitreq/src/proxy.rs
```diff
@@ -115,6 +275,131 @@ impl Proxy {
...
+        // Drain the bound address
+        match connect_resp[3] {
+            0x01 => { // IPv4: 4 bytes + 2 port
+                let mut buf = [0u8; 6];
+                stream.read_exact(&mut buf).map_err(Error::IoError)?;
+            }
+            0x03 => { // Domain: 1 len byte + domain + 2 port
+                let mut len = [0u8; 1];
+                stream.read_exact(&mut len).map_err(Error::IoError)?;
+                let mut buf = vec![0u8; len[0] as usize + 2];
```
TheBlueMatt: please dont allocate to throw away <= 258 bytes.

bitreq/src/proxy.rs
```diff
@@ -115,6 +275,131 @@ impl Proxy {
             _ => Err(Error::BadProxy),
         }
     }
+
+    /// Perform a SOCKS5 handshake on a connected TCP stream (sync).
+    #[cfg(feature = "std")]
+    pub(crate) fn socks5_handshake_sync(
```
TheBlueMatt: Can we DRY the handshake method? Maybe with a simple macro generating both the async and sync versions?

FreeOnlineUser:
@TheBlueMatt thanks for the review. All three addressed in 8cfda31.

**Allocation (proxy.rs:329):** Switched the domain-ATYP arm to `[u8; 257]` with a slice. Both sync and async paths fixed; matches the IPv4 / IPv6 arms in the same match.

**Polymorphism (connection.rs):** Added `Proxy::handshake_sync` / `Proxy::handshake_async` `pub(crate)` dispatchers, with private `http_connect_handshake_sync` / `_async` methods owning the moved CONNECT logic. The connection layer now treats every proxy uniformly:

```rust
Some(proxy) => {
    let mut tcp = Self::tcp_connect(&proxy.server, proxy.port, timeout_at)?;
    proxy.handshake_sync(&mut tcp, params.host, params.port)?;
    Ok(tcp)
}
```

The `is_socks5()` predicate has no remaining lib callers and was dropped. Added two mock-server tests for the HTTP CONNECT path through the new dispatcher.

**Sync/async DRY (proxy.rs:281):** Done with two `macro_rules!` macros (`socks5_handshake_body!` and `http_connect_handshake_body!`) at module scope. Each takes the call-site idents (`self`, `stream`, `target_host`, `target_port`) and a token-tree `$($maybe_await:tt)*` that's empty for sync and `.await` for async, inserted at every I/O call. Macro hygiene means the parameter list is explicit, but the body is single-source.

The four handshake methods now look like:

```rust
pub(crate) fn socks5_handshake_sync(
    &self, stream: &mut std::net::TcpStream,
    target_host: &str, target_port: u16,
) -> Result<(), Error> {
    use std::io::{Read, Write};
    socks5_handshake_body!(self, stream, target_host, target_port;)
}
```

Net effect: -59 lines in proxy.rs versus the previous round, and any future protocol change happens in one place.

Also picked up a dead `tokio::io::AsyncReadExt` import that fell out of the connection.rs rewrite.

bitreq/src/proxy.rs
```diff
@@ -23,6 +28,99 @@ pub struct Proxy {
...
+        }
+
+        Ok(())
+    }};
+}
+
+/// HTTP CONNECT handshake body, parameterised over sync/async.
+///
+/// See `socks5_handshake_body!` for the dispatch convention.
+macro_rules! http_connect_handshake_body {
```
TheBlueMatt, line 104: This appears to be unused.

FreeOnlineUser: The macro is used at proxy.rs:436 (`http_connect_handshake_sync`) and :448 (`http_connect_handshake_async`). The proxy module is gated on `feature = "proxy"` and `proxy = ["base64", "std"]` in Cargo.toml, so the `std` wrapper compiles whenever the macro does. `cargo check` warning-free on `--features proxy`, `--features async,proxy`, and `--all-features`.

If the underlying point is that the abstraction is too thin (~25-line body vs `socks5_handshake_body!`'s ~60), happy to inline. If it's the broader HTTP CONNECT path (same direction tnull raised in March), happy to scope removal as a follow-up. Let me know which.

TheBlueMatt: Errrrrr, I'm not really sure how search failed me here, apologies.

bitreq/src/proxy.rs
```diff
@@ -23,6 +28,99 @@ pub struct Proxy {
...
+/// SOCKS5 handshake body, parameterised over sync/async.
+///
+/// Pass `.await` for the async version, nothing for the sync version. The
+/// caller's `use` imports decide which `Read`/`Write` trait set is in scope,
+/// so the same body resolves to either `std::io` or `tokio::io` calls.
+///
+/// Macro hygiene means the call site must pass `self`, `stream`, and the
+/// target host/port explicitly.
+macro_rules! socks5_handshake_body {
+    ($self:ident, $stream:ident, $target_host:ident, $target_port:ident; $($maybe_await:tt)*) => {{
```
TheBlueMatt: nit: I believe this would be more correct with a ? not a * (which would ripple through the whole macro).

```suggestion
    ($self:ident, $stream:ident, $target_host:ident, $target_port:ident; $($maybe_await:tt)?) => {{
```

bitreq/src/proxy.rs
```diff
@@ -23,6 +28,99 @@ pub struct Proxy {
...
+        let (greeting, expected_method) = $self.socks5_greeting();
+        $stream.write_all(&greeting) $($maybe_await)*.map_err(Error::IoError)?;
+        $stream.flush() $($maybe_await)*.map_err(Error::IoError)?;
+
+        let mut greeting_resp = [0u8; 2];
+        $stream.read_exact(&mut greeting_resp) $($maybe_await)*.map_err(Error::IoError)?;
+        Self::socks5_check_greeting(&greeting_resp, expected_method)?;
```
TheBlueMatt: Doesn't really seem worth stubbing out to a function to check two bytes and return an error?

bitreq/src/proxy.rs
```diff
@@ -23,6 +28,99 @@ pub struct Proxy {
...
+            let n = $stream.read(&mut buf) $($maybe_await)*.map_err(Error::IoError)?;
+            if n == 0 {
+                break;
+            }
+            proxy_response.extend_from_slice(&buf[..n]);
+            if proxy_response.len() > MAX_PROXY_RESPONSE_SIZE {
+                return Err(Error::ProxyConnect);
+            }
+            if n < buf.len() {
+                // Partial read indicates end of response.
```
TheBlueMatt: No, it might just indicate we received one TCP packet and are waiting on more.

TheBlueMatt (commented):
Sorry for the delay, a few nits and one bug.

bitreq/src/proxy.rs
FreeOnlineUser: Worth raising the bigger question: should HTTP CONNECT come out entirely?

A few reasons:
- Making this read loop correct isn't trivial (Content-Length, chunked encoding, or read-until-close, each with their own edge cases).
- tnull floated removal back in March. I argued to keep it for payjoin-style HTTP-only callers at the time, but I don't have a concrete use case (ldk-node Tor is SOCKS5-only).
- Removing it deletes the second `macro_rules!` block, so the abstraction question on the SOCKS5 macro stands on its own.

I'm happy to either fix the response-read here or strip HTTP CONNECT in this PR. Slight lean toward removal. Let me know which way you'd prefer.

bitreq/src/proxy.rs
FreeOnlineUser: There's already a precedent for this in bitreq: `response.rs` defines a `maybe_await!` helper used by `define_read_methods!` to thread `.await` through a sync/async-templated body, with the outer macro taking the keyword via `$(, $await: tt)?`. Refactored both `socks5_handshake_body!` and `http_connect_handshake_body!` to use the same shape:

```rust
macro_rules! socks5_handshake_body {
    ($self:ident, $stream:ident, $target_host:ident, $target_port:ident $(, $await:tt)?) => {{
        ...
        maybe_await!($stream.write_all(&greeting), $($await)?).map_err(Error::IoError)?;
        ...
    }};
}

// sync:  socks5_handshake_body!(self, stream, target_host, target_port)
// async: socks5_handshake_body!(self, stream, target_host, target_port, await)
```

Aligns with the existing `define_read_methods!` convention. Pushed in `224343b`.

bitreq/src/proxy.rs
FreeOnlineUser: Inlined both 2-byte helpers (`socks5_check_greeting`, `socks5_check_auth`) into the handshake macro body and removed the standalone fns. Pushed in `224343b`.

bitreq/src/proxy.rs
```diff
@@ -86,6 +192,139 @@ impl Proxy {
...
+        req.push(0x01); // sub-negotiation version
+        req.push(user.len() as u8);
+        req.extend_from_slice(user.as_bytes());
+        req.push(pass.len() as u8);
+        req.extend_from_slice(pass.as_bytes());
+        Some(req)
+    }
+
+    /// Build the SOCKS5 CONNECT request for a domain target.
+    fn socks5_connect_request(target_host: &str, target_port: u16) -> Result<Vec<u8>, Error> {
```
TheBlueMatt: Please do not allocate a vec for < 8192 bytes. Same goes for `socks5_auth_request`.

bitreq/src/proxy.rs
TheBlueMatt: That's a fair question, I wouldn't be too against removing it, but we do also actually have HTTP response-parsing in this crate....maybe we should find a way to use the `Response` type we already have to read the headers so that the code here can be ~trivial?

bitreq/src/proxy.rs
FreeOnlineUser: Switched both to fixed-size stack buffers:
- `socks5_auth_request` → `([u8; 513], usize)` (max 3 + 255 + 255)
- `socks5_connect_request` → `([u8; 262], usize)` (max 4 + 1 + 255 + 2)

Sliced with the length at the call site. Pushed in `69b50bc`.

bitreq/src/proxy.rs
FreeOnlineUser: Took the latter path. Made `parse_status_line` `pub(crate)` and rewrote the handshake to read into an `[u8; 8192]` stack buffer until `\r\n\r\n`, then parse the status line through the shared helper:

```rust
let mut buf = [0u8; 8192];
let mut len = 0;
let header_end = loop {
    let n = maybe_await!($stream.read(&mut buf[len..]), $($await)?).map_err(Error::IoError)?;
    if n == 0 { return Err(Error::ProxyConnect); }
    len += n;
    if let Some(idx) = buf[..len].windows(4).position(|w| w == b"\r\n\r\n") {
        break idx;
    }
    if len == buf.len() { return Err(Error::ProxyConnect); }
};

let headers = core::str::from_utf8(&buf[..header_end]).map_err(|_| Error::ProxyConnect)?;
let status_line = headers.lines().next().ok_or(Error::ProxyConnect)?;
let (status_code, _) = crate::response::parse_status_line(status_line);
```

Drops `verify_response` (the partial-read-as-EOF bug goes with it) and the heap `Vec`. Added a `handshake_handles_split_response` test that writes the response in two TCP segments to exercise the loop. Pushed in `69b50bc`.

bitreq/src/proxy.rs
```diff
@@ -86,6 +197,149 @@ impl Proxy {
...
+        proxy: S,
+        user: &str,
+        password: &str,
+    ) -> Result<Self, Error> {
+        let mut p = Self::new_socks5(proxy)?;
+        p.set_credentials(user, password)?;
+        Ok(p)
+    }
+
+    /// Sets RFC 1929 username/password credentials on this proxy.
```
TheBlueMatt: This also applies to HTTP proxies, though, so calling out the RFC number is wrong. It should ideally allow for longer user/password in HTTP mode and the docs should be clearer.

bitreq/src/proxy.rs
```diff
@@ -86,6 +197,149 @@ impl Proxy {
...
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// let mut proxy = bitreq::Proxy::new_socks5("127.0.0.1:9050").unwrap();
+    /// proxy.set_credentials("session-1", "x").unwrap();
+    /// // ... later, for a fresh circuit:
+    /// proxy.set_credentials("session-2", "x").unwrap();
+    /// ```
+    pub fn set_credentials(&mut self, user: &str, password: &str) -> Result<(), Error> {
```
TheBlueMatt: If its gonna make them a `String`, it should take them as a `String`, not `&str`.

bitreq/src/proxy.rs
```diff
@@ -86,6 +197,149 @@ impl Proxy {
...
+        }
+        self.user = Some(user.to_string());
+        self.password = Some(password.to_string());
+        Ok(())
+    }
+
+    /// Build the SOCKS5 greeting bytes.
+    /// Returns (greeting_bytes, expected_auth_method).
+    fn socks5_greeting(&self) -> ([u8; 3], u8) {
+        let method = if self.user.is_some() { 0x02 } else { 0x00 };
```
TheBlueMatt, line 318: what about empty-user-some-password?

bitreq/src/proxy.rs
```diff
@@ -86,6 +197,149 @@ impl Proxy {
...
+    fn socks5_connect_request(
+        target_host: &str,
+        target_port: u16,
+    ) -> Result<([u8; 262], usize), Error> {
+        let host_bytes = target_host.as_bytes();
+        if host_bytes.len() > 255 {
+            return Err(Error::ProxyConnect);
+        }
+        let mut buf = [0u8; 262];
+        buf[..5].copy_from_slice(&[0x05, 0x01, 0x00, 0x03, host_bytes.len() as u8]);
```
TheBlueMatt: Are we allowed to "encode" an IP by just writing it out as a "domain" here, or do we need to encode it properly?

TheBlueMatt (commented):
A few last comments, otherwise lgtm.

bitreq/src/proxy.rs
FreeOnlineUser: Spec answer: ATYP 0x03 with a stringified IP is technically allowed (the field is opaque bytes), but de facto standard is to detect and encode properly.

Prior art:
- curl (`lib/socks.c:780-891`): tries `inet_pton(AF_INET6, ...)` then `AF_INET`, falls through to domain
- Go `x/net/internal/socks`: `net.ParseIP(host)` then branches on `nil`/`To4()`/IPv6
- `sfackler/rust-socks` and `sticnarf/tokio-socks`: encode the choice in the type system at API entry (`TargetAddr::Ip | Domain` enum + conversion trait)

Practical impact for our targets is low (Tor recognizes IP literals and doesn't DNS them), but conformance is the right default and it mirrors the response parser at lines 77-89, which already handles all three ATYPs.

Sketch:

```rust
let body_end = match target_host.parse::<IpAddr>() {
    Ok(IpAddr::V4(v4)) => {
        buf[3] = 0x01;
        buf[4..8].copy_from_slice(&v4.octets());
        8
    }
    Ok(IpAddr::V6(v6)) => {
        buf[3] = 0x04;
        buf[4..20].copy_from_slice(&v6.octets());
        20
    }
    Err(_) => {
        let host_bytes = target_host.as_bytes();
        if host_bytes.len() > 255 { return Err(Error::ProxyConnect); }
        buf[3] = 0x03;
        buf[4] = host_bytes.len() as u8;
        buf[5..5 + host_bytes.len()].copy_from_slice(host_bytes);
        5 + host_bytes.len()
    }
};
```

`[u8; 262]` buffer unchanged (domain max still wins). Existing test at line 624 asserts `0x03`; would parametrize over the three ATYPs.

Happy to fold this into the same push as the other May 21 nits.

TheBlueMatt: Yea, if everyone else does it probably let's do it too.

bitreq/src/proxy.rs
FreeOnlineUser: Agreed. Dropped the RFC 1929 framing from the opener and split the length-limit guidance by proxy kind: SOCKS5 enforces 1-255 / 0-255 (RFC 1929 wire-format), HTTP Basic (RFC 7617) has no protocol-level limit. New test `http_set_credentials_no_length_limit` exercises the HTTP path with 1024-byte credentials.

bitreq/src/proxy.rs
FreeOnlineUser: Done. `set_credentials` now takes `String`; `new_socks5_with_credentials` cascades the same change so the constructor and the setter agree. Callers can hand over an existing allocation. Existing tests and the doc example updated.

bitreq/src/proxy.rs
FreeOnlineUser: Rejected, matching RFC 1929 ULEN >= 1. The existing validation test covered `("", "pass")` but didn't name the empty-user-non-empty-password case explicitly; added that assertion next to it (and the equivalent constructor case). Happy to relax if you'd prefer Tor-style opaque credentials where ULEN=0 is acceptable.

bitreq/src/proxy.rs
FreeOnlineUser: Done. `target_host.parse::<IpAddr>()` then dispatch on `V4` / `V6` / `Err`. Mirrors curl's `inet_pton` flow and matches the response parser above which reads all three ATYPs. Mock server updated to dispatch on the ATYP byte; new tests `handshake_ipv4_literal_uses_atyp_one` and `handshake_ipv6_literal_uses_atyp_four` cover both literal paths. Existing domain tests now assert `ATYP=0x03` so we'd catch regression.

FreeOnlineUser:
Round 3 addresses the May 21 inline asks. Force-pushed `69b50bc → 3f2012c`.

- L271 (`set_credentials` doc framing): dropped the RFC 1929 wording from the opener; length-limit guidance now branches by proxy kind (SOCKS5 = RFC 1929 1-255 / 0-255; HTTP Basic = no protocol limit).
- L288 (`&str` → `String`): `set_credentials` and `new_socks5_with_credentials` now take `String`.
- L301 (empty-user-some-password): rejected, matching RFC 1929 ULEN >= 1. Added an explicit named assertion next to the existing validation test.
- L335 (IP ATYP encoding): `target_host.parse::<IpAddr>()` then dispatch on V4 / V6 / Err. New mock-server-driven tests for IPv4 and IPv6 literals; existing domain tests now assert `ATYP=0x03` explicitly.

Open follow-up (your May 18 ask, not done here): reusing the `Response` type for HTTP CONNECT header parsing. Happy to do as a separate PR after this lands, or fold in if you'd prefer.

26 unit tests + 4 doc-tests passing locally.

bitreq/src/proxy.rs
```diff
@@ -97,24 +399,97 @@ impl Proxy {
...
+        socks5_handshake_body!(self, stream, target_host, target_port)
+    }
 
-        match status_code {
-            "200" => Ok(()),
-            "401" | "407" => Err(Error::InvalidProxyCreds),
-            _ => Err(Error::BadProxy),
+    /// Perform a SOCKS5 handshake on a connected async TCP stream.
+    #[cfg(feature = "async")]
+    pub(crate) async fn socks5_handshake_async(
```
TheBlueMatt: I don't believe this or the sync variant of it need to be `pub(crate)`.

bitreq/src/proxy.rs
```diff
@@ -86,6 +198,196 @@ impl Proxy {
...
+        buf[1] = 0x01; // CMD = CONNECT
+        buf[2] = 0x00; // RSV
+
+        let body_end = match target_host.parse::<IpAddr>() {
+            Ok(IpAddr::V4(v4)) => {
+                buf[3] = 0x01;
+                buf[4..8].copy_from_slice(&v4.octets());
+                8
+            }
+            Ok(IpAddr::V6(v6)) => {
```
TheBlueMatt, line 374: Claude pointed out that this might not actually work since `target_host` may be `[]`-encoded (e.g. `[::1]`) which won't parse as a v6.

TheBlueMatt (commented):
Sorry for the delay here. One more comment claude pulled up, plus `lib.rs` still references HTTP proxies in the feature context. Also needs a rebase to fix CI anyway, otherwise let's def land this in the next few weeks.

FreeOnlineUser:
Thanks Matt, no worries on the delay. We're even, I've been sidetracked on another project myself. Addressed all four:

1. Bracketed IPv6 (`[::1]`): good catch. A bracketed literal failed the `IpAddr` parse and fell through to the domain path (ATYP 0x03) instead of ATYP 0x04. Now strips one surrounding bracket pair before parsing; the domain arm still uses the original host, so real hostnames (never bracketed) are unaffected. Added `handshake_bracketed_ipv6_literal_uses_atyp_four`.

2. `pub(crate)` on `socks5_handshake_sync` / `socks5_handshake_async`: dropped to private. They're only reached through the `handshake_sync`/`handshake_async` dispatchers and the tests, so nothing outside the module needs them.

3. `lib.rs` HTTP-proxy wording: the `proxy` feature blurb now reads "proxy support (HTTP CONNECT and SOCKS5)", and I fixed a second stale "Only HTTP CONNECT proxies are supported" line in the Proxy example.

4. Rebased onto current master (the earlier CI red was base drift, not the code), and added a CHANGELOG entry under Unreleased.

Nothing else touched.

TheBlueMatt (approved):

TheBlueMatt (approved):
ACK cce43d5f288bb021603be2c4520fcb9e61105be0

tcharding:
Merged by way of the forgejo instance: https://git.rust-bitcoin.org/rust-bitcoin/corepc/pulls/533

PR #534, opened 2026-03-20
Update GetBlockStats fields to be optional

Abeeujah:
This PR aims to update the `GetBlockStats` model/response to match the core's spec for [Docs fix](https://bitcoincore.org/en/doc/23.0.0/rpc/blockchain/getblockstats/) the rpc spec for getblockstats specifies that it takes in two arguments 

1. hash_or_height    (string or numeric, required) The block hash or height of the target block
2. stats             (json array, optional, default=all values)

If the stats arg is provided, only the value for the stats is returned, if not provided it defaults to all, and all the fields values are returned; indicating that all the fields for `GetBlockStats` is optional. 

To better model and represent this, all the fields in v17, v25, and `model::GetBlockStats` are now wrapped in `Option`.

- The docs has been updated to include the optional stats argument
- Tests has also been updated to accept the stats argument, and confirm it behaves as specified for all versions of core.

### DIFF
--- a/client/src/client_sync/v17/blockchain.rs
+++ b/client/src/client_sync/v17/blockchain.rs
@@ -112,12 +112,20 @@ macro_rules! impl_client_v17__get_block_header {
 macro_rules! impl_client_v17__get_block_stats {
     () => {
         impl Client {
-            pub fn get_block_stats_by_height(&self, height: u32) -> Result<GetBlockStats> {
-                self.call("getblockstats", &[into_json(height)?])
+            pub fn get_block_stats_by_height(
+                &self,
+                height: u32,
+                stats: Option<&[&str]>,
+            ) -> Result<GetBlockStats> {
+                self.call("getblockstats", &[into_json(height)?, into_json(stats)?])
             }
 
-            pub fn get_block_stats_by_block_hash(&self, hash: &BlockHash) -> Result<GetBlockStats> {
-                self.call("getblockstats", &[into_json(hash)?])
+            pub fn get_block_stats_by_block_hash(
+                &self,
+                hash: &BlockHash,
+                stats: Option<&[&str]>,
+            ) -> Result<GetBlockStats> {
+                self.call("getblockstats", &[into_json(hash)?, into_json(stats)?])
             }
         }
     };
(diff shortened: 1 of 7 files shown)


jamillambert (commented):
The v25 `GetBlockStats` struct needs to be changed so that the fields are `Option<T>` like you did for v23.

jamillambert (commented):
Most of it is looking good. A few comments:

Could you expand the test to call `getblockstats` a second time with only one stat in the `stats` argument. The client macro will need to be updated to accept this option arg. Then check that the returned type for v23 onwards has that stat and the other fields are None. The Core docs are not clear what happens in v17, it still takes the optional `stats` argument but does not list any of the fields as optional. If the fields are not returned in v17-22 there will be an error running the test and essentially the v17 `GetBlockStats` will need to be replaced by the v23 one you have done. And then clearly documented that if the `stats` argument is used only those stats are returned. i.e. the change in Core v23 might just be a docs fix.

jamillambert (commented):
Looking good. Just needs a comment change and the commit message updated to reflect the new changes that affect all versions.

jamillambert:
> Sorry the review is taking this long/much roundtrips

It's fine, it's not easy to know exactly how things are done in a specific repo as a new contributor, and `corepc` has a few peculiarities.

And one more roundtrip...

The test is now looking a bit unwieldy. There is a large section of repeated code. In fact the two tests are identical, or at least could be. The only required difference is how `node` is created. Can you move the `let node = Node::with_wallet(...);` into `blockchain__get_block_stats__modelled()` and then for both cases call a single helper function, which is exactly what is in `getblockstats_txindex` right now but pass in the `&node`.

jamillambert (dismissed):
ACK fcd37d3da2b2d1e36faad1676f6611a53013a18c

Thanks for sticking with it and addressing all my comments.

tcharding (dismissed):
ACK fcd37d3da2b2d1e36faad1676f6611a53013a18c

jamillambert (approved):
ACK 94f6a4f6020f7d932d7310ee957bf0e239b0ff5a

The new tests are nice.

tcharding (approved):
ACK 94f6a4f6020f7d932d7310ee957bf0e239b0ff5a

tcharding:
Thanks for the contribution man.

PR #537, opened 2026-03-30
Test loadtxoutset RPC method

### LINKED ISSUE #313: How to test `loadtxoutset`?
I could not get the test for `loadtxoutset` to work. Using `dumptxoutset` and then trying to load it does not work, the hash of the snapshot is not accepted. It looks like it has to be registered with the node before it is created.

This RPC has been left `UNTESTED`.

Abeeujah:
Testing loadtxoutset on regtest requires a deterministic chain that produces a state matching the hardcoded assumeutxo entry in Bitcoin Core’s [chainparams](https://github.com/bitcoin/bitcoin/blob/2b6af628b140dc9d415733e5dc1577f3e54819af/src/kernel/chainparams.cpp#L622-L626) . Without this alignment, a standard dump-and-load cycle on a random regtest chain will always be rejected.

Replicate Bitcoin Core's C++ TestChain100Setup deterministic chain (mocktime 1598887952, P2PK coinbase to compressed pubkey of private key 0x01, 110 blocks) so the resulting block hash matches the assumeutxo entry. Dump the UTXO set from the miner node, feed headers to a fresh node via submitheader, then load the snapshot and verify the model.

Wire the load_tx_out_set client macro into v27 through v30, which were missing the call.

closes #313

### DIFF
--- a/integration_test/tests/blockchain.rs
+++ b/integration_test/tests/blockchain.rs
@@ -35,6 +35,102 @@ fn blockchain__dump_tx_out_set__modelled() {
 }
 
 #[test]
+#[cfg(not(feature = "v25_and_below"))]
+fn blockchain__load_tx_out_set__modelled() {
+    // Regtest `loadtxoutset` requires replicating Bitcoin Core's C++
+    // TestChain100Setup to produce the exact chain whose height at `110`
+    // assumeutxo snapshot hash is hardcoded in chainparams.
+    //
+    // The C++ test framework uses:
+    //   - SetMockTime(1598887952), incrementing by 1 after each block
+    //   - P2PK coinbase output to compressed pubkey of private key 0x01
+    //   - 110 coinbase-only blocks (no wallet transactions)
+    //
+    // Matching these exactly helps us arrive at the expected hash at snapshot
+
+    #[cfg(feature = "v29_and_below")]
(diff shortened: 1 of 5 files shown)


integration_test/tests/blockchain.rs
```diff
@@ -34,6 +34,101 @@ fn blockchain__dump_tx_out_set__modelled() {
...
+            .expect("getblockheader")
+            .block_header()
+            .expect("parse block header");
+        node_b.client.submit_header(&header).expect("submitheader");
+    }
+
+    let loaded: LoadTxOutSet =
+        node_b.client.load_tx_out_set(dump_path).expect("loadtxoutset should succeed");
+
+    let model = loaded.into_model().expect("into_model");
```
tcharding:
```rust
    let json: LoadTxOutSet =
        node_b.client.load_tx_out_set(dump_path).expect("loadtxoutset should succeed");

    let model: Result<mtype::LoadTxOutSet, LoadTxOutSetError> = json.into_model();
    let model = model.unwrap();
```

tcharding (commented):
This is sick bro, top effort!

Reviewed: c33cab745f19da5c44568f4d37d4b6249754da07

tcharding (approved):
ACK 913d5265871b5315c0c9de4c9906e9df8b0b9302

tcharding:
This was nice work bro, keep it coming!

PR #539, opened 2026-04-02
Test walletdisplayaddress

Abeeujah:
`signer__wallet_display_address` tests `walletdisplayaddress` RPC call mocking an external signer.

The test mocks a Hardware Wallet (HWW) interaction by:
1. Creating a temporary shell script that acts as an external signer.
2. Handling the 'enumerate', 'getdescriptors', and 'displayaddress' commands expected by Bitcoin Core.
3. Spawning a node with the `-signer` argument pointing to this mock.
4. Creating a descriptor-based wallet with `external_signer` enabled.
5. Asserting that calling `wallet_display_address` returns the expected address from the mock signer.

Heavily Inspired by [wallet_signer.py](https://github.com/bitcoin/bitcoin/blob/master/test/functional/wallet_signer.py)

### DIFF
--- a/client/src/client_sync/v22/mod.rs
+++ b/client/src/client_sync/v22/mod.rs
@@ -144,6 +144,7 @@ crate::impl_client_v17__add_multisig_address!();
 crate::impl_client_v17__backup_wallet!();
 crate::impl_client_v17__bump_fee!();
 crate::impl_client_v21__create_wallet!();
+crate::impl_client_v22__create_wallet!();
 crate::impl_client_v17__dump_priv_key!();
 crate::impl_client_v17__dump_wallet!();
 crate::impl_client_v17__encrypt_wallet!();
--- a/client/src/client_sync/v22/wallet.rs
+++ b/client/src/client_sync/v22/wallet.rs
@@ -44,3 +44,50 @@ macro_rules! impl_client_v22__wallet_display_address {
         }
     };
 }
+
+/// Implements Bitcoin Core JSON-RPC API method `createwallet`.
+#[macro_export]
+macro_rules! impl_client_v22__create_wallet {
+    () => {
+        impl Client {
+            /// Creates a wallet with external_signer=true.
+            ///
+            /// > createwallet "wallet_name" ( disable_private_keys blank "passphrase" avoid_reuse descriptors load_on_startup external_signer )
+            /// >
+            /// > Creates and loads a new wallet.
+            /// >
+            /// > Arguments:
+            /// > 1. wallet_name             (string, required) The name for the new wallet. If this is a path, the wallet will be created at the path location.
+            /// > 2. disable_private_keys    (boolean, optional, default=false) Disable the possibility of private keys (only watchonlys are possible in this mode).
+            /// > 3. blank                   (boolean, optional, default=false) Create a blank wallet. A blank wallet has no keys or HD seed. One can be set using sethdseed.
+            /// > 4. passphrase              (string, optional) Encrypt the wallet with this passphrase.
+            /// > 5. avoid_reuse             (boolean, optional, default=false) Keep track of coin reuse, and treat dirty and clean coins differently with privacy considerations in mind.
+            /// > 6. descriptors             (boolean, optional, default=true) Create a native descriptor wallet. The wallet will use descriptors internally to handle address creation
+            /// > 7. load_on_startup         (boolean, optional) Save wallet name to persistent settings and load on startup. True to add wallet to startup list, false to remove, null to leave unchanged.
+            /// > 8. external_signer         (boolean, optional, default=false) Use an external signer such as a hardware wallet. Requires -signer to be configured. Wallet creation will fail if keys cannot be fetched. Requires disable_private_keys and descriptors set to true.
+            pub fn create_wallet_external_signer(&self, wallet: &str) -> Result<CreateWallet> {
+                let disable_private_keys = true;
+                let blank = false;
+                let passphrase = String::new();
+                let avoid_reuse = false;
+                let descriptors = true;
+                let load_on_startup = false;
+                let external_signer = true;
+
+                self.call(
+                    "createwallet",
+                    &[
+                        wallet.into(),
+                        disable_private_keys.into(),
+                        blank.into(),
+                        passphrase.into(),
+                        avoid_reuse.into(),
+                        descriptors.into(),
+                        load_on_startup.into(),
+                        external_signer.into(),
+                    ],
+                )
+            }
+        }
+    };
+}
--- a/integration_test/tests/signer.rs
+++ b/integration_test/tests/signer.rs
@@ -5,6 +5,8 @@
 #![allow(non_snake_case)] // Test names intentionally use double underscore.
 #![allow(unused_imports)] // Because of feature gated tests.
 
+use bitcoin::address::{NetworkUnchecked, ParseError};
+use bitcoin::Address;
 use integration_test::{Node, NodeExt as _, Wallet};
 use node::vtype::*;
 use node::{mtype, Input, Output}; // All the version specific types.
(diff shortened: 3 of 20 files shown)


integration_test/tests/signer.rs
```diff
@@ -33,3 +33,87 @@ fn signer__enumerate_signers() {
...
+                true.into(),                   // external_signer
+            ],
+        )
+        .expect("createwallet with external signer");
+
+    let address = "bcrt1qm90ugl4d48jv8n6e5t9ln6t9zlpm5th68x4f8g";
+    let json: WalletDisplayAddress =
+        node.client.wallet_display_address(address).expect("walletdisplayaddress");
+    assert_eq!(json.address, address);
+}
```
satsfy, line 86: NIT: the multi-line inline script in Rust is awkward to read. Please move the mock signer to a fixture file (e.g. `integration_test/tests/fixtures/mock_signer.sh`), and load/copy it from the test. Then I’ll re-test and ACK.

satsfy (changes requested):
Nice test coverage, and I tested this locally on Linux (`bitcoind v30.2.0`) with: `cargo test --manifest-path integration_test/Cargo.toml --features latest --test signer -- --nocapture`

Concept ACK on c53f47d60553ceaf7566bce7cde3e620dce57969

integration_test/tests/signer.rs
Abeeujah: I get your point, It's ~11 lines now, does  just what it should do, test`walletdisplayaddress`, it should be easier on the eyes now?

satsfy: Thanks for addressing, seem fair.

satsfy (approved):
tACK d599574cb3aa70d6c5b289fce6f7ff382ab2b127

Tested on Linux (bitcoind v30.2.0) with `cargo test --manifest-path integration_test/Cargo.toml --features latest --test signer -- --nocapture`

satsfy (commented):
I suppose you might also want to update the `mod.rs` documentation files where it says `//! | walletdisplayaddress | version + model | UNTESTED` like in `types/src/v30/mod.rs`, because now it is tested.

integration_test/tests/signer.rs
```diff
@@ -33,3 +33,59 @@ fn signer__enumerate_signers() {
...
+                true.into(),                   // disable_private_keys
+                false.into(),                  // blank
+                "".into(),                     // passphrase
+                false.into(),                  // avoid_reuse
+                true.into(),                   // descriptors
+                node::serde_json::Value::Null, // load_on_startup
+                true.into(),                   // external_signer
+            ],
+        )
+        .expect("createwallet with external signer");
```
tcharding: I rekon it would be better if we created a function `create_wallet_external_signer` in a macro `impl_client_v22__create_wallet` like we do for `create_descriptor_wallet`.

integration_test/tests/signer.rs
```diff
@@ -33,3 +33,59 @@ fn signer__enumerate_signers() {
...
+
+    let json: WalletDisplayAddress =
+        node.client.wallet_display_address(address).expect("walletdisplayaddress");
```
tcharding, line 80: Can you call `into_model` and explicitly return the error type as well please. 

```rust
    let json: LoadTxOutSet =
        node_b.client.load_tx_out_set(dump_path).expect("loadtxoutset should succeed");
    let model: Result<mtype::LoadTxOutSet, LoadTxOutSetError> = json.into_model();
    let model = model.unwrap();
```

client/src/client_sync/v22/mod.rs
```diff
@@ -143,6 +143,7 @@ crate::impl_client_v17__abort_rescan!();
 crate::impl_client_v17__add_multisig_address!();
 crate::impl_client_v17__backup_wallet!();
 crate::impl_client_v17__bump_fee!();
+crate::impl_client_v22__create_wallet!();
```
jamillambert: Nit. order by version. i.e. v21 before v22.

client/src/client_sync/v22/wallet.rs
```diff
@@ -44,3 +44,42 @@ macro_rules! impl_client_v22__wallet_display_address {
...
 }
+
+/// Implements Bitcoin Core JSON-RPC API method `createwallet`.
+#[macro_export]
+macro_rules! impl_client_v22__create_wallet {
+    () => {
+        impl Client {
+            /// Creates a wallet with external_signer=true.
+            ///
+            /// > createwallet "wallet_name" ( disable_private_keys blank "passphrase" avoid_reuse descriptors load_on_startup )
```
jamillambert:
```suggestion
            /// > createwallet "wallet_name" ( disable_private_keys blank "passphrase" avoid_reuse descriptors load_on_startup external_signer )
```

integration_test/tests/signer.rs
```diff
@@ -33,3 +35,52 @@ fn signer__enumerate_signers() {
...
     assert_eq!(first_tx.fingerprint, "deadbeef");
 }
+
+#[test]
+#[cfg(unix)]
+#[cfg(not(feature = "v21_and_below"))]
+fn signer__wallet_display_address() {
+    use std::os::unix::fs::PermissionsExt;
+
+    let address = "bcrt1qm90ugl4d48jv8n6e5t9ln6t9zlpm5th68x4f8g";
```
jamillambert, line 46:
```suggestion
    // Mock address and xpub from Bitcoin Core's `wallet_signer.py`.
    let address = "bcrt1qm90ugl4d48jv8n6e5t9ln6t9zlpm5th68x4f8g";
```

Or some other comment to where they came from.

integration_test/tests/signer.rs
```diff
@@ -33,3 +35,52 @@ fn signer__enumerate_signers() {
...
+    let node = Node::with_wallet(Wallet::None, &[&signer_arg]);
+
+    let _: CreateWallet = node
+        .client
+        .create_wallet_external_signer("hww")
+        .expect("createwallet with external signer");
+
+    let json: WalletDisplayAddress =
+        node.client.wallet_display_address(address).expect("walletdisplayaddress");
+    assert_eq!(json.address, address);
```
jamillambert:
```suggestion
```
This is redundant, the modelled address is checked below.

jamillambert (commented):
The test looks good. Just a few minor comments.

client/src/client_sync/v22/mod.rs
Abeeujah: I appreciate the nit :clap:

integration_test/tests/signer.rs
Abeeujah: true, has been dropped :+1:

client/src/client_sync/v22/mod.rs
```diff
@@ -144,6 +144,7 @@ crate::impl_client_v17__add_multisig_address!();
 crate::impl_client_v17__backup_wallet!();
 crate::impl_client_v17__bump_fee!();
 crate::impl_client_v21__create_wallet!();
+crate::impl_client_v22__create_wallet!();
```
tcharding, line 147: This is surprising, I'd expect there to be just a single macro that implements `createwallet`.

tcharding:
hmm the `create_wallet` macros are a bit of a mess. We have `create_wallet_descriptor` and this introduces a bit more mess.

Perhaps we should have a patch at the front of this PR that removes the `_descriptor` macro and just implements `impl_client_vXY__create_wallet` for each version that modifies it and in each just duplicate all the code from the previous one?

@jamillambert do you have an opinion on this?

We could also just merge this as is and fix as a follow up.

tcharding (dismissed):
ACK 4c9c96076fb8f46197b013bd0446fb0973cdae97

tcharding:
~Everything else looks good to me.~ Oh woops, the test name is wrong, needs `__modelled` on it.

client/src/client_sync/v22/mod.rs
Abeeujah: In [v21](https://bitcoincore.org/en/doc/0.21.0/rpc/wallet/createwallet/) `createwallet` was taking only 7 arguments, `external_signer` wasn't among them, until [v22](https://bitcoincore.org/en/doc/22.0.0/rpc/wallet/createwallet/) where the arguments became 8 and `external_signer` was the addition; that's why I couldn't do the `create_wallet_external_signer` in `impl_client_v21__create_wallet!`

jamillambert (approved):
ACK df260397396b9ca084aaa33d02de13dbada115c0

jamillambert:
> hmm the `create_wallet` macros are a bit of a mess.

Yeah, it's the only macro that has multiple version definitions at the same time. I wasn't sure if duplicating all the code in the v21 macro in v22 and so on is better or not. What is done here is functionally fine and can be cleaned up in a follow up.

tcharding (approved):
ACK df260397396b9ca084aaa33d02de13dbada115c0

tcharding:
> can be cleaned up in a follow up.

Works for me.

PR #540, opened 2026-04-07
ci: migrate GitHub Actions to cargo rbmt

### LINKED ISSUE #538: Move CI to `cargo-rbmt`

satsfy:
Fixes #538

This PR migrates the CI workflow from the maintainer-tools shell entrypoints to cargo rbmt, joins Stable/Nightly/MSRV test jobs into a single toolchain/lockfile matrix and makes the supporting library changes needed for checks to pass cleanly.

Summary:
- gate async TLS code on the async feature in bitreq
- fix the `webpki-roots` cfg name in bitreq
- add workspace and per-crate rbmt metadata
- migrate `rust.yaml` to cargo rbmt
- rename `lint-integtation-tests.sh` to `lint-integration-tests.sh`


tcharding (dismissed):
ACK 1ddbb2bcce9c7982707b9e17d2943ce210dfe4cf

tcharding:
Thanks man, top effort! @nyonson if you get a second could you take a look at this please?

bitreq/src/url.rs
```diff
@@ -383,6 +383,7 @@ impl Url {
     pub fn as_str(&self) -> &str { &self.serialization }
 
     /// Returns `true` if the URL scheme is "https" or "wss".
+    #[allow(dead_code)]
```
tnull: I think we currently only use this when

```
        #[cfg(all(feature = "proxy", feature = "std"))]
```

Please apply the same cfg-gate here rather than allowing dead code.

bitreq/src/url.rs
```diff
@@ -401,6 +402,7 @@ impl Url {
     ///
     /// The returned string includes the leading `/` (if present) and the `?`
     /// separator (if there's a query string). Returns "/" if the path is empty.
+    #[allow(dead_code)]
```
tnull: Same here, this is only used when `std`,  so please `cfg`-gate on that rather than allowing dead code.

bitreq/src/connection/rustls_stream.rs
```diff
@@ -13,16 +13,26 @@ use std::sync::OnceLock;
...
+    feature = "native-tls",
+    not(feature = "rustls"),
+    feature = "tokio-native-tls"
+))]
 use tokio_native_tls::TlsConnector as AsyncTlsConnector;
-#[cfg(feature = "tokio-rustls")]
+#[cfg(all(feature = "async", feature = "tokio-rustls"))]
 use tokio_rustls::{client::TlsStream, TlsConnector};
-#[cfg(feature = "rustls-webpki")]
+#[cfg(feature = "webpki-roots")]
```
tnull: This seems like a behavior change, can you expand why this is necessary?

bitreq/src/connection/rustls_stream.rs
```diff
@@ -13,16 +13,26 @@ use std::sync::OnceLock;
...
+#[cfg(all(feature = "async", feature = "tokio-rustls"))]
 use super::{AsyncHttpStream, AsyncTcpStream};
-#[cfg(all(feature = "native-tls", not(feature = "rustls"), feature = "tokio-native-tls"))]
+#[cfg(all(
+    feature = "async",
+    feature = "native-tls",
+    not(feature = "rustls"),
+    feature = "tokio-native-tls"
+))]
 use super::{AsyncHttpStream, AsyncTcpStream};
```
tnull, line 26: I'm confused why we have the duplicate line to begin with. But rather than adding more feature-gates on top, please try to DRY this up.

clippy.toml
```diff
@@ -1,2 +1 @@
-msrv = "1.75.0"
```
tnull, line 1: Why are we dropping this? Shouldn't we still allow respecting MSRV when linting? Or will it just use whatever is set in `Cargo.toml`?

.github/workflows/rust.yaml
```diff
@@ -10,88 +10,40 @@ name: Continuous integration
...
-    name: Test - nightly toolchain
+  Test: # 6 jobs: 3 toolchains × 2 lock files.
+    name: Test - ${{ matrix.toolchain }} toolchain (${{ matrix.dep }})
     needs: Prepare
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
       matrix:
         dep: [minimal, recent]
+        toolchain: [stable, nightly, msrv]
```
tnull, line 19: Where is the `msrv` toolchain defined? While a global MSRV might work out right now, I'm not sure this will always be the case as `bitreq`'s TLS dependencies might for example bump MSRV at some point. Would the entire ecosystem / all crates then fine with bumping MSRV, even if they don't share the same dependencies?

clippy.toml
tcharding: Yep `cargo rbmt` gets this from the manifest.

.github/workflows/rust.yaml
tcharding: cc @nyonson, do you remember if `[workspace.metadata.rbmt.toolchains]` can be overridden with crate-specific config?

bitreq/src/connection/rustls_stream.rs
tcharding: Good reviewing tnull, I saw this and skimmed over it.

Cargo.toml
```diff
@@ -14,3 +14,32 @@ path = "node"
 
 [patch.crates-io.corepc-types]
 path = "types"
+
+[workspace.metadata.rbmt.toolchains]
+nightly = "nightly-2025-09-12"
+stable = "stable"
+msrv = "1.75.0"
```
luisschwab: - MSRV is sourced from Cargo.toml > rust-version
- `stable` should be an explicit version

```suggestion
[workspace.metadata.rbmt.toolchains]
nightly = "nightly-2025-09-12"
stable = "1.94.1"
```

luisschwab (changes requested):

.github/workflows/rust.yaml
nyonson: The `msrv` is implicitly picked up from the `rust-version` keys in the package manifests. As of today, `cargo-rbmt` has a limitation where it only works in workspaces with 1 msrv. This can be fixed if needed, just a bit of complexity which hasn't been necessary.

luisschwab: > cc @nyonson, do you remember if [workspace.metadata.rbmt.toolchains] can be overridden with crate-specific config?

I was planning to migrate Floresta to `cargo-rbmt` soon. This would be handy, since the lib crates probably have lower MSRV than the binary itself.

Cargo.toml
satsfy: Accepted, thanks. Updated the toolchain config.

.github/workflows/rust.yaml
nyonson, line 19: FWIW I am not sure how valuable testing on the nighlty toolchain is, but doesn't hurt either. Some recent discussion over here: https://github.com/rust-bitcoin/rust-psbt/pull/87#issuecomment-4151565331

satsfy: Thanks for the discussion. For this PR, this is effectively a no-op. I’m happy to follow up with per-crate msrv support once we need it, or once cargo-rbmt supports it.

.github/workflows/rust.yaml
```diff
@@ -10,88 +10,40 @@ name: Continuous integration
...
-
-  Stable:                       # 2 jobs, one per lock file.
-    name: Test - stable toolchain
     runs-on: ubuntu-latest
-    strategy:
-      fail-fast: false
-      matrix:
-        dep: [minimal, recent]
+    outputs:
+      rbmt_version: ${{ steps.read_rbmt_version.outputs.value }}
```
nyonson: It looks like this output is only used with the `setup-rbmt` action. Since the `rbmt-version` file is following the standard convention in name and location, the `setup-rbmt` action would use it by default without this `Prepare` job. But being explicit doesn't hurt.

.github/workflows/rust.yaml
```diff
@@ -106,28 +58,24 @@ jobs:
...
-        run: rustup component add clippy
-      - name: "Set dependencies"
-        run: cp Cargo-${{ matrix.dep }}.lock Cargo.lock
-      - name: "Run test script"
+          rbmt-version: ${{ needs.Prepare.outputs.rbmt_version }}
+      - name: "Run lints"
         run: |
-          ./maintainer-tools/ci/run_task.sh lint
-          ./contrib/lint-integtation-tests.sh
+          cargo rbmt --lock-file ${{ matrix.dep }} lint \
```
nyonson, line 50: Just curious for feedback on rbmt, does the `fuzz` package not play nice here?

nyonson (commented):
ACK: fff9fcf60c94d2fc2351eace4163c3bb3e0e1cbd

rbmt stuff looks good to me! I don't have context on the other stuff.

.github/workflows/rust.yaml
satsfy: This PR aims to not modify the previous CI job list, which included nightly. I suppose nightly is here as early warning for upcoming compiler or toolchain changes.

nyonson: Trackin it over here: https://git.rust-bitcoin.org/rust-bitcoin/rust-bitcoin-maintainer-tools/issues/119

luisschwab:
Just need to fix the lint CI error, otherwise LGTM

bitreq/src/url.rs
satsfy: Nice, I changed the config gate. But those methods are used in std builds even without proxy. If I remove the proxy it causes CI errors. So I replaced `#[allow(dead_code)]` with `#[cfg(feature = "std")]`.

bitreq/src/url.rs
satsfy: Replaced `#[allow(dead_code)]` with `#[cfg(feature = "std")]` here too.

bitreq/src/connection/rustls_stream.rs
satsfy: Apologies. After several rounds of revisions and CI runs, I overlooked this specific change. I’ve now fully reverted it and verified it is not necessary.

bitreq/src/connection/rustls_stream.rs
satsfy: Fully reverted it and verified it is not necessary.

.github/workflows/rust.yaml
satsfy: TYSM! I was able to remove 29 loc with no behavioral changes.

.github/workflows/rust.yaml
satsfy: package jsonrpc-fuzz is included and works with rbmt ok from what I've seen.

satsfy:
Addressed all review comments after the force-pushes and cut a fair amount of LOC.

tcharding (approved):
ACK 0e2d33611c6fc223f617751e4a77423451e55850

tcharding:
Waiting for #549 before merge.

luisschwab (approved):
ACK 0e2d33611c6fc223f617751e4a77423451e55850

tnull:
ACK 0e2d33611c6fc223f617751e4a77423451e55850

Issue #544, opened 2026-04-09
bug(corepc-types): type of prune_target_size field in model::GetBlockchainInfo should be i64 instead of u32


RazorBest:
You can configure a bitcoin-core node with `prune=5000` or greater (which is in MB). As a consequence, the RPC call `getblockchaininfo` will return `"prune_target_size": 5242880000` , or greater, which is the value in bytes of the prune limit that was set for the node. This value is outside of `u32`.

Demo:
```rust
use std::error::Error;

use corepc_client::client_sync::Auth;
use corepc_client::client_sync::v29::Client;

fn main() -> Result<(), Box<dyn Error>> {
    let rpc_host = "http://127.0.0.1:8332";
    let cookie_path = "/home/btc/.bitcoin/.cookie";
    let auth = Auth::CookieFile(cookie_path.into());
    let client = Client::new_with_auth(rpc_host, auth)?;

    let rpc_result = client.get_blockchain_info()?;
    println!("RPC response: {rpc_result:?}");

    let generic_result = rpc_result.into_model()?;
    println!("Generic response: {generic_result:?}");

    Ok(())
}
```

Cargo.toml:
```
[dependencies]
corepc-client = { version = "0.10.0", features = ["client-sync"] }
```

Configure a node with `prune=5000` or higher. Change `rpc_host` and `cookie_path` accordingly.

Output:
```
RPC response: GetBlockchainInfo ...
Error: Numeric(Overflow { field: "prune_target_size", value: 5242880000 })
```

Expected behaviour: no error.

---

`v17::GetBlockchainInfo` and `v29::GetBlockchainInfo` both define the field `prune_target_size` as `Option<i64>`. So, only `model::GetBlockchainInfo::prune_target_size` seems to be the outlier.

I don't know if it was a deliberate decision to represent all the version non-specific type fields integers as 32-bit ints. But unless proven wrongly, I'll just categorize this as a bug.

RazorBest:
I later discovered these comments: https://github.com/rust-bitcoin/corepc/blob/207e17d09c2b839f672ab877cb141e4e5636e357/types/src/lib.rs#L48-L77

Which show that the choice to convert all the i64 to u32 was deliberate. And any integer that exceeds u32 is against the expectations. However, it's not clear whether "the expectations" are an invariant of this library.

tcharding:
[...]

PR #547, opened 2026-04-10
Change the type of prune_target_size field in model::GetBlockchainInfo from u32 to u64

RazorBest:
Solves https://github.com/rust-bitcoin/corepc/issues/544

## Manual testing

**Before:**

Run `git checkout 26e3998a283091abeed33a44a8176ec62c15c4d8` (the commit that introduced the integration test). Then:
```
cd integration_test/
cargo test --release --features <bitcoind_version>
```

Example: `cargo test --release --features 28_1`.

This test should fail in the commit it was introduced.

**After:**

Checkout back to the branch and run the same test again. Now, the test should succeed.

### DIFF
--- a/integration_test/tests/blockchain.rs
+++ b/integration_test/tests/blockchain.rs
@@ -206,7 +206,7 @@ fn blockchain__get_block__modelled() {
 
 #[test]
 fn blockchain__get_blockchain_info__modelled() {
-    let node = Node::with_wallet(Wallet::None, &[]);
+    let node = Node::with_wallet(Wallet::None, &["-prune=10000"]);
 
     let json: GetBlockchainInfo = node.client.get_blockchain_info().expect("rpc");
     let model: Result<mtype::GetBlockchainInfo, GetBlockchainInfoError> = json.into_model();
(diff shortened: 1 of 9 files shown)


integration_test/tests/blockchain.rs
```diff
@@ -213,6 +213,16 @@ fn blockchain__get_blockchain_info__modelled() {
     model.unwrap();
 }
 
+/// The into_model works correctly for getblockchaininfo, for a `-prune=10000` node
+#[test]
+fn blockchain__pruned__get_blockchain_info__modelled() {
```
jamillambert: I would just replace the existing test with this one. It covers everything and then it doesn't break the naming convention with pruned.

Also can you move this commit to after the fix so that each commit in the series passes the tests.

Otherwise it all looks good.

RazorBest: Done

jamillambert: Sorry I meant keep the old test name without the "pruned" bit.

RazorBest: Then I'm also removing the comment because none of the other tests in the file have a comment that explains what the test does.

RazorBest: Done

jamillambert (approved):
ACK b10dc06c2a203f19f13440292e970ffd5679ebb4
Nice one.

tcharding (approved):
ACK b10dc06c2a203f19f13440292e970ffd5679ebb4

tcharding:
Waiting for #549 before merge.

PR #549, opened 2026-04-13
Prepare stack of releases

tcharding:
After releasing `corepc-types 0.11.0` there was a fix that had to be done to `jsonrpc`. I then forgot to release `client` and `node`.
    
Yesterday I released `client v0.11.0` but then when attempting the `node v0.11.0` release found it was broken because we did a bunch of API additions (and breaks) since `types` was released.
    
- Bump `types` to `v0.12.0`
- Bump `client` to  `v0.12.0`
- Bump `node` straight to `v0.12.0` skipping `v0.11.0` so it's version number stays in sync with the other two.

FTR there is no need to yank `client 0.11.0` because it is functional and harmless.

### DIFF
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -238,7 +238,7 @@ checksum = "773648b94d0e5d620f64f280777445740e61fe701025087ec8b57f45c791888b"
 
 [[package]]
 name = "corepc-client"
-version = "0.11.0"
+version = "0.12.0"
 dependencies = [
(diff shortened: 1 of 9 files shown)


jamillambert (approved):
ACK 038508dd6a9c5765a1b67a08f94514d724708f48

tcharding:
Tagged and published `types`, `client` and `node` - all `v0.12.0`.

PR #556, opened 2026-04-17
Bump `rustls` and related crates to latest versions

tnull:
Fixes #553.
Alternative to #536, which also bumps `rustls-webpki` and other related crates.

Since there have recently been some advisories, we bump `rustls` and related crates such as `rustls-webpki` to latest versions.

PR #536, opened 2026-03-28
Bitreq: Deps: bump `rustls` to version 0.23

hydra-yse:
This PR bumps `bitreq`'s rustls version to 0.23, mainly to allow dependents to ship with newer versions of rustls without pulling separate versions.

PR #557, opened 2026-04-17
CI: Add `cargo-audit` workflow

tnull:
Run security audits against all three checked-in lock files using the official rustsec/audit-check action. Triggers on push, PRs, and daily via cron to catch newly published advisories.

Co-Authored-By: HAL 9000

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,28 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+  schedule:
+    - cron: '0 0 * * *'
+
+name: Audit
+
+jobs:
+  audit:
+    name: Security audit - ${{ matrix.lockfile }}
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      checks: write
+    strategy:
+      fail-fast: false
+      matrix:
+        lockfile: [Cargo-minimal.lock, Cargo-recent.lock]
+    steps:
+      - uses: actions/checkout@v4
+      - name: Use ${{ matrix.lockfile }}
+        run: cp ${{ matrix.lockfile }} Cargo.lock
+      - uses: rustsec/audit-check@v2.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -18,10 +18,14 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        lockfile: [Cargo-minimal.lock, Cargo-recent.lock]
+        lockfile: [Cargo.lock, Cargo-minimal.lock, Cargo-recent.lock]
     steps:
       - uses: actions/checkout@v4
+      - name: Generate fresh Cargo.lock
+        if: ${{ matrix.lockfile == 'Cargo.lock' }}
+        run: cargo generate-lockfile
       - name: Use ${{ matrix.lockfile }}
+        if: ${{ matrix.lockfile != 'Cargo.lock' }}
         run: cp ${{ matrix.lockfile }} Cargo.lock
       - uses: rustsec/audit-check@v2.0.0
         with:
--- a/Cargo-minimal.lock
+++ b/Cargo-minimal.lock
@@ -192,9 +192,9 @@ checksum = "1fd0f2584146f6f2ef48085050886acf353beff7305ebd1ae69500e27c67f64b"
 
 [[package]]
 name = "bytes"
-version = "1.11.0"
+version = "1.11.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


tnull:
Also bumped some dependencies, rest should be done in #556 or #536.

luisschwab (approved):
ACK 2e56f00af80e5bc8a2c09c73f033f94ff923ec0d

luisschwab:
cc @satsfy

.github/workflows/audit.yml
```diff
@@ -0,0 +1,29 @@
...
+  audit:
+    name: Security audit - ${{ matrix.lockfile }}
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      checks: write
+    strategy:
+      fail-fast: false
+      matrix:
+        lockfile: [Cargo.lock, Cargo-minimal.lock, Cargo-recent.lock]
```
satsfy: Why is `Cargo.lock` part of the matrix?

tcharding: > against all three checked-in lock files

Yeah there should only be two. Did I commit one in `electrsd` accidentally or something?

tcharding:
As it stands no-one is ever going to look at the auction run (well I definitely won't remember to do it). Can we have the action raise an issue if there are warnings? Or is that going to be too noisy? I don't have a lot of experience with `cargo audit`. (We do this already in rust-bitcoin for mutation testing when zombies survive.)

luisschwab:
The GitHub bot will usually open an issue when there is a new RUSTSEC (see https://github.com/bitcoindevkit/bdk/issues/2178).

tcharding:
Does that bot hit projects to use `cargo audit`?

luisschwab:
Actually, I **think** we need to use the `actions-rust-lang/audit` action for it to open issues. On my personal repos I only use the bare `cargo audit` and it does not open any issues there.

This is `bdk`'s audit job:

```yml
name: Audit

on:
  push:
    paths:
      # Run if workflow changes
      - '.github/workflows/audit.yml'
      # Run on changed dependencies
      - '**/Cargo.toml'
      - '**/Cargo.lock'
      # Run if the configuration file changes
      - '**/audit.toml'
  # Rerun periodically to pick up new advisories
  schedule:
    - cron: '0 0 * * *' # Nightly
  # Run manually
  workflow_dispatch:

jobs:
  audit:
    runs-on: ubuntu-latest
    permissions:
      contents: read
      issues: write
    steps:
      - uses: actions/checkout@v6
        with:
          persist-credentials: false
      - uses: actions-rust-lang/audit@v1
        name: Audit Rust Dependencies
```

I could be talking out of my ass, I'm not certain about this.

tcharding:
Sounds legit :)

tnull:
> As it stands no-one is ever going to look at the auction run (well I definitely won't remember to do it). Can we have the action raise an issue if there are warnings? Or is that going to be too noisy? I don't have a lot of experience with `cargo audit`. (We do this already in rust-bitcoin for mutation testing when zombies survive.)

This will already do this, hence the `issues: write` permissions. See for example LDK Node where we use it (https://github.com/lightningdevkit/ldk-node/blob/main/.github/workflows/audit.yml), example issue: https://github.com/lightningdevkit/ldk-node/issues/836.


> Actually, I **think** we need to use the `actions-rust-lang/audit` action for it to open issues. On my personal repos I only use the bare `cargo audit` and it does not open any issues there.

No, `audit-check` is definitely sufficient, but it will only open issues via the `cron`-scheduled jobs: 

"The action does not raise issues when it is not triggered from a "cron" scheduled workflow." (see https://github.com/rustsec/audit-check#granular-permissions)

If we want to validate that it works as expected, we could split out the fixes / bump commits to a second PR and first ensure the job opens an issue for the advisories found. Should we do that?

.github/workflows/audit.yml
tnull: Ah, right, dropped it.

tnull: Actually, it might make sense to add a step to generate a clean-slate `Cargo.lock` for this and then have the audit job check it? Otherwise we could theoretically miss advisories that don't pertain to the specific versions checked-in via `Cargo-{minimal, recent}.lock`, but would still be present if users would just `cargo build` regularly?

tnull: Now added a `generate-lockfile` step in https://github.com/rust-bitcoin/corepc/pull/557/changes/52e3dad9d6e671109614d45d19a07dcdb27ec716, so we don't *just* check our checked-in versions over and over again.

luisschwab:
@tcharding can you re-trigger the failing job?

luisschwab (approved):
ACK 52e3dad9d6e671109614d45d19a07dcdb27ec716

jamillambert (approved):
ACK 52e3dad9d6e671109614d45d19a07dcdb27ec716

tcharding:
> Should we do that?

If you say it works I'm happy to merge and run with it. Super easy to change later if needed.

PR #563, opened 2026-04-21
bitreq: Fix tokio-rustls feature gate

jamillambert:
The code was gated on the dependency instead of the feature alias which
enables other required dependencies. This caused a build failure when
both `std` and `tokio-rustls` features were enabled and not the other
required dependencies.

Change the feature gate to the feature alias instead of the dependency.

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -158,13 +158,13 @@ impl Write for HttpStream {
     }
 }
 
-#[cfg(feature = "tokio-rustls")]
+#[cfg(any(feature = "async-https-rustls", feature = "async-https-rustls-probe"))]
 type AsyncSecuredStream = rustls_stream::AsyncSecuredStream;
 
 #[cfg(feature = "async")]
 pub(crate) enum AsyncHttpStream {
     Unsecured(AsyncTcpStream),
-    #[cfg(feature = "tokio-rustls")]
+    #[cfg(any(feature = "async-https-rustls", feature = "async-https-rustls-probe"))]
     Secured(Box<AsyncSecuredStream>),
 }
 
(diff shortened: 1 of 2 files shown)


tnull (commented):
Thanks, makes sense I think. But if we only now discovered this, it seems we might be lacking test coverage in CI that would have found this? Could we add the missing pieces in this PR?

jamillambert:
> Thanks, makes sense I think. But if we only now discovered this, it seems we might be lacking test coverage in CI that would have found this? Could we add the missing pieces in this PR?

I had a further look into this and it only fails to build with the exact feature set of `std` and `tokio-rustls`. There is no practical way to check every combination of features in CI, and I think the current method of checking a sub set is  a good compromise. It did still catch the issue, just not in every run.

I also changed the fix to gate on the features `async-https-rustls` and `async-https-rustls-probe`, instead of adding `async`.

jamillambert:
`Audit / Security audit` failure is unrelated to this PR and fixed in #565

tnull (approved):
ACK f4f1c19c0e05028d929d2dd872c7699c9edf8583

luisschwab (approved):
ACK f4f1c19c0e05028d929d2dd872c7699c9edf8583

ton-anywhere:
ACK https://github.com/rust-bitcoin/corepc/commit/f4f1c19c0e05028d929d2dd872c7699c9edf8583

tcharding:
FTR I don't know what I"m supposed to do about red CI jobs for `cargo audit`? I just merged right past them.

tnull:
> FTR I don't know what I"m supposed to do about red CI jobs for `cargo audit`? I just merged right past them.

Yeah, I think that fine on a per-PR basis. The CI checks are mostly there to ensure that no new known vulnerabilities are introduced. But if it annoys you and it's happening too often (in my experience it's rare) maybe we just reduce it to the cronjob?

PR #564, opened 2026-04-22
`bitreq`: Evict dead entries from async connection pool

### LINKED ISSUE #562: bitreq: `Client` connection pool retains dead connections indefinitely on I/O error
`bitreq::Client` caches `Arc<AsyncConnection>` by host+port in a `HashMap` and
never evicts an entry, never revalidates before reuse, and never replaces an
entry on error. Once a pooled socket is dead (e.g. closed by the peer after an
idle timeout, or reset), the pool hands out the same dead `Arc` to every
subsequent caller for that key, permanently — until the process restarts.

Because `send_async` returns the underlying error to the caller without
touching the pool, retry logic in downstream crates is defeated: each retry
calls `send_async` again, gets the same `Arc<AsyncConnection>` back, and fails
on the same dead socket.

We hit this in a bitcoin RPC client (`bitcoind-async-client` 0.10.x, which now
uses bitreq) talking to a local `bitcoind`. After ~30 seconds of idle, bitcoind
closes the RPC socket per its `rpcservertimeout`. The next request fails, and
then all subsequent requests fail indefinitely — even though a fresh TCP
connection to the same port succeeds instantly. Only restarting the client
process recovers.

- `bitreq` 0.3.4 (latest as of writing)
- observed via `bitcoind-async-client` 0.10.2 against `bitcoind` 29.x on Linux

tnull:
Fixes #562.

Validate pool entries on acquire and clean up after every send:

- Remove entries whose keep-alive deadline has passed or whose inner
  state has been poisoned, rather than blindly cloning the `Arc`.
- Evict on send failure, `Connection: close`, or malformed
  `Keep-Alive`.
- Refresh LRU position on a cache hit.

Pre-send insertion of a fresh `Arc` on a miss is kept so concurrent
callers arriving during the first round-trip can share the socket for
pipelining; `reusable_until` and the post-send eviction together
ensure they never observe a known-dead `Arc`.

### DIFF
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -8,6 +8,7 @@
 
 use std::collections::{hash_map, HashMap, VecDeque};
 use std::sync::{Arc, Mutex};
+use std::time::Instant;
 
 use crate::connection::AsyncConnection;
 use crate::request::{OwnedConnectionParams as ConnectionKey, ParsedRequest};
@@ -16,7 +17,10 @@ use crate::{Error, Request, Response};
 /// A client that caches connections for reuse.
 ///
 /// The client maintains a pool of up to `capacity` connections, evicting
-/// the least recently used connection when the cache is full.
+/// the least recently used connection when the cache is full. Pooled
+/// connections are validated on every acquire: an entry whose keep-alive
+/// deadline has passed — or whose underlying socket has been poisoned by
+/// a previous failure — is dropped and a fresh connection is opened.
 ///
 /// # Example
 ///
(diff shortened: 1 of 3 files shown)


bitreq/src/client.rs
```diff
@@ -62,39 +66,90 @@ impl Client {
...
+            }
+        }
+    }
 
-        // Send the request
-        conn.send(parsed_request).await
+    /// Removes any pool entry for `key`. No-op if the slot is already empty.
+    fn evict(&self, key: &ConnectionKey) {
+        let mut state = self.r#async.lock().unwrap();
+        state.connections.remove(key);
```
storopoli, line 149: I think this should evict by identity, not just by `ConnectionKey`.

There is a race where an older poisoned `Arc` can finish later and remove a newer healthy entry for the same key:

1. tasks A and B are both using the old pooled connection for `K`
2. A fails and marks the old connection dead
3. task C opens and inserts a fresh connection for `K`
4. B reaches this cleanup path later and `remove(key)` drops C's fresh entry

That doesn't break in-flight requests, but it does create avoidable reconnect churn under load. I think `evict` needs to compare the currently pooled `Arc` against the one that just finished and only remove on `Arc::ptr_eq`.

Permalinks:
- `send_async` cleanup: https://github.com/tnull/corepc/blob/708d9c388aa148eb3f11a2b3e5d72d361ab116bf/bitreq/src/client.rs#L87-L95
- `evict`: https://github.com/tnull/corepc/blob/708d9c388aa148eb3f11a2b3e5d72d361ab116bf/bitreq/src/client.rs#L146-L153

storopoli (changes requested):

TheBlueMatt (commented):
I'm not sure I'm understanding the fix here - if we hit a dead connection when we go to request we should fix the Connection, not build a new one in the Client?

tnull:
Spent some time trying to reproduce #562 and couldn't really. This PR is now also somewhat superseded by https://github.com/rust-bitcoin/corepc/pull/661 (w.r.t. `keep-alive` handling), so going ahead closing this.

PR #569, opened 2026-04-23
bitreq: TLS modules refactor

ton-anywhere:
Similar to [this pull request](https://github.com/rust-bitcoin/corepc/pull/563), I encountered an error using the  `std` and `rustls` features together.

This PR makes TLS backend selection explicit by using cfg gates with cargo features instead of relying on implicit dependency features.

While working on mixed TLS feature combinations, I ran into cases where backend selection became unclear so I had to decouple sync and async modes.
 
 This PR:
  - splits TLS handling into `rustls_stream` and `native_tls_stream`
  - selects sync and async backends independently
  - gives `rustls` precedence over `native-tls` when both backends are enabled for the same sync or async mode

### Error Example

image

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -29,14 +29,14 @@ use crate::{Error, Method, ResponseLazy};
 
 type UnsecuredStream = TcpStream;
 
-#[cfg(feature = "rustls")]
+#[cfg(any(feature = "rustls", feature = "https-native-tls"))]
 mod rustls_stream;
-#[cfg(feature = "rustls")]
+#[cfg(any(feature = "rustls", feature = "https-native-tls"))]
 type SecuredStream = rustls_stream::SecuredStream;
 
 pub(crate) enum HttpStream {
     Unsecured(UnsecuredStream, Option<Instant>),
-    #[cfg(feature = "rustls")]
+    #[cfg(any(feature = "rustls", feature = "https-native-tls"))]
     Secured(Box<SecuredStream>, Option<Instant>),
     #[cfg(feature = "async")]
     Buffer(std::io::Cursor<Vec<u8>>),
@@ -81,7 +81,7 @@ impl Read for HttpStream {
                 timeout(inner, *timeout_at)?;
                 inner.read(buf)
             }
-            #[cfg(feature = "rustls")]
+            #[cfg(any(feature = "rustls", feature = "https-native-tls"))]
             HttpStream::Secured(inner, timeout_at) => {
                 timeout(inner.get_ref(), *timeout_at)?;
                 inner.read(buf)
@@ -111,7 +111,7 @@ impl Write for HttpStream {
                 set_socket_write_timeout(inner, *timeout_at)?;
                 inner.write(buf)
             }
-            #[cfg(feature = "rustls")]
+            #[cfg(any(feature = "rustls", feature = "https-native-tls"))]
             HttpStream::Secured(inner, timeout_at) => {
                 set_socket_write_timeout(inner.get_ref(), *timeout_at)?;
                 inner.write(buf)
@@ -137,7 +137,7 @@ impl Write for HttpStream {
                 set_socket_write_timeout(inner, *timeout_at)?;
                 inner.flush()
             }
-            #[cfg(feature = "rustls")]
+            #[cfg(any(feature = "rustls", feature = "https-native-tls"))]
             HttpStream::Secured(inner, timeout_at) => {
                 set_socket_write_timeout(inner.get_ref(), *timeout_at)?;
                 inner.flush()
@@ -158,13 +158,21 @@ impl Write for HttpStream {
     }
 }
 
-#[cfg(any(feature = "async-https-rustls", feature = "async-https-rustls-probe"))]
+#[cfg(any(
+    feature = "async-https-rustls",
+    feature = "async-https-rustls-probe",
+    feature = "async-https-native-tls"
+))]
 type AsyncSecuredStream = rustls_stream::AsyncSecuredStream;
 
 #[cfg(feature = "async")]
 pub(crate) enum AsyncHttpStream {
     Unsecured(AsyncTcpStream),
-    #[cfg(any(feature = "async-https-rustls", feature = "async-https-rustls-probe"))]
+    #[cfg(any(
+        feature = "async-https-rustls",
+        feature = "async-https-rustls-probe",
+        feature = "async-https-native-tls"
+    ))]
     Secured(Box<AsyncSecuredStream>),
 }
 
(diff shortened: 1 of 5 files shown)


ton-anywhere:
CI failing because it's treating `_https-rustls` as public standalone feature. If this PR is Ok, I can adjust it on the rust-bitcoin-maintainer-tools project.

bitreq/src/connection/rustls_stream.rs
```diff
@@ -1,19 +1,19 @@
...
 
-#[cfg(feature = "rustls")]
+#[cfg(feature = "_https-rustls")]
 use alloc::sync::Arc;
 use std::io;
 use std::net::TcpStream;
 use std::sync::OnceLock;
 
-#[cfg(all(feature = "native-tls", not(feature = "rustls")))]
+#[cfg(all(feature = "https-native-tls", not(feature = "_https-rustls")))]
```
TheBlueMatt: Can we gate this whole module on rustls so this isn't so much of a mess, or at least move to `mod rustls_stuff  { ... }; pub use rustls_stuff::*`?

TheBlueMatt (commented):
Hmm, rather than an indirect feature can we gate on the actual features we need?

ton-anywhere:
@TheBlueMatt  you mean using `any(feature = "https-rustls", feature = "https-rustls-probe"` instead of using `_https-rustls`, and than refactor it into separate modules for rustls and native-tls so we don't have as many cfg gates? I think it would work really well! If yes, I can work on this. :+1:

TheBlueMatt:
Yea, that is what I was thinking.

ton-anywhere:
@TheBlueMatt updated

bitreq/src/connection.rs
```diff
@@ -29,14 +29,51 @@ use crate::{Error, Method, ResponseLazy};
...
+))]
 mod rustls_stream;
-#[cfg(feature = "rustls")]
-type SecuredStream = rustls_stream::SecuredStream;
+
+#[cfg(any(
+    all(
+        feature = "https-native-tls",
+        not(any(feature = "https-rustls", feature = "https-rustls-probe"))
+    ),
```
TheBlueMatt, line 44: I don't think we intend to allow selection of a separate sync-tls-provider and async one.

bitreq/src/connection.rs
```diff
@@ -29,14 +29,51 @@ use crate::{Error, Method, ResponseLazy};
...
+        not(any(feature = "async-https-rustls", feature = "async-https-rustls-probe"))
+    )
+))]
+mod native_tls_stream;
+
+#[cfg(all(
+    feature = "https-native-tls",
+    not(any(feature = "https-rustls", feature = "https-rustls-probe"))
+))]
+use self::native_tls_stream as sync_tls_stream;
```
TheBlueMatt, line 56: Its probably simpler to just have a sub-module in `rustls_stream` (maybe renaming it) and re-export. That way there's a single API to `connection.rs` and we arent conditional-compiling which module to use here.

bitreq/src/connection/native_tls_stream.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    #[cfg(feature = "log")]
+    log::trace!("Establishing TLS session to {host}.");
+
+    let tls = connector.connect(host, tcp).map_err(native_tls_err)?;
+
+    Ok(HttpStream::Secured(Box::new(tls), None))
+}
+
+// === ASYNC native-tls ===
```
TheBlueMatt, line 57: Please don't commit claude's useless code section definition comments for a tiny file.

ton-anywhere: @TheBlueMatt  come on, do you even read what you write? I'm trying to help here but it's been really hard to work with you.

You commited the same type o comment in the master.  this one is no different:
https://github.com/rust-bitcoin/corepc/blame/46c7bf4284748ee3da549d4463841e67db3bd00e/bitreq/src/connection/rustls_stream.rs#L69
image

I don't even use the fking Claude. I'm spending hours on all those pull requests. Come on. Let's be a bit more professionals here.

I agree 100% about removing all those useless comments, but this is the useless pattern you've already used there..

bitreq/src/connection.rs
ton-anywhere: What is your suggestion?
The alternatives I came up were either convoluted or do not pass in tests like this one:

`cargo --locked build --no-default-features "--features=async-https-native-tls https async-https-rustls native-tls rustls rustls-webpki std tokio webpki-roots"`

I was using this and it failed:

```
    #[cfg[all(
        any(feature = "async-https-native-tls", feature = "https-native-tls"),
        not(any(feature = "https-rustls", feature = "https-rustls-probe"))
    )]
```

Because it was opening all async gates(async backend present) but rustls priority was taking over. 

The simplest solution I can think of without a large rework on all async gates(many) would be just ignore sync on priorities when an async feature is on. But that would change in behavior we intended(rustls priority over native-tls). Otherwise we could simply not compile, but that's again a change in behavior.

But, I bet most of the users won't mix with both configs at the same time. So if they set a async feature it will be set for both sync/async.  The gate was necessary  to work on edge cases like the one above..

bitreq/src/connection.rs
ton-anywhere: really? using a single file to store code for both backends is a missing opportunity IMO. You mean like the enabled/disabled stuff I've initially tough for the tls configs on client.rs? That's doable, but I don't see what's are we winning here.

bitreq/src/connection.rs
TheBlueMatt: > I was using this and it failed:

Yes, that makes more sense, no? Your command there passes with just that one change, do you have a specific thing that's failing? Its not clear to me how many cfgs would have to change for this.

> Because it was opening all async gates(async backend present) but rustls priority was taking over.

Yes, that seems correct? If we have an async feature we should enable async and if we have a rustls feature (even sync!) we should use rustls for async, imo. This allows for the case of "downstream application enables the rustls feature because they want rustls, and they want it everywhere, even if some other library enabled native-tls"

bitreq/src/connection.rs
TheBlueMatt: The maze of conditional compilations is kinda irritating, so, yea, reducing them would be nice. It doesn't strictly have to be a single file, but that seems like the easiest way to get a smaller set of conditional compilation flags. You could also move the conditional-module up to connection.rs if you feel strongly about multiple files.

bitreq/src/connection/native_tls_stream.rs
TheBlueMatt: Apologies if I struck a nerve. Whether there was a line in the old code or not, we probably don't need to have separators in a 50-line file. Its certainly not a critical note, just suggesting ways to clean this up a bit, because its a mess (though sadly that's not your fault, rust's conditional compilation sucks and we have to fight with it...). Claude is famous for doing that, so I assumed you likely were using it (or some other LLM) - this project certainly has no issue with LLM use, nor is that a judgement.

tcharding: By 'project' he means `bitreq`. I personally I am super allergic to LLM output and bite peoples heads off all the time ... bots are making my days less enjoyable and the fallout is non-bot stuff is getting tarred with the same brush. We all have to evolve I guess.

tcharding:

PR #570, opened 2026-04-24
electrsd: include in workspace and CI

### LINKED ISSUE #552: Include `electrsd` in the workspace
Currently `electrsd` is excluded from the workspace. One reason is that `cargo test --all-features` does not work for `electrsd`.

Likely we can do the same feature gate dance that we do in `bitcoind` to get this working.

To close this issue:

- Do what ever is needed to get `--all-features` working
- Configure `cargo rbmt` stuff as needed in `electrsd` manifest.
- Move `electrsd` to the `members` array in the repo manifest.

satsfy:
Closes #552

This PR includes electrsd in corepc workspace. With this change, now test, lints and formats also consider electrsd.

In order for this inclusion to fit in workspace:
- The `cargo --all-features` test faced several version conflicts becasue these tests enable all electrsd feature flags at the same time. The solution developed was to create a new test-only feature flag called `all_features_test` that is gets also enabled by default for finer version control, such that only the latest electrsd version runs.
- Version compiler constant is gated through the new flag, making only the latest electrsd version activated.
- All version-specific call sites have handle the new flag.
- Fixes formatting and linting errors as well as adding appropriate rbmt lint configurations for electrsd.
- Updates corresponding lock files for electrsd dependencies.

All of the commits are ordered to the same end, enabling CI to pass and tests to run.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -36,6 +36,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BITCOIND_SKIP_DOWNLOAD: "1"
+      ELECTRSD_SKIP_DOWNLOAD: "1"
     strategy:
       fail-fast: false
       matrix:
@@ -61,6 +62,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BITCOIND_SKIP_DOWNLOAD: "1"
+      ELECTRSD_SKIP_DOWNLOAD: "1"
     strategy:
       fail-fast: false
       matrix:
@@ -78,6 +80,7 @@ jobs:
     runs-on: ubuntu-latest
     env:
       BITCOIND_SKIP_DOWNLOAD: "1"
+      ELECTRSD_SKIP_DOWNLOAD: "1"
     strategy:
       fail-fast: false
       matrix:
@@ -184,7 +187,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - uses: dtolnay/rust-toolchain@stable
-      - run: cd electrsd && cargo test --features ${{ matrix.features }}
+      - run: cd electrsd && cargo test --no-default-features --features ${{ matrix.features }}
 
 
   test-electrs-esplora:
@@ -201,7 +204,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - uses: dtolnay/rust-toolchain@stable
-      - run: cd electrsd && cargo test --features ${{ matrix.features }}
+      - run: cd electrsd && cargo test --no-default-features --features ${{ matrix.features }}
 
   test-electrs-no-download:
     name: Test Electrs no auto-download features
@@ -209,6 +212,7 @@ jobs:
     env:
       CARGO_TERM_COLOR: always
       RUST_LOG: debug
+      ELECTRSD_SKIP_DOWNLOAD: "1"
       ELECTRS_EXEC: "/home/runner/.cargo-install/electrs/bin/electrs"
     steps:
       - uses: dtolnay/rust-toolchain@stable
@@ -234,8 +238,10 @@ jobs:
         uses: baptiste0928/cargo-install@v2
         with:
           crate: electrs
+          version: 0.10.6
+          locked: true
       - run: "echo $ELECTRS_EXEC"
       - name: Checkout Crate
         uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
-      - run: cd electrsd && cargo test --features 'bitcoind_25_2,bitcoind_download'
+      - run: cd electrsd && cargo test --no-default-features --features 'bitcoind_25_2,electrs_0_10_6,bitcoind_download'
--- a/electrsd/Cargo.toml
+++ b/electrsd/Cargo.toml
@@ -59,3 +59,19 @@ bitcoind_0_20_2 = ["bitcoind/0_20_2"]
 bitcoind_0_19_1 = ["bitcoind/0_19_1"]
 bitcoind_0_18_1 = ["bitcoind/0_18_1"]
 bitcoind_0_17_2 = ["bitcoind/0_17_2"]
+
+[package.metadata.rbmt.lint]
+# The electrsd crate requires a version feature to compile
+features = ["electrs_0_10_6"]
+# Duplicates from build-dep/dep splits, --target=all, or upstream version conflicts.
+allowed_duplicates = [
+    "anyhow",           # build-dep and regular dep, same version
+    "base64",           # upstream version conflicts
+    "bitcoin_hashes",   # build-dep (download) and transitive dep, same version
+    "bitflags",         # upstream version conflicts
+    "bitreq",           # build-dep (download) and regular dep, same version
+    "libc",             # same version, appears twice under --target=all
+    "linux-raw-sys",    # same version, appears twice under --target=all
+    "rustix",           # same version, appears twice under --target=all
+    "windows-sys",      # upstream version conflicts
+]
--- a/electrsd/Cargo.toml
+++ b/electrsd/Cargo.toml
@@ -24,6 +24,7 @@ nix = { version = "0.25.0" }
 env_logger = { version = "0.10" }
 
 [build-dependencies]
+anyhow = { version = "1.0.66", default-features = false, features = ["std"], optional = true }
 bitcoin_hashes = { version = "0.14", optional = true }
 zip = { version = "0.6", default-features = false, optional = true, features = [
   "bzip2",
@@ -37,7 +38,7 @@ default = ["electrs_0_10_6"]
 legacy = []
 
 # download is not supposed to be used directly only through selecting one of the version feature
-download = ["bitcoin_hashes", "zip", "bitreq"]
+download = ["anyhow", "bitcoin_hashes", "zip", "bitreq"]
 
 esplora_a33e97e1 = ["download", "legacy"]
 electrs_0_8_10 = ["download"]
--- a/electrsd/Cargo.toml
+++ b/electrsd/Cargo.toml
@@ -60,6 +60,9 @@ bitcoind_0_19_1 = ["bitcoind/0_19_1"]
 bitcoind_0_18_1 = ["bitcoind/0_18_1"]
 bitcoind_0_17_2 = ["bitcoind/0_17_2"]
 
+# This flag should never be manually set. Used to handle `--all-features` was used.
+all_features = []
+
 [package.metadata.rbmt.lint]
 # The electrsd crate requires a version feature to compile
 features = ["electrs_0_10_6"]
--- a/electrsd/Cargo.toml
+++ b/electrsd/Cargo.toml
@@ -32,6 +32,8 @@ zip = { version = "0.6", default-features = false, optional = true, features = [
 bitreq = { version = "0.3.5", path = "../bitreq", optional = true, features = ["https"] }
 
 [features]
+default = ["electrs_0_10_6"]
+
 legacy = []
 
 # download is not supposed to be used directly only through selecting one of the version feature
@@ -62,6 +64,10 @@ bitcoind_0_17_2 = ["bitcoind/0_17_2"]
 
 # This flag should never be manually set. Used to handle `--all-features` was used.
 all_features = []
+
+[package.metadata.docs.rs]
+features = ["electrs_0_10_6"]
+rustdoc-args = ["--cfg", "docsrs"]
 
 [package.metadata.rbmt.lint]
 # The electrsd crate requires a version feature to compile
--- a/electrsd/build.rs
+++ b/electrsd/build.rs
@@ -2,18 +2,17 @@
 fn main() {}
 
 #[cfg(feature = "download")]
-fn main() {
-    download::download()
-}
+fn main() { download::download() }
 
 #[cfg(feature = "download")]
 mod download {
-    use bitcoin_hashes::{sha256, Hash};
     use std::fs::File;
     use std::io::{BufRead, BufReader, Cursor};
     use std::os::unix::fs::PermissionsExt;
     use std::path::Path;
     use std::str::FromStr;
+
+    use bitcoin_hashes::{sha256, Hash};
 
     include!("src/versions.rs");
 
@@ -45,17 +44,13 @@ mod download {
         let expected_hash = get_expected_sha256(&download_filename).unwrap();
         let out_dir = std::env::var_os("OUT_DIR").unwrap();
         let electrs_exe_home = Path::new(&out_dir).join("electrs");
-        let destination_filename = electrs_exe_home
-            .join(&download_filename_without_extension)
-            .join("electrs");
+        let destination_filename =
+            electrs_exe_home.join(&download_filename_without_extension).join("electrs");
 
         dbg!(&destination_filename);
 
         if !destination_filename.exists() {
-            println!(
-                "filename:{} version:{} hash:{}",
-                download_filename, VERSION, expected_hash
-            );
+            println!("filename:{} version:{} hash:{}", download_filename, VERSION, expected_hash);
 
             let download_endpoint =
                 std::env::var("ELECTRSD_DOWNLOAD_ENDPOINT").unwrap_or(GITHUB_URL.to_string());
@@ -73,11 +68,8 @@ mod download {
             let mut outfile = std::fs::File::create(&destination_filename).unwrap();
 
             std::io::copy(&mut file, &mut outfile).unwrap();
-            std::fs::set_permissions(
-                &destination_filename,
-                std::fs::Permissions::from_mode(0o755),
-            )
-            .unwrap();
+            std::fs::set_permissions(&destination_filename, std::fs::Permissions::from_mode(0o755))
+                .unwrap();
         }
     }
 }
--- a/electrsd/build.rs
+++ b/electrsd/build.rs
@@ -1,8 +1,9 @@
-#[cfg(not(feature = "download"))]
-fn main() {}
+fn main() { download::start().unwrap(); }
 
-#[cfg(feature = "download")]
-fn main() { download::download() }
+#[cfg(any(docsrs, not(feature = "download")))]
+mod download {
+    pub(crate) fn start() -> Result<(), ()> { Ok(()) }
+}
 
 #[cfg(feature = "download")]
 mod download {
@@ -19,35 +20,29 @@ mod download {
     const GITHUB_URL: &str =
         "https://github.com/RCasatta/electrsd/releases/download/electrs_releases";
 
-    fn get_expected_sha256(filename: &str) -> Result<sha256::Hash, ()> {
-        let file = File::open("sha256").map_err(|_| ())?;
+    fn get_expected_sha256(filename: &str) -> anyhow::Result<sha256::Hash> {
+        let file = File::open("sha256")?;
         for line in BufReader::new(file).lines().map_while(Result::ok) {
             let tokens: Vec<_> = line.split("  ").collect();
             if tokens.len() == 2 && filename == tokens[1] {
-                return sha256::Hash::from_str(tokens[0]).map_err(|_| ());
+                return Ok(sha256::Hash::from_str(tokens[0]).unwrap());
             }
         }
-        Err(())
+        panic!("no sha256 entry for {} in electrsd/sha256", filename);
     }
 
-    pub fn download() {
+    pub(crate) fn start() -> anyhow::Result<()> {
         if std::env::var_os("ELECTRSD_SKIP_DOWNLOAD").is_some() {
-            return;
+            return Ok(());
         }
 
-        if !HAS_FEATURE {
-            return;
-        }
         let download_filename_without_extension = electrs_name();
         let download_filename = format!("{}.zip", download_filename_without_extension);
-        dbg!(&download_filename);
-        let expected_hash = get_expected_sha256(&download_filename).unwrap();
+        let expected_hash = get_expected_sha256(&download_filename)?;
         let out_dir = std::env::var_os("OUT_DIR").unwrap();
         let electrs_exe_home = Path::new(&out_dir).join("electrs");
         let destination_filename =
             electrs_exe_home.join(&download_filename_without_extension).join("electrs");
-
-        dbg!(&destination_filename);
 
         if !destination_filename.exists() {
             println!("filename:{} version:{} hash:{}", download_filename, VERSION, expected_hash);
@@ -56,20 +51,23 @@ mod download {
                 std::env::var("ELECTRSD_DOWNLOAD_ENDPOINT").unwrap_or(GITHUB_URL.to_string());
             let url = format!("{}/{}", download_endpoint, download_filename);
 
-            let downloaded_bytes = bitreq::get(url).send().unwrap().into_bytes();
+            let downloaded_bytes = bitreq::get(url.clone()).send()?.into_bytes();
 
             let downloaded_hash = sha256::Hash::hash(&downloaded_bytes);
-            assert_eq!(expected_hash, downloaded_hash);
+            assert_eq!(expected_hash, downloaded_hash, "expected hash of {} is not matching", url);
             let cursor = Cursor::new(downloaded_bytes);
 
-            let mut archive = zip::ZipArchive::new(cursor).unwrap();
-            let mut file = archive.by_index(0).unwrap();
-            std::fs::create_dir_all(destination_filename.parent().unwrap()).unwrap();
-            let mut outfile = std::fs::File::create(&destination_filename).unwrap();
-
+            let mut archive = zip::ZipArchive::new(cursor)?;
+            let mut file = archive.by_index(0)?;
+            let parent = destination_filename.parent().unwrap();
+            std::fs::create_dir_all(parent)?;
+            let mut outfile = std::fs::File::create(&destination_filename)?;
             std::io::copy(&mut file, &mut outfile).unwrap();
-            std::fs::set_permissions(&destination_filename, std::fs::Permissions::from_mode(0o755))
-                .unwrap();
+            std::fs::set_permissions(
+                &destination_filename,
+                std::fs::Permissions::from_mode(0o755),
+            )?;
         }
+        Ok(())
     }
 }
--- a/electrsd/build.rs
+++ b/electrsd/build.rs
@@ -22,7 +22,7 @@ mod download {
 
     fn get_expected_sha256(filename: &str) -> Result<sha256::Hash, ()> {
         let file = File::open("sha256").map_err(|_| ())?;
-        for line in BufReader::new(file).lines().flatten() {
+        for line in BufReader::new(file).lines().map_while(Result::ok) {
             let tokens: Vec<_> = line.split("  ").collect();
             if tokens.len() == 2 && filename == tokens[1] {
                 return sha256::Hash::from_str(tokens[0]).map_err(|_| ());
--- a/electrsd/src/ext.rs
+++ b/electrsd/src/ext.rs
@@ -4,7 +4,8 @@
 use std::thread;
 use std::time::Duration;
 
-use electrum_client::{bitcoin::Txid, ElectrumApi};
+use electrum_client::bitcoin::Txid;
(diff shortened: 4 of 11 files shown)


electrsd/src/versions.rs
```diff
@@ -10,18 +10,20 @@ const OS: &str = "undefined";
...
-#[cfg(feature = "electrs_0_9_1")]
+#[cfg(all(feature = "electrs_0_9_1", not(feature = "esplora_a33e97e1")))]
 const VERSION: &str = "v0.9.1";
 
-#[cfg(feature = "electrs_0_9_11")]
+#[cfg(all(feature = "electrs_0_9_11", not(feature = "electrs_0_9_1")))]
 const VERSION: &str = "v0.9.11";
 
-#[cfg(feature = "electrs_0_10_6")]
+#[cfg(all(feature = "electrs_0_10_6", not(feature = "electrs_0_9_11")))]
```
jamillambert: These guards only exclude the previous version, not all previous versions. Which I assume was the intention.

e.g. `cargo build --features electrs_0_8_10,electrs_0_10_6` enables two different versions of the `const` and therefore fails to build. It only works with `--all-features` because there is only one `const` without a `not(feature)`. If the intention is just to get all-features to build there must be clearer way.

electrsd/src/lib.rs
```diff
@@ -189,18 +180,18 @@ impl ElectrsD {
         args.push("--network");
         args.push(conf.network);
 
-        #[cfg(not(feature = "legacy"))]
+        #[cfg(not(all(feature = "esplora_a33e97e1", not(feature = "electrs_0_8_10"))))]
```
jamillambert: These changes are confusing. It looks like it is gated on `electrs_0_8_10`. And gating on legacy shouldn't need changing, in an all features build only the legacy version is used.

EDIT: is this the fix for electrsd 0.8.10 on your system? if so can you put it in a separate commit to save this confusion.

electrsd/src/versions.rs
satsfy: Thanks! My mistake here, I imported the  `bitcoind/src/versions.rs` features patterns, but forgot to include the constraint portion that forces the user to select an electrsd version. Will include that back in.

electrsd/src/lib.rs
satsfy: This change is necessary because  --all-features these 2 flags simultaneously.

I will refactor the section in a separate commit because esplora handles cookies differently. In another commit I will also isolate a test behavior change for 0.8.10 too.

satsfy:
One thing I thought to include is a default electrsd version, like bitcoind in `0_17_2`, I suppose either `0_8_10` or `0_10_6`. I have not included a default because it required including `std` (for bitreq's `get()` to download), and I'm not sure if it's this PR's scope.

Edit: I suppose we need the default so tests work with an explicit version.

electrsd/src/lib.rs
satsfy: The legacy flag was removed because it was not semantically meaningful anymore on latest change.

satsfy:
Latest force push improves the legibility of the solution, sets a default electrs version, reworked the version feature graph so the `cfg` cascade in `versions.rs` is unambiguous. Replaced the `legacy` feature with `versions::USE_LEGACY_COOKIE` so `--all-features` tests pass.

tcharding:
Needs rebase mate. Out of interest what made you go with `electrs_0_8_10` instead of `electrs_0_10_6`?

satsfy:
Great point, I made a mistake regarding `0_8_10`. I had matched bitcoind pattern, but accidentally reversed the order of the versions. The highest version in bitcoind chain is the lowest, meanwhile I put the highest for electrsd. Latest rebase reverses that order.

EDIT: The problem with v0.10.6 in CI was caused by using the legacy `--jsonrpc-import` arg after 0.8.10. That nuance was lost on me because execution would fail on my Ubuntu 22.04 for requiring GLIBC 2.38 to test 0.10.6 properly.

satsfy:
On the latest rebase:
- New electrs feature version cascade: `electrs_0_10_6 -> 0_9_11 -> 0_9_1 -> 0_8_10 -> esplora`.
- Centralized the version-specific behavior in `versions.rs` with legible constants `USE_LEGACY_COOKIE`, `USE_JSONRPC_IMPORT`, `USE_VERBOSE_ARG`.
- Esplora is tested with `--no-default-features` so it does not accidentally combine with the default electrs version.
- `no-download` job now enables `bitcoind_download` because it still needs a bitcoind binary while using the system-provided `ELECTRS_EXEC`.
- Corrected the bitcoind `23_2` feature name in [ee9443c](https://github.com/rust-bitcoin/corepc/pull/570/commits/ee9443cdeb3c2b66326a1b21e1947fd5d7e8efd4) (is this correct? Should I have done it in a separate PR)? 
- Optional change that I think should be merged: make electrsd CI coverage easier to debug by putting `RUST_LOG=debug` on electrsd jobs.

tcharding:
Can we separate the fmt changes from actual code changes in c61ea18df38ee6976e25bbf7497fe386ad5af370

electrsd/src/ext.rs
```diff
@@ -4,12 +4,13 @@
...
 
-use electrum_client::{bitcoin::Txid, ElectrumApi};
+use electrum_client::bitcoin::Txid;
+use electrum_client::ElectrumApi;
 
 use crate::ElectrsD;
 
 impl ElectrsD {
-    #[cfg(not(feature = "electrs_0_8_10"))]
+    #[cfg(not(all(feature = "electrs_0_8_10", not(feature = "electrs_0_9_1"))))]
```
tcharding: These feature gates are making my eyes bleed. not, all, not - is there any way we can make this easier to reason about?

satsfy:
Latest rebase:
- Split rustfmt cleanup (17866ee8b27c9c2418f9e8b7338bf6e4f2d5ef0e), lint error fixes (33158b2aad27c09fba4f98088466c3c39a1a9d1e) and new rbmt lint configuration (91437126548c6c07005bcb45abb72b09e7eecf45).
- Fix Cargo.toml link typo (28022c5bc15f44363c5e38819001bb47927ae686)
- Simplifies version conditions:
  - Using `any` in `#[cfg(any(feature = "electrs_0_9_1", not(feature = "electrs_0_8_10")))]`
  - or `crate::versions::IS_ELECTRS_0_8_10`.

electrsd/src/ext.rs
satsfy: I simplified version conditions with `any` like
```rs
#[cfg(any(feature = "electrs_0_9_1", not(feature = "electrs_0_8_10")))]
```
And also used simpler conditions like `crate::versions::IS_ELECTRS_0_8_10` in another spot.

satsfy:
Latest rebase updates lockfiles because of upstream changes to fix ci

electrsd/src/ext.rs
```diff
@@ -4,12 +4,13 @@
...
 
-use electrum_client::{bitcoin::Txid, ElectrumApi};
+use electrum_client::bitcoin::Txid;
+use electrum_client::ElectrumApi;
 
 use crate::ElectrsD;
 
 impl ElectrsD {
-    #[cfg(not(feature = "electrs_0_8_10"))]
+    #[cfg(any(feature = "electrs_0_9_1", not(feature = "electrs_0_8_10")))]
```
tcharding: Can you explain what you want to achieve with this feature gate please?

tcharding: E.g "I want this code block to run for these versions: a, b, c"

tcharding:
No need for the last patch mate, we have #571 ready to release that change. I was planning on merging this after that release.

electrsd/src/ext.rs
satsfy: So bitcoind feature gates, which served as a model for making electrsd features, are organized into a chain:
```toml
30_2 = ["30_0"]
30_0 = ["29_0"]
...
0_18_1 = ["0_17_2"]
0_17_2 = []
```
In bitcoind's file [client_versions.rs:48](https://github.com/rust-bitcoin/corepc/blob/49ffe90d36d6aa42200ad7ce6fd4fca527e90d1e/bitcoind/src/client_versions.rs#L48), you can find a similar example of this gate:
```rs
#[cfg(all(feature = "0_21_2", not(feature = "22_1")))]
pub use corepc_client::{client_sync::v21::*, types::v21 as vtype};
```
So the pattern is `all(X, not(Y))`. 
Now imagine I want the negation of a particular version. Say, I need `not(version X)`. 
I would express that in feature gates as `not(all(X, not(Y)))`
But you mentioned above its hard to understand what that means and I agree.
So I did `not(all(X, not(Y)))` = `any(not(X), Y)` = `any(Y, not(X))` (De Morgan's laws) such as in:
```rs
#[cfg(any(feature = "electrs_0_9_1", not(feature = "electrs_0_8_10")))]
```
Would you prefer if I experimented with some less complex ways of solving the problem? I suppose I'd need to break away from the bitcoind solution and design a new feature flag solution. We do things this way to prevent users from enabling 2 flags at the same time, and the program not knowing which to use.

The alternative would be writing an "O(n²)" solution where every flag is checked against every other flag to ensure the same functionality. On a new electrsd version N, the versions file would be incremented in at least N lines of code.

satsfy:
Rebased removing `Cargo.toml` typo patch.

electrsd/src/ext.rs
jamillambert: Could this check be done in one place, e.g. `build.rs` and a new feature created like `electrs_0_8_10_only` then all of the `any` `not` feature gates changes can be `feature = "electrs_0_8_10_only"` etc. The new `const bool`s can also then be removed.

satsfy: That's a good solution, I implemented it on the latest force push.

satsfy:
Latest force push implements [configuration of version checks in the build file](https://github.com/rust-bitcoin/corepc/pull/570#discussion_r3169080380).

tcharding:
Sorry if I'm being brain dead but I don't see right now how the `_only` features work. Also we have 

```rust
[cfg(all(feature = "esplora_a33e97e1", not(feature = "electrs_0_8_10")))]
```

which would be able to be an `_only` feature if those worked how I think you mean them to, right?

tcharding:
The changes in d165d9d6 merge already. I'm surprised there is no conflicts for this PR now.

satsfy:
In the latest force push, I have rebased to kill compatibility concerns.

>which would be able to be an _only feature if those worked how I think you mean them to, right?

That is a good point, I really had missed it. Features like `electrs_0_9_1_only` are now used, I verified everywhere the solution could benefit from it and implemented there. For example, the legacy flags still is removed because all we need is to use esplora only feature.

> Sorry if I'm being brain dead but I don't see right now how the _only features work. 

So when you use version `3`, features `[3,2,1]` get activated in cascade (like bitcoind). When you need version `1`, only `[1]` gets activated. Previously, if you needed something to happen at only version 2, you needed to say `cfg!(2, not(3))`. This ensures that only one version gets activated at a time: it is always the highest version available in "flagspace". `build.rs` now basically iterates through all flags from highest to lowest, and whenever it hits flag `3` (highest), it stops iterating and creates a new flag called `3_only` in the "flagspace". So the flag `X_only` really now is the user-selected flag, `X`. The user doesn't see these `_only` flags. He just sets `elecstrsd = { 10_6_0 }` and everything works for him. He is either obligated to select a version or forced to use the default, and only one version gets used at a time.

tcharding:
ooooo is `CARGO_FEATURE_foo` set by the compiler? I totally missed that. Thanks for your patience.

electrsd/src/ext.rs
```diff
@@ -87,11 +85,15 @@ mod test {
             .txid()
             .unwrap();
 
+        #[cfg(electrs_0_8_10_only)]
+        {
+            // the 0.8.10 version doesn't have a mempool, so we need to mine the tx
+            bitcoind.client.generate_to_address(1, &generate_address).unwrap();
+            electrsd.trigger().unwrap();
+        }
```
tcharding: So it's intentional that this code block runs for esplora and all versions of electrs above 0.8.10?

electrsd/src/lib.rs
```diff
@@ -85,11 +83,7 @@ pub struct Conf<'a> {
 
 impl Default for Conf<'_> {
     fn default() -> Self {
-        let args = if cfg!(feature = "electrs_0_9_1")
-            || cfg!(feature = "electrs_0_8_10")
-            || cfg!(feature = "esplora_a33e97e1")
-            || cfg!(feature = "legacy")
-        {
+        let args = if cfg!(any(esplora_a33e97e1_only, electrs_0_8_10_only, electrs_0_9_1_only)) {
```
tcharding: So this is elects 9.1 and below, right? Why not say `not(electrs_0_9_11)`

electrsd/src/lib.rs
```diff
@@ -189,36 +176,24 @@ impl ElectrsD {
...
-            args.push("--cookie-file");
-            cookie_file = format!("{}", bitcoind.params.cookie_file.display());
-            args.push(&cookie_file);
-        }
-
-        #[cfg(feature = "legacy")]
-        let mut cookie_value;
-        #[cfg(feature = "legacy")]
-        {
+        let cookie_value = if cfg!(feature = "esplora_a33e97e1_only") {
```
tcharding: Cool, this is clear! Way better than `not(electrs_0_8_10)`.

tcharding:
Sorry for the iterations, this feature gating and weird feature setup (caused by the repo not you) is hard to get right. I'm going to need to come back to this with fresh eyes.

electrsd/Cargo.toml
```diff
@@ -31,21 +31,19 @@ zip = { version = "0.6", default-features = false, optional = true, features = [
...
 download = ["bitcoin_hashes", "zip", "bitreq"]
-
-esplora_a33e97e1 = ["download", "legacy"]
-electrs_0_8_10 = ["download"]
-electrs_0_9_1 = ["download"]
-electrs_0_9_11 = ["download"]
-electrs_0_10_6 = ["download"]
-
 bitcoind_download = ["bitcoind/download", "download"]
+default = ["electrs_0_10_6", "download"]
```
tcharding: Why is default to enable `download`? We don't do that in `bitcoind`, there was some reason we stopped doing so but I forget right now.

.github/workflows/rust.yaml
```diff
@@ -184,7 +187,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: Swatinem/rust-cache@v2
       - uses: dtolnay/rust-toolchain@stable
-      - run: cd electrsd && cargo test --features ${{ matrix.features }}
+      - run: cd electrsd && RUST_LOG=debug cargo test --features ${{ matrix.features }}
```
tcharding: Why do we need this when its set above? Does `RUST_LOG: debug` not work?

tcharding: But in `test-electrs-no-download` its not set. Can we pick one way and do it uniform?

electrsd/Cargo.toml
```diff
@@ -61,3 +59,26 @@ bitcoind_0_20_2 = ["bitcoind/0_20_2"]
...
+    "linux-raw-sys",    # same version, appears twice under --target=all
+    "rustix",           # same version, appears twice under --target=all
+    "windows-sys",      # upstream version conflicts
+]
+
+[lints.rust]
+unexpected_cfgs.level = "warn"
+unexpected_cfgs.check-cfg = [
+  'cfg(feature, values("electrs_0_9_11_only", "electrs_0_9_1_only", "electrs_0_8_10_only", "esplora_a33e97e1_only"))',
+  'cfg(esplora_a33e97e1)',
```
tcharding: We shouldn't need the last one because its a feature, right?

tcharding: Actually we don't need any of these do we because of the loop in `build.rs`?

tcharding:
```diff
+    for v in versions {
+        println!("cargo:rustc-check-cfg=cfg({}_only)", v);
+    }
```

electrsd/build.rs
```diff
@@ -1,28 +1,47 @@
...
+    // Tells rust these are valid and expected configs.
+    for v in versions {
+        println!("cargo:rustc-check-cfg=cfg({}_only)", v);
+    }
+
+    for v in versions {
+        let env_name = format!("CARGO_FEATURE_{}", v.to_uppercase());
+        if std::env::var_os(&env_name).is_some() {
+            // Emits cfg for the highest enabled version only and then returns.
+            println!("cargo:rustc-cfg={}_only", v);
```
tcharding:
```suggestion
            println!("cargo:rustc-cfg={}_and_below", v);
```

I think I finally get it. The 'only' was throwing me off before big time.

tcharding:
Sorry got to run half way through review.

tcharding:
Wrong button.

electrsd/Cargo.toml
satsfy: I'll remove it. I left it in because a few weeks ago, the download came by default for every feature flag of elecstrd, and some other PR removed, but I had left it on by default.

.github/workflows/rust.yaml
satsfy: Oh my bad, I had missed it. Will put it on env for every electrsd job.

electrsd/src/lib.rs
satsfy: Will change.

electrsd/src/ext.rs
satsfy: Hmm seems like the code block runs only on 0.8.10. In any other version, it doesn't run. I proved it by:
- Writing code in this code block that doesn't compile, for testing.
- Ran version immediately below 0.8.10 `cargo test --no-default-features --features bitcoind_22_1,bitcoind_download,esplora_a33e97e1` - Tests ok
- Ran version immediately next to 0.8.10 `cargo test ... electrs_0_9_11` - Tests ok
- Ran `cargo test ... electrs_0_8_10` - Fails to compile.

electrsd/Cargo.toml
satsfy: I think I had put it because of some `cargo check` or lint warning, but I can't reproduce it anymore, so I'm dropping the change.

satsfy:
Force-pushed. Changes:
- Rebase with master.
- Update lockfiles to fix CI errors.
- In `versions.rs`, replace `_only` cfg aliases with chained `not(...)` guards . It was a mistake to make that change because the solution would not compile for electrs versions below 0.10.6. I could not use the `_only` suffixed flags they are not present at build time, causing a compilation error. But if you do add to `Cargo.toml` (which I tried), they break `-all-features` tests because all the `_only` flags are included when only one of them should. 
- Move `RUST_LOG=debug` out of inline invocation and into the job env block.
- Add `--no-default-features` back to cargo test runs to prevent the default features from leaking into other versions in CI test.
- Drop `download` from the default feature set.
- Revert `cfg!(feature = "esplora_a33e97e1_only")` back to `cfg!(esplora_a33e97e1_only)`.

electrsd/build.rs
satsfy: Sorry, but the `_only` indeed means only a specific version. The latest change should clear up the confusion, because this older version was not using the only properly. It was an error.

electrsd/src/ext.rs
tcharding: Oh the problem is that in b3ca435033207206316cf9beedde81318c2d5bbe the feature gate is different so I posted this comment while reviewing in the terminal without reading the code here and without realising that its changed in a later patch. In general try not to change things in a later patch in the same PR, its fine now mentioning just for learning sake.

tcharding:
```rust
 
+        #[cfg(feature = "electrs_0_8_10")]
`

File: electrsd/src/versions.rs [hunk 12]
```diff
@@ -7,21 +7,33 @@ const OS: &str = "linux";
...
-const VERSION: &str = "esplora_a33e97e1a1fc63fa9c20a116bb92579bbf43b254";
+// An explicit version of Electrs must be selected by enabling some feature.
+// We check this here instead of in `lib.rs` because this file is included in `build.rs`.
+#[cfg(all(
+    not(feature = "electrs_0_8_10"),
+    not(feature = "electrs_0_9_1"),
+    not(feature = "electrs_0_9_11"),
+    not(feature = "electrs_0_10_6"),
+    not(feature = "esplora_a33e97e1"),
+))]
```
tcharding:
```suggestion
#[cfg(not(feature = "esplora_a33e97e1"))]
```

Recently worked out this can be simplified in `bitcoind` too. This works because of the cascading features. Same change in this file at the bottom.

electrsd/src/versions.rs
```diff
@@ -7,21 +7,33 @@ const OS: &str = "linux";
...
 
-#[cfg(feature = "electrs_0_9_11")]
+#[cfg(all(feature = "electrs_0_9_11", not(feature = "electrs_0_10_6")))]
 const VERSION: &str = "v0.9.11";
 
-#[cfg(feature = "electrs_0_10_6")]
-const VERSION: &str = "v0.10.6";
+#[cfg(all(feature = "electrs_0_9_1", not(feature = "electrs_0_9_11")))]
+const VERSION: &str = "v0.9.1";
+#[cfg(all(feature = "electrs_0_8_10", not(feature = "electrs_0_9_1")))]
```
tcharding: Why not use the `_only` flags?

tcharding:
Reviewed: 285011c943540ac6f6f924b3282b74ba5a9b6cf6

jamillambert (commented):
I am not sure the removal of the `legacy` feature is a good idea. This could break downstream for no real benefit.

Other than that and the few features gates that should be `_only` the rest looks good.

Reviewed 285011c943540ac6f6f924b3282b74ba5a9b6cf6

electrsd/src/versions.rs
jamillambert: There was another gate I thought was wrong then I realised the same, really the feature means `esplora_a33e97e1 and later versions`

electrsd/src/versions.rs
jamillambert: I agree, this was one of the reasons to have the new feature.

satsfy: I have removed these flags. They were too complex. What I did instead if to gate with.
```rs
#[cfg(all(feature = "electrs_0_8_10", not(feature = "all_features_test")))]
```
I think this is reasonable because it reads very clear meaning.

satsfy:
The latest force push reworks the solution a lot, after some suggestions:
- Revert almost all flag changes and simplify the PRs diff. Cascading features were entirely removed.
- Tighten scope, removing some unnecessary changes (the 0.8.10 unrelated fix and CI debug logs) to other PRs. 
- Removed the build-time-created flags.
- This is the simple solution I should have done from the start: add a `all_features_test` flag that gets enabled only when `cargo test --all-features` gets run. The user is never supposed to use that flag. Feature gating became more complex but much more readable.

NOTE: cannot move lint for format changes to other PRs because workspace now exercises it, causing CI failures. It is an inseparable part of the PR.

> I am not sure the removal of the legacy feature is a good idea

I'm new-ish to OSS, and didn't realize how this would affect downstream. The latest push adds it back and keeps the flag situation intact for downstream.

electrsd/src/versions.rs
satsfy: Cascading features were removed entirely in the latest version.

satsfy:
Latest push simple rebase + lockfile update.

tcharding:
Looking good. To review I decided to just hack around and see what feel out. Want to look at the last commit on this branch and pull out any changes that you like?

https://github.com/tcharding/corepc/tree/push-zmqllslturol

satsfy:
Latest force push takes suggestions from @tcharding's commit. Read bitcoind and electrsd side by side to pick what seemed appropriate. Also fixes the README instructions.

jamillambert (commented):
The lockfiles should be in `Include electrsd in workspace` 10a2bf0f618dddf9fd9fa7e1bd1ceda8ebda89e2, currently they are not correct for all the patches until then.

satsfy:
> lockfiles should be in Include electrsd in workspace

Done in latest force push

jamillambert (approved):
ACK d2c019ab912b189e8c11b00a2ebebd02f497d61d

Thanks for sticking with it.

tcharding:
Further follow ups if you want them, just things I noticed during this round of review.

`src/lib.rs` could be cleaner

```rust
//! Electrsd
//!
//! Utility to run a regtest electrsd process, useful in integration testing environment.

mod error;
mod ext;
mod versions;

pub extern crate bitcoind;
pub extern corepc_client;
pub extern crate electrum_client;

use std::env;
use std::ffi::OsStr;
use std::path::PathBuf;
use std::process::{Child, Command, Stdio};
use std::time::Duration;

use bitcoind::anyhow::Context;
use bitcoind::serde_json::Value;
use bitcoind::tempfile::TempDir;
use bitcoind::{anyhow, get_available_port, BitcoinD};
use electrum_client::raw_client::{ElectrumPlaintextStream, RawClient};
use log::{debug, error, warn};

#[rustfmt::skip]                // Keep pubic re-exports separate.
pub use error::Error;
```

- Remove `#![warn(missing_docs)]`

electrsd/src/lib.rs
```diff
@@ -189,43 +179,36 @@ impl ElectrsD {
...
         let p2p_socket;
-        if cfg!(feature = "electrs_0_8_10")
-            || cfg!(feature = "esplora_a33e97e1")
-            || cfg!(feature = "legacy")
+        if !cfg!(feature = "all_features")
+            && cfg!(any(
+                feature = "electrs_0_8_10",
+                feature = "esplora_a33e97e1",
+                feature = "legacy",
+            ))
```
tcharding: Mate this sort of if/not/not/if feature gating is what we are trying to get away from. The reason is that it is **way** too hard to read. I had to spend 3 minutes trying to workout what it means _and_ IIUC, and its telling that I'm not even 100% sure, its also broken.

For any build that does not use `--all-features` the first clause is false which makes the `&&` always false which makes the outer `!` always true. But anyway we want it simple, brain-dead-simple is the mantra.

satsfy: I see didn't make it simple enough. I'll make intermediary boolean logic variables so that this and other conditions read like:

```rs
if !is_all_features_test && should_version_use_jsonrpc_import {
``` 
The assumption here is that all feature tests would enable all flags, so we must specifically root it out.

However the boolean logic is correct. You can prove it by running locally:
```
cargo test --all-features
cargo test --no-default-features --features bitcoind_22_1,bitcoind_download,electrs_
0_9_11
cargo test --no-default-features --features bitcoind_22_1,bitcoind_download,electrs_
0_9_1
cargo test --no-default-features --features bitcoind_22_1,bitcoind_download,electrs_
0_8_10
cargo test --no-default-features --features bitcoind_22_1,bitcoind_download,esplora_a33e97e1
```

satsfy:
First force push: made the feature gating more readable and applied the suggestions for lib.rs header and crate re-exports.
Second force push: fix a commit order mistake.

electrsd/src/lib.rs
tcharding: Goodness, please excuse me. Yesterday I missed totally that there are 2 calls to `cfg!` - bother. I was reading it as this (which is obviously syntactically incorrect)
```
if !cfg!(feature = "all_features")
            && any(
                feature = "electrs_0_8_10",
                feature = "esplora_a33e97e1",
                feature = "legacy",
            )
```

apoelstra:
d64d55778a0d3b7246d197e4cce9fa399c5058b6 needs rebase

tcharding:
Aight, sorry for all the iterations on this. I'm happy with this now. Tiny nit, `_TEST` in `IS_ALL_FEATURES_TEST` might be slightly better as `_BUILD` because one can pass `--all-features` to `cargo check`, `cargo build`, or `cargo test`.

But anyways, you have to rebase. Ill merge after that whether or not you change the const name. Thanks for you patience and my at times dismal reviews on this one.

tcharding:
Reviewed: d64d55778a0d3b7246d197e4cce9fa399c5058b6

satsfy:
Force push rebases and modified the const name.

tcharding (approved):
ACK bfe8abc89f17b1903240765b4e32e3121e060b22

tcharding:
BOOM! Thanks for sicking with it.

PR #571, opened 2026-04-29
Release tracking PR: `electrsd 0.38.1`

tcharding:
We released a typo in the `electrsd` manifest. Requires release to push the changes to crates.io

Fix the typo, bump the version number, and add a changelog entry.

PR #542, opened 2026-04-08
Import `electrsd` and rename `corepc-node` back to `bitcoind`

tcharding:
Just so its clear. Here we import `electrsd` and bump the version so we can release it. We also re-name `corepc-node` back to the original name `bitcoind` and bump the version so it can be released also.

Tested with `cargo publish -p bitcoind --dry-run`

This is another attempt at #403. That one was so long ago I'm keeping it for the record and opening a new one.

This was done while `electrsd` is at  commit: `d2fd8af upgrade electrs esplora`. The full commit hash is in the git log.

PR #573, opened 2026-04-30
electrsd: fix bitcoind 23_2 feature version

satsfy:
Seems that bitcoind feature version name should have been 23_2 because that's what shows up in bitcoind.


satsfy:
Also, I see in [electrsd/README.md](https://github.com/rust-bitcoin/corepc/blob/49ffe90d36d6aa42200ad7ce6fd4fca527e90d1e/electrsd/README.md?plain=1#L21) that it suggests for the user to write the following:
```rs
electrsd = { version= "0.23", features = ["corepc-node_23_1", "electrs_0_9_1"] }
```
Is this 23_1 correct? I'm not sure so I didn't include in this PR.

tcharding:
> Also, I see in [electrsd/README.md](https://github.com/rust-bitcoin/corepc/blob/49ffe90d36d6aa42200ad7ce6fd4fca527e90d1e/electrsd/README.md?plain=1#L21) that it suggests for the user to write the following:
> 
> ```rust
> electrsd = { version= "0.23", features = ["corepc-node_23_1", "electrs_0_9_1"] }
> ```
> 
> Is this 23_1 correct? I'm not sure so I didn't include in this PR.

Ah that is totally stale. Should be `bitcoind_23_2`

tcharding:
Turns out my shell alias for grepping repos was to restrictive and there are a bunch of mentions of `corepc-node` still in various READMEs (my mistake from #542).

Can you do a separate PR that fixes all those please? Then this one can go in on top of that, and then we can point release any crates that got changed.

tcharding:
FWIW this changes is correct but means we have to do a major release not a point release.

tcharding:
I did the `electrs` readme fix already in https://github.com/rust-bitcoin/corepc/pull/571

jamillambert:
> Turns out my shell alias for grepping repos was to restrictive and there are a bunch of mentions of `corepc-node` still in various READMEs (my mistake from #542).

I looked at this too when reviewing that PR, and I didn't understand how I missed them so I looked again. I think it's because there are 1622 occurrences of "node" on master, so we had to be more specific, and the "corepc-node" that need to be changed are not in "bitcoind/".

Good catch though.

jamillambert (approved):
ACK ba1d7d904e06d242c6d3a40fb2f441e5203bfa2e

Should wait for #574 before cutting a new release.

PR #578, opened 2026-05-03
jsonrpc(fix): set host header to URL hostname instead of resolved IP

### LINKED ISSUE #577: jsonrpc: simple_http sets Host header to resolved IP instead of URL hostname
## Description

The `simple_http` transport in the `jsonrpc` crate constructs the HTTP `Host` header from the resolved socket address (`<IP>:<port>`) instead of the URL's authority component. This violates [RFC 9112 §3.2](https://www.rfc-editor.org/rfc/rfc9112#section-3.2):

  > A client MUST send a Host header field [...] in all HTTP/1.1 request messages. If the target URI includes an authority component, then a client MUST send a field value for Host that is identical to that authority component, excluding any userinfo subcomponent and its "@" delimiter [...].

Impact: any consumer of `Client::simple_http` cannot reach a server that sits behind a host-routed reverse proxy. The proxy receives `Host: <ip>:<port>`, which doesn't match any vhost rule, and serves its default backend (typically a 404).

## Reproduce

```sh
# Terminal 1: tiny listener that prints what it receives
python3 -c '
import socket
s = socket.socket(socket.AF_INET6); s.bind(("::", 8080)); s.listen()
c, _ = s.accept()
print(c.recv(4096).decode(errors="replace"))
c.send(b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n")
'
```

```sh
# Terminal 2: minimal client
cargo new --quiet /tmp/host-bug && cd /tmp/host-bug
echo 'jsonrpc = "0.19"' >> Cargo.toml
echo 'fn main() {
    let c = jsonrpc::Client::simple_http("http://localhost:8080/", None, None).unwrap();
    let _ = c.send_request(c.build_request("ping", None));
}' > src/main.rs
cargo run --quiet
```

Terminal 1 prints the actual outgoing request:

```
POST / HTTP/1.1
host: [::1]:8080                          <-- bug: resolved IP, not "localhost:8080"
Content-Type: application/json
Content-Length: 54

{"method":"ping","params":null,"id":1,"jsonrpc":"2.0"}
```

The `host:` value is `[::1]:8080` (the address `localhost` resolved to), but RFC 9110 §7.2 requires it to be `localhost:8080` (the host from the target URI). The bug only manifests when the URL hostname differs from its DNS resolution, exactly the case for any deployment behind a host-routed reverse proxy. (Switch to `http://127.0.0.1:8080/` and the symptom hides because the URL hostname *is* the resolved IP.)

## Expected behavior

The on-the-wire `Host` header should be the URL's authority component verbatim (with userinfo `user:pass@` stripped, IPv6 brackets preserved per RFC 7230 §2.7.1, and explicit/omitted ports following the URL):

```
POST / HTTP/1.1
host: localhost:8080
...
```

## Root cause

`jsonrpc/src/http/simple_http.rs`, three connected lines:

1. **Line 282** — `check_url()` calls `to_socket_addrs()` on the URL's authority and returns only `(SocketAddr, path)`. The hostname string exists transiently and is dropped after resolution.
2. **Lines 40-51** — `SimpleHttpTransport` has no field for the URL hostname; only `addr: SocketAddr` and `path: String`.
3. **Lines 152-153** — the request serializer writes `self.addr.to_string()` (resolved `<IP>:<port>`) to the `host:` header:
   ```rust
   request_bytes.write_all(b"host: ")?;
   request_bytes.write_all(self.addr.to_string().as_bytes())?;
   ```

The struct can't carry the hostname, so the parser has nowhere to put it, and the serializer has nothing but the IP to use.

## Environment

- `jsonrpc 0.19.0` from crates.io (current code on `master` is identical at the affected lines).
- Reproduces on Linux (also on macOS — same `simple_http.rs` codepath).

lorenzolfm:
### Description and Notes

Fixes [rust-bitcoin/corepc#577](https://github.com/rust-bitcoin/corepc/issues/577).

The `simple_http` writes the resolved socket address to the outgoing `Host` HTTP header instead of the URL's authority component, violating [RFC 9112 §3.2](https://www.rfc-editor.org/rfc/rfc9112#section-3.2):

> If the target URI includes an authority component, then a client MUST send a field value for Host that is identical to that authority component, excluding any userinfo subcomponent and its "@" delimiter.

### The fix

The bug: `SimpleHttpTransport` had no field for the URL hostname, so `check_url` couldn't preserve it across DNS resolution, so the request serializer had nothing but the resolved `SocketAddr` to put in the `Host` header. The fix updates all three.

1. **Struct** (`SimpleHttpTransport`, line 40): add a `host_header: String` field. Stores the URL's authority component verbatim `host` or `host:port`.

2. **URL parser** (`check_url`, line 282): now returns `(SocketAddr, host_header, path)` instead of `(SocketAddr, path)`. Captures `after_auth.to_owned()` before calling `to_socket_addrs()`.

3. **Wire serializer** (`try_request`, line 152-153): writes `self.host_header` to the `Host:` header instead of `self.addr.to_string()`.

`set_url` is updated to thread the new tuple. The `Default` impl initializes `host_header` consistently with the default `addr` (`127.0.0.1:DEFAULT_PORT`).

### The tests

Two new tests, plus an existing test extended.

**`request_uses_url_host_in_host_header`**: the regression test.

Binds a `TcpListener` on `localhost:0`, points the client at `http://localhost:<port>/`, captures the outgoing request bytes via `stream.read()`, and asserts the `host:` line of the request equals `localhost:<port>`. Against unfixed code it sees `[::1]:<port>` or `127.0.0.1:<port>` (whichever the system resolves `localhost` to first)

**`check_url_host_header`**: unit test for the parser.

Table over hostname / IP / IPv6-bracketed / userinfo URL forms. Asserts the URL authority is preserved in `host_header` verbatim:

```rust
("http://example.com/",                        "example.com"),
("http://example.com:8080/",                   "example.com:8080"),
("https://example.com/walletname",             "example.com"),
("https://example.com:443/",                   "example.com:443"),
("localhost:22",                               "localhost:22"),
("http://me:weak@localhost:22/wallet",         "localhost:22"),
("http://127.0.0.1:8332/",                     "127.0.0.1:8332"),
("http://[2001:0db8:85a3:...]:8300",           "[2001:0db8:85a3:...]:8300"),
```

The userinfo case (`me:weak@`) verifies credentials are stripped; the IPv6 case verifies brackets round-trip; the no-port and explicit-port cases verify both default-port branches.

**`test_urls`**: existing test, extended.

Already exercised the parser for all valid URLs. Updated to destructure the new 3-tuple and assert that `host_header` round-trips through `Builder::url()` alongside the existing `addr` and `path` checks. Same coverage as before plus the new field.

### DIFF
--- a/jsonrpc/src/http/simple_http.rs
+++ b/jsonrpc/src/http/simple_http.rs
@@ -39,6 +39,7 @@ const DEFAULT_TIMEOUT: Duration = Duration::from_millis(1);
 #[derive(Clone, Debug)]
 pub struct SimpleHttpTransport {
     addr: net::SocketAddr,
+    host_header: String,
     path: String,
     timeout: Duration,
     /// The value of the `Authorization` HTTP header.
@@ -57,6 +58,7 @@ impl Default for SimpleHttpTransport {
                 net::IpAddr::V4(net::Ipv4Addr::new(127, 0, 0, 1)),
                 DEFAULT_PORT,
             ),
+            host_header: format!("127.0.0.1:{}", DEFAULT_PORT),
             path: "/".to_owned(),
             timeout: DEFAULT_TIMEOUT,
             basic_auth: None,
@@ -83,7 +85,8 @@ impl SimpleHttpTransport {
     pub fn set_url(&mut self, url: &str) -> Result<(), Error> {
         let url = check_url(url)?;
         self.addr = url.0;
-        self.path = url.1;
+        self.host_header = url.1;
+        self.path = url.2;
         Ok(())
     }
 
@@ -150,7 +153,7 @@ impl SimpleHttpTransport {
         request_bytes.write_all(b" HTTP/1.1\r\n")?;
         // Write headers
         request_bytes.write_all(b"host: ")?;
-        request_bytes.write_all(self.addr.to_string().as_bytes())?;
+        request_bytes.write_all(self.host_header.as_bytes())?;
         request_bytes.write_all(b"\r\n")?;
         request_bytes.write_all(b"Content-Type: application/json\r\n")?;
         request_bytes.write_all(b"Content-Length: ")?;
@@ -279,7 +282,7 @@ impl SimpleHttpTransport {
 
 /// Does some very basic manual URL parsing because the uri/url crates
 /// all have unicode-normalization as a dependency and that's broken.
-fn check_url(url: &str) -> Result<(SocketAddr, String), Error> {
+fn check_url(url: &str) -> Result<(SocketAddr, String, String), Error> {
     // The fallback port in case no port was provided.
     // This changes when the http or https scheme was provided.
     let mut fallback_port = DEFAULT_PORT;
@@ -319,6 +322,11 @@ fn check_url(url: &str) -> Result<(SocketAddr, String), Error> {
         split.next().unwrap_or(s)
     };
 
+    // The Host header value is the URL's authority verbatim — including an
+    // explicit port if present, omitting it otherwise. Brackets around IPv6
+    // literals are preserved (RFC 7230 §2.7.1).
+    let host_header = after_auth.to_owned();
+
     // (4) Parse into socket address.
     // At this point we either have <host_name> or <host_name_>:<port>
     // `std::net::ToSocketAddrs` requires `&str` to have <host_name_>:<port> format.
@@ -331,7 +339,7 @@ fn check_url(url: &str) -> Result<(SocketAddr, String), Error> {
     };
 
     match addr.next() {
-        Some(a) => Ok((a, path.to_owned())),
+        Some(a) => Ok((a, host_header, path.to_owned())),
         None => Err(Error::url(url, "invalid hostname: error extracting socket address")),
     }
 }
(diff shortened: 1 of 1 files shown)


tcharding:
Appreciate the contribution. Is this LLM generated? The PR description smells a bit like it. Can you state what you wrote and what is bot written please. From your GitHub account it looks like you know what you are doing but I'd just like an idea so I can guide my review.

Thanks and beleza.

lorenzolfm:
The fix and  `check_url_host_header` was written by me. The `request_uses_url_host_in_host_header` and test cases for `check_url_host_header` was LLM generated.

The approach was first adding a test the failed. This is the LLM generated test `request_uses_url_host_in_host_header`. Then I tried to work backwards and see what was missing.

jamillambert (approved):
ACK 622faedb732aeac2492143bb4e8f7ffd874a20f9

This all looks good to me. Thanks for the issue and the fix.

lorenzolfm:
Hey @tcharding living a ping so you dont forget about this
Happy to address anything you think is worthwile as well

:)

luisschwab:
cc @satsfy

jsonrpc/src/http/simple_http.rs
```diff
@@ -279,7 +282,7 @@ impl SimpleHttpTransport {
 
 /// Does some very basic manual URL parsing because the uri/url crates
 /// all have unicode-normalization as a dependency and that's broken.
-fn check_url(url: &str) -> Result<(SocketAddr, String), Error> {
+fn check_url(url: &str) -> Result<(SocketAddr, String, String), Error> {
```
satsfy, line 285: Nit: I think a struct would be cleaner than using a tuple `(SocketAddr, String, String)` for the return value.

Didn't think it was worth holding up the PR for, but it would be a nice fix.

satsfy (approved):
tACK 622faedb732aeac2492143bb4e8f7ffd874a20f9 - Tested by removing fixes and running regression test, and it fails with `127.0.0.1:45439 =/= localhost:45439`, as expected.

Nice catch. Wonder how you found the error, proxies?

lorenzolfm:
> Nice catch. Wonder how you found the error, proxies?

Yeah, installed florestad on my homelab, exposing its rpc endpoint through nginx and try to call it on my computer

Issue #583, opened 2026-05-06
Bug: Internal panic in `bitreq` when stress-testing


ErikDeSmedt:
I hit the following error in bitreq during stress-testing

```
Panic occurred: panicked at /home/erik/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bitreq-0.3.5/src/connection.rs:401:17:
At all times, the next_request_id should be higher than the readable id

Backtrace:
   0: captaind::common::set_panic_hook::{{closure}}
             at /home/erik/code/ark/server/src/bin/captaind/../common/mod.rs:6:19
   1: <alloc::boxed::Box<F,A> as core::ops::function::Fn<Args>>::call
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/alloc/src/boxed.rs:1985:9
   2: std::panicking::rust_panic_with_hook
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:841:13
   3: std::panicking::begin_panic_handler::{{closure}}
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:699:13
   4: std::sys::backtrace::__rust_end_short_backtrace
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/sys/backtrace.rs:174:18
   5: __rustc::rust_begin_unwind
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/std/src/panicking.rs:697:5
   6: core::panicking::panic_fmt
             at /rustc/1159e78c4747b02ef996e55082b704c09b970588/library/core/src/panicking.rs:75:14
   7: bitreq::connection::AsyncConnection::send::{{closure}}
             at /home/erik/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bitreq-0.3.5/src/connection.rs:401:17
   8: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /nix/store/8zhq6gihfmj1a30dihin3zzyjc4n980x-rust-mixed/lib/rustlib/src/rust/library/core/src/future/future.rs:133:9
   9: bitreq::client::Client::send_async::{{closure}}
             at /home/erik/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/bitreq-0.3.5/src/client.rs:97:35
  10: bitcoind_async_client::client::Client::call::{{closure}}
[...]

PR #584, opened 2026-05-06
Some pipeline fixes in Bitreq

ErikDeSmedt:
We switched recently to [bitcoind-asyn-client](https://github.com/alpenlabs/bitcoind-async-client) and I saw some panics and requests getting stuck.. See https://github.com/rust-bitcoin/corepc/issues/583

I identified two issues
- A counter overflow on `next_request_id` which results in panics
- If a connection goes bad we throw-away some data. The result is that open requests stay open. They never succeed nor fail.

I didn't take the time to get deeply familiar with how bitreq works.
I'm happy to see suggestions on other approaches to fix the problem.

### DIFF
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -6,7 +6,7 @@ use std::net::{TcpStream, ToSocketAddrs};
 #[cfg(feature = "async")]
 use std::pin::Pin;
 #[cfg(feature = "async")]
-use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::atomic::{AtomicIsize, AtomicUsize, Ordering};
 #[cfg(feature = "async")]
 use std::sync::{Arc, Mutex};
 #[cfg(feature = "async")]
@@ -217,13 +217,19 @@ impl AsyncWrite for AsyncHttpStream {
     }
 }
 
+/// The starting value of [`AsyncConnectionState::permits`].
+#[cfg(feature = "async")]
+const MAX_SEND_PERMITS: isize = isize::MAX;
+
 #[cfg(feature = "async")]
 struct AsyncConnectionState {
     write: AsyncMutex<WriteHalf<AsyncHttpStream>>,
     read: AsyncMutex<tokio::io::BufReader<ReadHalf<AsyncHttpStream>>>,
-    /// The ID of the next request we'll send. If this reaches [`usize::MAX`] no further requests
-    /// can be sent on this socket and a new connection must be made. Thus, in order to limit the
-    /// connection to sending N new requests, this may be set to [`usize::MAX`] - N.
+    /// Remaining permits for sending new requests. Acquired via `fetch_sub(1)`; the counter is
+    /// alive while it remains positive. Poisoned by storing `0`; capped by keep-alive `max=N` via
+    /// `fetch_min(N)`.
+    permits: AtomicIsize,
+    /// Monotonic ID of the next request to send, used only for pipelining order.
     next_request_id: AtomicUsize,
     /// The ID of the next request which is readable from the socket. If we're pipelining this may
     /// be a few behind [`Self::next_request_id`]. If this is [`usize::MAX`], the socket is in an
@@ -300,6 +306,7 @@ impl AsyncConnection {
                 read,
             )),
             write: AsyncMutex::new(write),
+            permits: AtomicIsize::new(MAX_SEND_PERMITS),
             next_request_id: AtomicUsize::new(0),
             readable_request_id: AtomicUsize::new(0),
             min_dropped_reader_id: AtomicUsize::new(usize::MAX),
@@ -405,7 +412,7 @@ impl AsyncConnection {
                 let next_read = conn.readable_request_id.load(Ordering::Acquire);
                 let next_request = conn.next_request_id.load(Ordering::Acquire);
                 debug_assert!(
-                    next_request >= next_read,
+                    next_read == usize::MAX || next_request >= next_read,
                     "At all times, the next_request_id should be higher than the readable id"
                 );
             }
@@ -421,7 +428,7 @@ impl AsyncConnection {
             if !request.config.pipelining {
                 // If we're not pipelining, wait for any existing pipelined requests to complete.
                 // Specifically, wait until we have both locks and either we're going to build a
-                // new connection (because `next_request_id` is `usize::MAX`) or there are no
+                // new connection (because `permits` is exhausted/poisoned) or there are no
                 // pending readers (because `next_request_id` and `readable_request_id` are the
                 // same).
                 read = Some(Self::timeout(request.timeout_at, conn.read.lock()).await?);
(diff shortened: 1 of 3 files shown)


bitreq/src/connection.rs
```diff
@@ -474,7 +480,16 @@ impl AsyncConnection {
                 let socket_timeout = *conn.socket_new_requests_timeout.lock().unwrap();
                 let socket_timed_out = Instant::now() > socket_timeout;
 
-                request_id = conn.next_request_id.fetch_add(1, Ordering::Relaxed);
+                request_id = conn
+                    .next_request_id
```
TheBlueMatt: what kind of weird platform are you running on that you overflowed this? ISTM the right solution is just make it an AtomicU64 for the broken machines trying to use this on 32-bit.

stevenroose (approved):

bitreq/src/connection.rs
ErikDeSmedt: I am using `amd64`. I don't see how this can be platform related.

When some failures occur `conn.next_request_id `is set to `Usize::MAX`.
I understand the intent of setting `usize::Max` is to stop using this connection.

However, `AtomicUsize` overflows on any platform according to the [docs](https://doc.rust-lang.org/std/sync/atomic/struct.AtomicUsize.html#method.fetch_add).

TheBlueMatt: Oh sorry I misunderstood that this was counting past to overflow, forgot it was writing MAX elsewhere.

bitreq/src/connection.rs
```diff
@@ -474,7 +480,16 @@ impl AsyncConnection {
...
+                    .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
```
TheBlueMatt: Meh, wouldn't this be simpler by just changing the sentinel to `usize::MAX / 2` and accepting the race isn't going to manage to run through `usize::MAX / 2` elements before the first thread can get into `retry_new_connection` and re-write the sentinel value? Avoids the read+CAS in the optimistic case.

ErikDeSmedt:
I added a test to reproduce some of the issues I observed.
I created  a server that only allows one request per connection using a `Keep-Alive: max=1`-header.

The old code set `next_request_id` to `usize::Max - 1` to indicate that we can only make one request on that connection.
I often ended up in an infinite [loop](https://github.com/rust-bitcoin/corepc/blob/92c44b3e2ad282f691a181ece965f02db7323bfa/bitreq/src/connection.rs#L512-L525) where we kept waiting for `readable_request_id` to catch-up to this very high value.

I switched the approach a bit.

The `next_request_id` is now just a counter that increments on every request.
I introduced `permits` which is the number of requests that we are still allowed to make.

bitreq/src/connection.rs
```diff
@@ -214,19 +217,26 @@ impl AsyncWrite for AsyncHttpStream {
...
     next_request_id: AtomicUsize,
-    /// The ID of the next request which is readable from the socket. If we're pipelining this may
-    /// be a few behind [`Self::next_request_id`]. If this is [`usize::MAX`], the socket is in an
-    /// indeterminate state and no further reading is allowed. Any pending requests must either be
-    /// retried or failed.
+    /// ID of the next request readable from the socket; may lag [`Self::next_request_id`].
     readable_request_id: AtomicUsize,
+    /// Set when the read half is in an indeterminate state and pending pipelined readers must
+    /// abort and retry on a fresh connection.
+    reads_dead: AtomicBool,
```
TheBlueMatt: Hmm, why should we not just keep using `readable_request_id`? I'm not sure "more atomics" is really more readable since it now means we have to carefully review for cases where the atomic reads splitbrain.

bitreq/src/connection.rs
```diff
@@ -214,19 +217,26 @@ impl AsyncWrite for AsyncHttpStream {
...
     write: AsyncMutex<WriteHalf<AsyncHttpStream>>,
-    read: AsyncMutex<ReadHalf<AsyncHttpStream>>,
-    /// The ID of the next request we'll send. If this reaches [`usize::MAX`] no further requests
-    /// can be sent on this socket and a new connection must be made. Thus, in order to limit the
-    /// connection to sending N new requests, this may be set to [`usize::MAX`] - N.
+    read: AsyncMutex<tokio::io::BufReader<ReadHalf<AsyncHttpStream>>>,
+    /// Remaining permits for sending new requests. Acquired via `fetch_sub(1)`; an underflow is
+    /// detected as `prev == 0 || prev > MAX_SEND_PERMITS`. Poisoned by storing `0`; capped by
+    /// keep-alive `max=N` via `fetch_min(N)`.
+    permits: AtomicUsize,
```
TheBlueMatt: This would probably be more readable as an `AtomicIsize` where we can just check against 0.

ErikDeSmedt: Thanks. That is  a very good suggestion

bitreq/src/connection.rs
ErikDeSmedt: You are right. It doesn't help readability.

storopoli:
@ErikDeSmedt don't know if helps but we've seen this as well (https://github.com/alpenlabs/alpen/pull/1832#issuecomment-4492569702) and our short-term solution was to disable `debug_assertions` in `bitreq`. You can place this in your `Cargo.toml`:

```toml
# Disable debug_assertions in bitreq to dodge a known concurrency race in its
# `next_request_id >= readable_request_id` invariant, which fires under
# concurrent bitcoind RPC traffic in coverage/debug builds. Tracked upstream at
# https://github.com/rust-bitcoin/corepc/issues/583 (PR #584 unmerged).
[profile.dev.package.bitreq]
debug-assertions = false
```

ErikDeSmedt:
We have a few tests which sometimes just get stuck. The task just keeps running forever.
I have a hunch that this might be related. Is this something you have observed as well?

storopoli:
> We have a few tests which sometimes just get stuck. The task just keeps running forever. I have a hunch that this might be related. Is this something you have observed as well?

No all my recent tests are passing both unit and functional. Strange...

EDIT: with of course having to disable `debug_assertions` in `bitreq`

TheBlueMatt (approved):
Gah, github didn't send me an email about the push only your coments.

ACK 11098da75459773193132546c265e6740df136ea

TheBlueMatt:
@tcharding we should probably cut a release for this one.

tcharding:
Can do. @ErikDeSmedt do you mind running the formatter pls?

ErikDeSmedt:
@tcharding : I ran the formatter

tcharding:
> @tcharding we should probably cut a release for this one.

Bother, missed it in the last releases. Looks like we can do this in a point release, nice. @jamillambert after I merge this can you do the release tracking PR for that please (just bitreq).

tcharding:
Sorry @TheBlueMatt I don't have  commit hash 11098da75459773193132546c265e6740df136ea as a known revision so I can't range-diff againts your previous ACK to see that the force push was formatting only. Can you re-ack please.

TheBlueMatt (approved):
Please do not rebase a PR unless it actually conflicts. range-diff is great but its much nicer to be able to diff-tree.

Anyway, ACK 86a5f885eb005a18af1ca34bb6fab23879ce345f

PR #585, opened 2026-05-06
bitcoind: call wait in drop to reap process

nyonson:
Over in [rust-psbt we are seeing integration test failures](https://git.rust-bitcoin.org/rust-bitcoin/rust-psbt/actions/runs/58/jobs/8/attempt/1) which result in a CI runner spinning for hours, even though the test failed in a minute. Setting aside why the test failed (looks bitcoind related but dunno what), I am wondering if we should have a stronger guarantee in the drop to clean up the process. I looked back in the [old archive for the original impl reasoning](https://github.com/rust-bitcoin/bitcoind/pull/52) and don't see much.

I could also see this being a big no-no in a Drop impl, but for a high level thing like the bitcoind process maybe better than not?

### DIFF
--- a/bitcoind/src/lib.rs
+++ b/bitcoind/src/lib.rs
@@ -573,10 +573,13 @@ impl BitcoinD {
 
 impl Drop for BitcoinD {
     fn drop(&mut self) {
+        // Frist attempt graceful shutdown for persistent directories,
+        // always fallback to force kill and wait for process to be reaped.
         if let DataDir::Persistent(_) = self.work_dir {
             let _ = self.stop();
         }
         let _ = self.process.kill();
+        let _ = self.process.wait();
     }
 }
 


RCasatta:
Current code guarantees that failed stop RPC call will kill the process while in this PR it will not. I think adding wait could be fine like this:

```
  if let DataDir::Persistent(_) = self.work_dir {
      let _ = self.stop();
  }
  let _ = self.process.kill();
  let _ = self.process.wait();
```

nyonson:
Thanks @RCasatta, I totally missed that, updated in c08ea326303a79bfb0ab0e8918d8d97efc80462b

tcharding (approved):
ACK c08ea326303a79bfb0ab0e8918d8d97efc80462b

tcharding:
Hey @jamillambert lets do a point  release to get this out.

PR #589, opened 2026-05-13
ci: enable debug logging on electrsd jobs

satsfy:
Two CI tests for electrsd (`test-electrs-esplora` and `test-electrs-no-download`) run without `RUST_LOG=debug` set, so any spurious electrsd failure is hard to diagnose from the action logs.

This PR sets `RUST_LOG=debug` uniformly across all electrsd jobs. The only CI change is that it logs more meaningful information on errors.

Note that the third electrsd CI job `test-electrs` has `RUST_LOG: debug` set, meaning the CI job's config is presently inconsistent.

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -192,6 +192,7 @@ jobs:
     runs-on: ubuntu-24.04
     env:
       CARGO_TERM_COLOR: always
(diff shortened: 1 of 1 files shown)


tcharding (approved):
ACK 9abdb93639ef8da82d1847a86c4bcec947b644f5

Issue #457, opened 2026-01-15
Audit docs in `model`


tcharding:
Docs in the `model` module often are copied from core. As such they often mention the unit but `rust-bitcoin` types do not have a unit e.g., `Amount` is not in BTC or sats, it just is.

Audit the whole module.

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/457

PR #592, opened 2026-05-15
types: fix getblockstats doc for v21 and later

satsfy:
Docs are wrong in getblockstats txs field. 

The txs field counts all transactions in the block, including the coinbase. Core's source passes `block.vtx.size()` directly, which includes the coinbase transaction at index 0. [Bitcoin Core doc agrees](https://github.com/bitcoin/bitcoin/blob/ddb94fd3e1061c465e425629d6c7d4b292301da5/src/rpc/blockchain.cpp#L2025).

Example, I just called
```fish
bitcoin-cli -regtest generatetoaddress 101 (bitcoin-cli -regtest getnewaddress)
``` 

Now I run
```fish
bitcoin-cli -regtest getblockstats 50 | grep "\"txs\""
  "txs": 1,
``` 

Returns 1 because that is the coinbase transaction.

Also, `swtotal_weight` and `total_weight` fields say "divided by segwit scale factor (4)" but it was dropped from the Core docs at v21 and should not appear in v21 or later.


jamillambert (commented):
I checked the docs and it looks like this changed in v20, i.e. v17-19 say (excluding coinbase). I didn't check if it actually changed or if the docs were fixed. I say this because I know of other cases where the docs in v17 or so were incorrect.

satsfy:
Great catch, my bad for letting that slip

So in version v20 [it says excluding](https://bitcoincore.org/en/doc/0.20.0/rpc/blockchain/getblockstats/).
In version v21 [it says including](https://bitcoincore.org/en/doc/0.21.0/rpc/blockchain/getblockstats/).

Also fix swtotal_weight and total_weight in and later because the phrase "divided by segwit scale factor (4)" was dropped from the Core docs at v21 and should not appear in v21+.

jamillambert:
> So in version v20 [it says excluding](https://bitcoincore.org/en/doc/0.20.0/rpc/blockchain/getblockstats/). In version v21 [it says including](https://bitcoincore.org/en/doc/0.21.0/rpc/blockchain/getblockstats/).

I just checked again to see why we get a different version where this changed and it changed between 20.1 and 20.2. We use 20.2 in types.

tcharding:
I read the PR description and it reads like this is a docs only fix but then the patch diff is big and there are loads of code changes. Is this PR introducing new types _just_ to fix the docs? If so this is a policy change for the repo and cannot be done lightly. Looks like this is tackling the problems outline in #457 and #59 but only partially. If so we need a more principled approach.

satsfy:
Closing the PR because it reads like a NACK for this specific context.

The better approach for this, I suppose, will come with the codegen production client I'm working on, which would solve all of these problems (down to v24 of Bitcoin Core) with 100% assured correctness for documentation problems like this.

For the testing sync client, chunking them up like this PR is not ideal, per review. Moving forward, I'll only open a PR if I have a substantial amount of changes like this to justify.

PR #593, opened 2026-05-19
electrsd: set default features for bitreq

tankyleo:
If and only if the download feature is set, bitreq is required. In that case the bitreq std feature is needed, so we enable default-features for bitreq to set the std feature. No other bitreq feature is enabled.

### DIFF
--- a/electrsd/Cargo.toml
+++ b/electrsd/Cargo.toml
@@ -29,9 +29,7 @@ zip = { version = "0.6", default-features = false, optional = true, features = [
   "bzip2",
   "deflate",
 ] }
-bitreq = { version = "0.3.5", path = "../bitreq", default-features = false, optional = true, features = [
-  "https",
-] }
+bitreq = { version = "0.3.5", path = "../bitreq", optional = true, features = ["https"] }
 
 [features]
 legacy = []


tankyleo:
cc @tnull

jamillambert (commented):
The title and change don't match. I think the title is correct and the `bitreq/std` feature should only be enabled when download is by removing `std` from `bitreq` and adding instead to the `download` fearture:
`download = ["bitcoin_hashes", "zip", "bitreq/std"]`

tankyleo:
> The title and change don't match. I think the title is correct and the `bitreq/std` feature should only be enabled when download is by removing `std` from `bitreq` and adding instead to the `download` fearture: `download = ["bitcoin_hashes", "zip", "bitreq/std"]`

If we do this, then this applies to the `https` feature too right? So we would do: `download = ["bitcoin_hashes", "zip", "bitreq/std", "bitreq/https"]`

I would find this uglier, especially since only `download` uses bitreq, but let me know.

jamillambert:
> > The title and change don't match. I think the title is correct and the `bitreq/std` feature should only be enabled when download is by removing `std` from `bitreq` and adding instead to the `download` fearture: `download = ["bitcoin_hashes", "zip", "bitreq/std"]`
> 
> If we do this, then this applies to the `https` feature too right? So we would do: `download = ["bitcoin_hashes", "zip", "bitreq/std", "bitreq/https"]`
> 
> I would find this uglier, especially since only `download` uses bitreq, but let me know.

Hmm, yeah, you could change the tile so it is correct instead.  Removing `default-features = false` from `bitreq` also fixes the problem since the only default feature is `std`.

electrsd/Cargo.toml
```diff
@@ -31,6 +31,7 @@ zip = { version = "0.6", default-features = false, optional = true, features = [
 ] }
 bitreq = { version = "0.3.5", path = "../bitreq", default-features = false, optional = true, features = [
   "https",
+  "std",
```
tcharding: While correct I also rekon this would be better:

```toml
bitreq = { version = "0.3.5", path = "../bitreq", optional = true, features = ["https"] }
```

tcharding:
Thanks Leo, I debugged this myself today and got lost wondering how the heck this got into master.

electrsd/Cargo.toml
tankyleo: Thanks for the review, done below

jamillambert (approved):
ACK adde75f1417701ca49f4068c20baf0f17713b2d0

tcharding (approved):
ACK adde75f1417701ca49f4068c20baf0f17713b2d0

PR #595, opened 2026-05-19
initializing the default crypto provider for electrsd

ulrichard:
When I want to use electrsd v39 in bdk-reserves (https://github.com/bitcoindevkit/bdk-reserves/pull/42), I ran into errors about the rustls default crypto provider not being initialized.

### DIFF
--- a/bitcoind/build.rs
+++ b/bitcoind/build.rs
@@ -14,6 +14,7 @@ mod download {
 
     use anyhow::Context;
     use bitcoin_hashes::{sha256, Hash};
+    use bitreq::rustls::crypto::{ring, CryptoProvider};
     use flate2::read::GzDecoder;
     use tar::Archive;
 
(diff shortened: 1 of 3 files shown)


bitcoind/build.rs
```diff
@@ -80,6 +81,9 @@ mod download {
         }
         let out_dir = PathBuf::from(std::env::var_os("OUT_DIR").unwrap());
 
+        // specify the default crypto provider for rustls
+        let _ = CryptoProvider::install_default(ring::default_provider());
```
tnull, line 85: Note that this will likely set the default for `rustls` workspace-wide, i.e., potentially overriding any choice any other crate makes. I don't think we should call this manually.

ulrichard:
Strangely, after I did those changes to corepc, bdk-reserves showed some other errors. And after resolving the new errors, the changes here are no longer necessary.

PR #605, opened 2026-05-24
Change `fee_delta` to `SignedAmount`

jamillambert:
`fee_delta` can be negative.

Change it from `Amount` to `SignedAmount` and update the test to confirm that a negative value is accepted.

### DIFF
--- a/integration_test/tests/mining.rs
+++ b/integration_test/tests/mining.rs
@@ -68,8 +68,28 @@ fn mining__get_network_hash_ps() {
 fn mining__get_prioritised_transactions() {
     let node = BitcoinD::with_wallet(Wallet::Default, &[]);
     node.fund_wallet();
+    let (_address, txid) = node.create_mempool_transaction();
 
-    let _ = node.client.get_prioritised_transactions().expect("getprioritisedtransactions");
+    let fee_delta = SignedAmount::from_sat(-10_000);
+    node.client.prioritise_transaction(&txid, fee_delta).expect("prioritisetransaction");
+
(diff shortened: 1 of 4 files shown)


tcharding (approved):
ACK 783247f5176c2e4c6f854a5e884e2cd69337b256

integration_test/tests/mining.rs
```diff
@@ -68,8 +68,24 @@ fn mining__get_network_hash_ps() {
 fn mining__get_prioritised_transactions() {
     let node = BitcoinD::with_wallet(Wallet::Default, &[]);
     node.fund_wallet();
-
-    let _ = node.client.get_prioritised_transactions().expect("getprioritisedtransactions");
+    let (_address, txid) = node.create_mempool_transaction();
+    let transaction: GetTransaction = node.client.get_transaction(txid).expect("gettransaction");
+    let transaction: Result<mtype::GetTransaction, GetTransactionError> = transaction.into_model();
+    let transaction = transaction.unwrap();
```
tcharding: nit: We don't need this explicit type stuff because its tested elsewhere (I assume). Its not a big deal but i rekon we should remove it because its confusing for devs later on and this crate is hard enough already.

jamillambert: Removed. Also fixed the formatting issue and added a required version feature gate to the test since `modified_fee` only exists from v27 and higher.

tcharding (approved):
ACK 9bcbdb3fab995876ede6f95bd6d8c50ca4eb0291

PR #606, opened 2026-05-26
tests: add curated list of client test invariants from Bitcoin Core

### LINKED ISSUE #58: Test using Core's test cases
Currently we have only limited test coverage. And for the RPC methods that are tested we only test a subset of things e.g.,

- We don't test all optional fields
- We don't ensure all fields are non-default e.g., they may be being filled in by `serde` and not actually deserialized.

An idea, courtesy of @stevenroose, is to grab the tests from Core used for testing the RPC code and then try to parse all of those.

satsfy:
Closes #58.

This PR adds 79 tests corresponding to invariants picked after analyzing Bitcoin Core tests from `test/functional` folder. Identical test copies from core proved unproductive, extensive, and tested much more than necessary for corepc, so invariants (not tests) needed to be filtered. By putting these tests in `tests/`, there is no additional effort on `Cargo.toml`.

The criteria:
- Not a test of Core's behavior.
- Only the client layer is tested (meaning the tests apply to the state before the into_model conversion).
- If a particular assertion failed, that means corepc has a bug.
- The tests are complementary and do not repeat or collide with existing corepc integration tests.

Main targets:
- Optional fields to prevent absence failures that the into_model conversions failed to catch.
- Round trips not exercised or already tested.
- Optional argument paths.
- Fragile serde renames.
- Consistency across fields (such as  `banned_until == ban_created + ban_duration`).
- Nested structs.
- Tests that prove the mutual sanity of RPCs, compressing simple RPC tests into one (such as getblockhash tip matches best block, get_deployment_info + get_deployment_info_tip).

All test check corepc honor a particular invariant. Previously untested fields are tested. Many tests contain `is_some()` because exact value assertions are not necessary, as long as the result parses into a particular type. Comprehensiveness was not a target because we test only what Core has. Flags are checked to be covering only the versions it should.

How LLM was used: I read through Core tests to see what we were missing. Started developing a philosophy for assertions to pick. Realized that copying tests directly wouldn't work, so we would need to pick invariants (some are hard to reproduce, need Core's particulars, or test no meaningful corepc behavior). Asked Claude Opus 4.7 to convert the entire Bitcoin Core surface, and started iterating on it, covering gaps where needed, shaving off useless assertions, and sharpening the philosophy of tests.

I have found various errors (such as rediscovering the inconsistency in numeric types) in RPCs/docs, saving them to investigate and open issues later.

### DIFF
--- a/integration_test/README.md
+++ b/integration_test/README.md
@@ -33,3 +33,8 @@ alias test28='BITCOIND_EXE=/opt/bitcoin-28.2/bin/bitcoind cargo test --features=
 alias test29='BITCOIND_EXE=/opt/bitcoin-29.0/bin/bitcoind cargo test --features=29_0'
 alias test30='BITCOIND_EXE=/opt/bitcoin-30.2/bin/bitcoind cargo test --features=30_2'
 ```
+
+## Bitcoin Core Tests
+
+Tests derived from Bitcoin Core's `test/functional/` folder live in
+`tests/` alongside the modelled tests, with a `_core` filename suffix.
--- a/integration_test/tests/blockchain_core.rs
+++ b/integration_test/tests/blockchain_core.rs
@@ -1,0 +1,421 @@
+// SPDX-License-Identifier: CC0-1.0
+
+//! Tests derived from Bitcoin Core rpc_blockchain.py, rpc_getchaintips.py,
+//! rpc_txoutproof.py, rpc_scantxoutset.py, rpc_gettxspendingprevout.py,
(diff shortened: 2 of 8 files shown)


tcharding:
Mad, bit sad that we didn't find any bugs ...

tcharding (approved):
ACK 6f78f2a0e8dab490d45f47d80a948f7aabae0901

tcharding:
Good effort man, this is a ton of work.

PR #610, opened 2026-05-26
Gate out core tests from v31

### LINKED ISSUE #609: v31 tests fail
There is a CI failure on master for v31 due to the new tests added in #606. The PR was reviewed before the v31 support was on master, but now with it and v31 support the tests fail.

jamillambert:
v31 does not have all of the RPC changes implemented yet which causes the newly added tests to fail.

Gate out the tests so they are not run in v31.

Closes #609


satsfy (approved):
tACK fd3709c - Locally master break, with this commit, v31 passes.

Good catch, should have rebased it before putting up the PR.

tcharding (approved):
ACK fd3709c7f6299af87af31b86a9c966ca051b84cf

Issue #60, opened 2025-02-12
Fee rate units


tcharding:
The Core RPC API is not consistent with the unit used for fee rates. I can't remember if it changes between versions or between methods but sometimes its `BTC/kB` and sometimes its `satoshis per virtual byte`. Fortunately these are different types (f64 and u64 respectively).

However, in the ecosystem, fee rates are also measured in sats per kilo weight unit. Both sats per vb and sats per kwu use `u64` - we should make some effort to ensure that Core does not use both and that we are not mixing these up.

Perhaps #58 will help with this if we are checking values returned.

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/60

PR #615, opened 2026-05-29
Finish off support for Bitcoin Core v31.0

### LINKED ISSUE #599: Finish off support for Core v31
If #598 merges/releases we can then do this.

Go through all the `TODO`s added in `types/src/v31/mod.rs` and remove them. For each one there will be one or more tests in `integration_test` feature gated out with `v30_and_below`. 

After this is done the only instance left of `v30_and_below` should be on `wallet__set_tx_fee`.

Also add tests for the three new methods added in v31:

- `getmempoolcluster`
-  `abortprivatebroadcast`
-  `getprivatebroadcastinfo`

This issue can best be done by a ton of small PRs IMO.

satsfy:
Closes https://github.com/rust-bitcoin/corepc/issues/599

Completes v31 support across the client, types, and integration tests. Makes v31 the default version.

- set v31 as the latest version and re-export the v31 types.
- enabled the v31 integration tests.
- Dropped the incorrect "fields changed" doc notes on several v31 types.

Fixed:
  - getpeerinfo (network): added `inv_to_send` and `last_inv_sequence`, and made `startingheight` optional now that it is only returned under `-deprecatedrpc=startingheight`.
  - logging (control): added the new `kernel` and `privatebroadcast` categories.
  - getwalletinfo (wallet): no longer returns `paytxfee`.
  - getdeploymentinfo (blockchain): added the `script_flags` field.
  - gettxspendingprevout (blockchain): fixed the result type to carry the new `spendingtx` and `blockhash` fields, plus added support for the new `mempool_only` and `return_spending_tx` arguments.
  - getblock (blockchain): added the `coinbase_tx` object returned at verbosity 1, 2 and 3.

Added:
- cluster mempool (blockchain): new `getmempoolcluster` and `getmempoolfeeratediagram` RPCs, the `chunk_weight` and `chunk_fee` fields on `getmempoolentry`, `getrawmempool`, `getmempoolancestors` and `getmempooldescendants`, and the `cluster_limit` and `optimal` fields on `getmempoolinfo`.
- private broadcast (raw transactions): new `getprivatebroadcastinfo` and `abortprivatebroadcast` RPCs.

LLM usage: Claude helped on generating tests for new RPCs, verifying things over and over again, attempting to call every RPC, spotting additional unfixed things, and reviewing every change the PR.

When appropriate, comments have been checked to match v31 documentation exactly.


satsfy:
Latest force push uses enums for getpeerinfo types, maintaining full v31 docs in corepc (before, it simply discarded relevant messages from Bitcoin Core documentation).

EDIT: below i retriggered CI

types/src/model/blockchain.rs
```diff
@@ -777,7 +810,7 @@ pub struct GetMempoolInfo {
     pub total_fee: Option<f64>,
     /// Maximum memory usage for the mempool.
     pub max_mempool: u32,
-    /// Minimum fee rate in BTC/kB for a transaction to be accepted.
+    /// Minimum fee rate in BTC/kvB for a transaction to be accepted.
```
tcharding: This change shouldn't be in here because we should not be mentioning the rate in `model` because the `FeeRate` abstracts this. Can you either just remove this change or throw an additional patch at the front that removes these sorts of docs everywhere in `model`. 

Related: #60

types/src/v19/blockchain/into.rs
```diff
@@ -222,6 +224,9 @@ impl GetMempoolInfo {
         let min_relay_tx_fee = crate::btc_per_kb(self.min_relay_tx_fee)?;
 
         Ok(model::GetMempoolInfo {
+            limit_cluster_count: None,
+            limit_cluster_size: None,
+            optimal: None,
```
tcharding: While not technically wrong, I'm a bit anal about code cleanlyness. Can these go down in the correct order pls?

tcharding: Same for all the places you add `foo: None` pls.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +47,604 @@ impl GetMempoolCluster {
...
+    /// Conversion of the `descendant` field failed.
+    Descendant(ParseAmountError),
+    /// Conversion of the `chunk` field failed.
+    Chunk(ParseAmountError),
+}
+
+impl fmt::Display for MempoolEntryFeesError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Base(e) => write!(f, "conversion of the `base` field failed: {}", e),
```
tcharding: In this (and `rust-bitcoin`) repos we use the style of matching on `*self` and then pattern matching on `(ref e)`. The reason is its just a bit more explicit to what the compiler would otherwise do for us. It was a style choice made a long time ago, to favour explicit over implicit.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +47,604 @@ impl GetMempoolCluster {
...
+}
+
+impl fmt::Display for MempoolEntryError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            Self::Numeric(e) => write!(f, "numeric: {}", e),
+            Self::Wtxid(e) => write!(f, "conversion of the `wtxid` field failed: {}", e),
+            Self::Fees(e) => write!(f, "conversion of the `fees` field failed: {}", e),
+            Self::Depends(e) => write!(f, "conversion of the `depends` field failed: {}", e),
+            Self::SpentBy(e) => write!(f, "conversion of the `spentby` field failed: {}", e),
```
tcharding: These should be using `write_err` like we do elsewhere.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +47,604 @@ impl GetMempoolCluster {
...
+            Self::Ancestor(e) => Some(e),
+            Self::Descendant(e) => Some(e),
+            Self::Chunk(e) => Some(e),
+        }
+    }
+}
+
+/// Error when converting a `MempoolEntry` type into the model type.
+#[derive(Debug)]
+pub enum MempoolEntryError {
```
tcharding: All these error types go in a separate file called `error.rs`.

types/src/v31/blockchain/mod.rs
```diff
@@ -26,15 +34,409 @@ pub struct GetMempoolCluster {
...
     /// Fees of the transactions in this chunk.
     #[serde(rename = "chunkfee")]
-    // FIXME: This is probably a float. https://github.com/rust-bitcoin/corepc/pull/598#discussion_r3283795812
-    pub chunk_fee: u64,
+    pub chunk_fee: f64,
     /// Sigops-adjusted weight of all transactions in this chunk.
     #[serde(rename = "chunkweight")]
     pub chunk_weight: u64,
-    /// Transactions in this chunk in mining order (txids).
+    /// Transactions in this chunk in mining order.
```
tcharding, line 51: Why this change?

tcharding:
Looks pretty good man. To create all the new types you just copied the one from the latest version and added the fields, right? I.e for `MempoolEntry` you copied the one from `v24`, right?

tcharding:
Reviewed: 1d584d8c78b55bfceddfcd6c8b5abce795f178bd

types/src/v31/blockchain/mod.rs
satsfy: In Bitcoin Core, it looks like this 
```cpp
RPCResult{RPCResult::Type::ARR, "txs", "transactions in this chunk in mining order",
    {RPCResult{RPCResult::Type::STR_HEX, "txid", "transaction 
```
And in [the docs](https://bitcoincore.org/en/doc/31.0.0/rpc/blockchain/getmempoolcluster/)
```
"txs" : [             (json array) transactions in this chunk in mining order
  "hex",              (string) transaction id
  ...
]
```
There is no `(txids)`. Whenever this kind of conflict happens (knowing that our existing docs have errors sometimes) I took no chances and copied exactly what core said to prevent drift. I thought this would be the way to go because we will have codegen types which effortlessly copy precisely the same docs as core.

tcharding: Yep nice, spot on. Thanks

satsfy:
> Looks pretty good man. To create all the new types you just copied the one from the latest version and added the fields, right? I.e for `MempoolEntry` you copied the one from `v24`, right?

Yes. Everything that didn't need to change hasn't changed. Diffs only in new/modified fields or the corresponding docs. All "institutional memory" is kept.

jamillambert (commented):
I've had a look through and it all looks good so far. There is a lot here at so will need to have another look tomorrow.

types/src/model/blockchain.rs
satsfy: The issue backlink already captures this follow-up, so I'll remove the mention of "BTC/kB". File `types/src/v31/blockchain/mod.rs` still contains the "BTC/kvB" because Core has in docs.

types/src/v19/blockchain/into.rs
satsfy: Yes, my oversight. Will do.

types/src/v31/blockchain/into.rs
satsfy: Change done, I changed the pattern above into the corepc pattern.

types/src/v31/blockchain/into.rs
satsfy: Refactored the file to match corepc.

types/src/v31/blockchain/into.rs
satsfy: Done

tcharding:
The first commit doesn't build. Can you get the PR into a state where I can review patch by patch and on each I can run `test30` and `test31` cleanly (in `integration_test/`) please? See the README for these commands if you haven't already.

satsfy:
Force push fixed the bissectability of every commit.

EDIT: no, a90eb319 is broken.

apoelstra:
150016fb2657afc1aaa1baa0751ab20d22996e8f needs rebase

satsfy:
Force push fixes the other bad force pushes I made. Should be good to go.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +46,395 @@ impl GetMempoolCluster {
...
+            difficulty: self.difficulty,
+            chain_work,
+            n_tx: crate::to_u32(self.n_tx, "n_tx")?,
+            previous_block_hash,
+            next_block_hash,
+        })
+    }
+}
+
+impl GetBlockVerboseThree {
```
jamillambert, line 411: This doesn't seem right. It should do the parsing of each field in the into_model function like is done for VerboseTwo.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +46,395 @@ impl GetMempoolCluster {
...
+        Ok(model::GetMempoolFeerateDiagram(entries))
+    }
+}
+
+impl GetDeploymentInfo {
+    /// Converts version specific type to a version nonspecific, more strongly typed type.
+    pub fn into_model(
+        self,
+    ) -> Result<model::GetDeploymentInfo, crate::v23::GetDeploymentInfoError> {
+        use crate::v23::GetDeploymentInfoError as E;
```
jamillambert:
```suggestion
        use GetDeploymentInfoError as E;
```

Errors from previous versions should be reexported from the module, i.e. in this case `types/src/v30/blockchain/mod.rs` and then added to the `use super` at the top of the into.rs file.

types/src/v31/mod.rs
```diff
@@ -28,30 +28,30 @@
 //! |:-----------------------------------|:---------------:|:--------------------------------------:|
 //! | dumptxoutset                       | version + model |                                        |
 //! | getbestblockhash                   | version + model |                                        |
-//! | getblock                           | version + model | TODO: Fields have changed since v30    |
+//! | getblock                           | version + model | Includes 'coinbase_tx' object          |
```
jamillambert:
```suggestion
//! | getblock                           | version + model | Includes additional 'verbose' type     |
```

These comments in the table are saying there are additional types, i.e. more than just the `GetBlock` type, not what fields were changed in the version. I didn't check the rest of the RPCs in this table but any that have a verbose type should have this comment and all the other comments about what changed removed.

types/src/v31/blockchain/into.rs
satsfy: Yes my mistake. Has been fixed.

types/src/v31/blockchain/into.rs
satsfy: Everything used from previous versions is now exported from the module.

types/src/v31/mod.rs
satsfy: Done

satsfy:
Latest force push:
- Made `GetRawTransactionVerboseWithPrevout::into_model_with_prevouts` public in v29 so v31 can call it.
- Refactored the `into_model` conversions to follow corepc pattern.
- Routed any call that reach back into older version though `mod.rs`.
- `GetBlockVerboseThree::into_model` now builds correctly.
- Cleaned up the notes column in the v31 method table.

satsfy:
Clean up accidental garbage file

jamillambert (approved):
ACK daf4571be4b564dd86a3f1c0cd370622d19c5143

Nice one

integration_test/tests/control.rs
```diff
@@ -27,10 +27,12 @@ fn control__help() {
 }
 
 #[test]
-#[cfg(feature = "v30_and_below")]
 fn control__logging() {
     let node = BitcoinD::with_wallet(Wallet::None, &[]);
-    let _: Logging = node.client.logging().unwrap();
+    let logging: Logging = node.client.logging().unwrap();
+    #[cfg(not(feature = "v30_and_below"))]
+    let _ = (logging.kernel, logging.privatebroadcast);
```
tcharding, line 34: What is the intention of this line? It doesn't prove anything, unless I'm missing something.

types/src/v31/wallet/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+//! The JSON-RPC API for Bitcoin Core `v31` - wallet.
+//!
+//! Types for methods found under the `== Wallet ==` section of the API docs.
+
+mod into;
+
+use serde::{Deserialize, Serialize};
+
+pub use super::{GetWalletInfoError, GetWalletInfoScanning, LastProcessedBlock};
```
tcharding, line 11: This works but it will break if we later change these types. Better to import them from `v30` I rekon.

types/src/v31/wallet/into.rs
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: CC0-1.0
+
+use super::{GetWalletInfo, GetWalletInfoError, GetWalletInfoScanning};
```
tcharding, line 3: While this works I'd prefer a new error type otherwise the 

```rust
    /// Conversion of the `pay_tx_fee` field failed.
    PayTxFee(ParseAmountError),
```
variant is never used.

types/src/v31/blockchain/mod.rs
```diff
@@ -4,10 +4,23 @@
...
 use serde::{Deserialize, Serialize};
 
+pub use self::error::{
+    GetMempoolClusterError, GetMempoolFeerateDiagramError, GetTxSpendingPrevoutError,
+    MapMempoolEntryError, MempoolEntryError, MempoolEntryFeesError,
+};
+use super::{DeploymentInfo, GetBlockVerboseThreeTransaction, GetBlockVerboseTwoTransaction};
+pub use super::{
+    GetBlockVerboseOneError, GetBlockVerboseThreeError, GetBlockVerboseTwoError,
+    GetDeploymentInfoError, GetMempoolInfoError,
```
tcharding, line 21: I didn't look closely but like mentioned elsewhere if these have unused variants then we want new error types. If not import directly from the version specific module.

tcharding: FTR I'm not totally sure we have this correct in other places in the crate. I may have been slack reviewing before and let them in. Thats the nature of this crate unfortunately, PRs are  often massive.

types/src/model/blockchain.rs
```diff
@@ -42,6 +42,23 @@ pub struct GetBestBlockHash(pub BlockHash);
...
+///
+/// Introduced in Bitcoin Core v31.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct CoinbaseTransaction {
+    /// The coinbase transaction version.
+    pub version: i32,
+    /// The coinbase transaction's locktime.
+    pub locktime: u32,
+    /// The coinbase input's sequence number (nSequence).
+    pub sequence: u32,
```
tcharding, line 55: These three can all be concrete types from `rusti-bitcoin`.

types/src/model/blockchain.rs
```diff
@@ -42,6 +42,23 @@ pub struct GetBestBlockHash(pub BlockHash);
...
+    /// The coinbase input's script.
+    pub coinbase: String,
+    /// The coinbase input's first (and only) witness stack element, if present.
+    pub witness: Option<String>,
```
tcharding, line 59: Maybe this one too, didn't look closely.

types/src/model/blockchain.rs
```diff
@@ -42,6 +42,23 @@ pub struct GetBestBlockHash(pub BlockHash);
 #[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
 pub struct GetBlockVerboseZero(pub Block);
 
+/// The coinbase transaction of a block. Part of `getblock` at verbosity 1, 2 and 3.
+///
+/// Introduced in Bitcoin Core v31.
+#[derive(Clone, Debug, PartialEq, Eq, Deserialize, Serialize)]
+pub struct CoinbaseTransaction {
```
tcharding, line 49: Can you put this below where its used please. I always favour this style based on the idea to put the most important stuff first in a file.

types/src/v31/blockchain/into.rs
```diff
@@ -27,3 +44,432 @@ impl GetMempoolCluster {
...
+    pub fn into_model(self) -> model::CoinbaseTransaction {
+        model::CoinbaseTransaction {
+            version: self.version,
+            locktime: self.locktime,
+            sequence: self.sequence,
+            coinbase: self.coinbase,
+            witness: self.witness,
+        }
+    }
+}
```
tcharding, line 281: For your learning, this code hints that the type is wrong. There is not need to have an `into_model` function for a type that does no type conversions.

types/src/model/blockchain.rs
tcharding: Oh I see in other places we don't do this. If a type is used in a bunch of places we do tend to throw it up top of the file. Just do what you like here.

client/src/client_sync/v31/blockchain.rs
```diff
@@ -20,3 +20,42 @@ macro_rules! impl_client_v31__get_mempool_cluster {
...
+#[macro_export]
+macro_rules! impl_client_v31__get_tx_spending_prevout {
+    () => {
+        impl Client {
+            pub fn get_tx_spending_prevout(
+                &self,
+                outputs: &[bitcoin::OutPoint],
+                mempool_only: bool,
+                return_spending_tx: bool,
+            ) -> Result<GetTxSpendingPrevout> {
```
tcharding, line 46: Typically I'd try to avoid boolean args like this. Is there some precedence in the repo already for doing this? 

Another option would be to just always pass these args in. I did not look too closely though.

tcharding (approved):
ACK daf4571be4b564dd86a3f1c0cd370622d19c5143

tcharding:
I'm going to merge this as is because its big. Can you see to review suggestions as a follow up please?

tcharding:
After the follow up PR we can do a release for this stuff. Well done man!

integration_test/tests/control.rs
satsfy: I see my bad, the test is pointless. I was trying to enforce field presence so that I get errors like this:
```bash
❯ cargo test --features 31_0,download
error[E0609]: no field `kernel` on type `bitcoind::vtype::Logging`
  --> tests/control.rs:34:22
34 |     let _ = (logging.kernel, logging.privatebroadcast);
   |                      ^^^^^^ unknown field              
```
But the line `let _: Logging = node.client.logging().unwrap();` already does it. Removing.

client/src/client_sync/v31/blockchain.rs
satsfy: Oh, so this is entirely optional. I'm removing it.

PR #623, opened 2026-06-15
Simplify verify categories

### LINKED ISSUE #601: Investigate why 'returns boolean' and 'Method::new_nothing` co-exist
In #591 the changeset included 

```diff
-    Method::new_nothing("getnetworkhashps", "get_network_hashes_per_second"),
+    Method::new_numeric("getnetworkhashps", "get_network_hashes_per_second"),
```

And the method was marked `returns boolean` in `mod.rs` docs.

I would have thought `verify` would have caught this discrepancy (ignoring that it should have been float not a boolean).

jamillambert:
There are a few different verify categories for RPCs that do not have a version specific type, and therefore also no model or any required reexports. All of them are have the same requirements in types and verify cannot test what type is actually returned by the RPC, this is for testing, so they can all be in the same category.

- Fix some existing problems that the improved verify caught.
- Create a new category new_no_type and replace all of the others. And improve the verify error messages.

Closes #601


tcharding (approved):
ACK 3094e82861f271d5ac913afd28dbe8256093fb18

PR #632, opened 2026-06-17
Improve bitcoind multithreading

jamillambert:
I got timeouts and errors reading tmp files when running the tests with 16 threads. After this fix it runs fine.

Use kill + wait cleanup on all retry/error paths.

Fix base-client creation path so it no longer bails without cleanup and instead retries cleanly.

Fix final wallet-load failure path to always terminate/reap before returning.

Create consts for the retry and timeout values and increase them.

Assisted-by: GitHub Copilot


tcharding (approved):
ACK bc2ba6e6312ed2a92e894ef141a3dd486264c2da

tcharding:
FTR this would have been better as a bunch of patches.

jamillambert:
> FTR this would have been better as a bunch of patches.

Sorry my bad. I should have spent more time tidying up after the LLM. Feels like that's half my job now.

Issue #633, opened 2026-06-17
Untested RPCs


jamillambert:
There are some RPCs in types that are untested:
```
== Mining ==
submitblock

== Rawtransactions ==
descriptorprocesspsbt
signrawtransactionwithkey

== Wallet ==
setlabel
```

Found using the fix of verify by @RuslanProgrammer in #630

jamillambert:
Closing since all development has now shifted to forgejo.

PR #634, opened 2026-06-17
Fix all the issues with verify --tests

jamillambert:
There are a bunch of issues with test names and labels in verify that cause the fixed `verify` with `--tests` to fail.

Fix all of them, including the RPCs that are actually not tested by adding the label to the types table and also mentioning them in the new issue #633.


tcharding (approved):
ACK d64ada2ceb1a92eef74f4e7da07514c6958da33c

tcharding:
Nice one man. We should probably consider running verify with `--tests` in CI. I was too lazy initially.

jamillambert:
> Nice one man. We should probably consider running verify with `--tests` in CI. I was too lazy initially.

Yeah, it's quite a quick check. I wonder if it can be added to the end of the integration_test CI since that needs to be run first to generate the output file. But only run it if the test passes, otherwise you will get a whole list of missing tests for the ones that weren't run.

Opened #639

PR #644, opened 2026-06-22
Run `verify --tests` in CI

### LINKED ISSUE #639: Run `verify --tests` in CI?
This could be added to the end of the `Integration tests` CI job, which has to be run first to generate the output file for `verify`. It is a quick check so won't add much time to the CI run.

jamillambert:
Fix the two problems highlighted when adding `verify --tests` to CI:
`submitpackage` is untested for v26 - Add `UNTESTED` label to the types table.
`abortprivatebroadcast` test did not check the model - Fix the test.

Add `verify --tests` to CI:
For every version `integration_tests` capture the output and run `verify --tests` with the output generated. Run in the same job as the tests since they are needed to generate the file.

Closes #639

### DIFF
--- a/.github/workflows/rust.yaml
+++ b/.github/workflows/rust.yaml
@@ -160,7 +160,19 @@ jobs:
       - name: "Run integration tests"
         env:
           BITCOIND_DOWNLOAD_DIR: ${{ github.workspace }}/.cache/corepc/bitcoind
-        run: cd integration_test && cargo test --features=${{ matrix.version }},download
+        run: |
+          set -o pipefail
+          test_output="${{ runner.temp }}/integration-test-${{ matrix.version }}.out"
+          cd integration_test
+          cargo test --features=${{ matrix.version }},download 2>&1 | tee "$test_output"
+      - name: "Verify integration test coverage"
+        run: |
+          test_output="${{ runner.temp }}/integration-test-${{ matrix.version }}.out"
+          verify_version="${{ matrix.version }}"
+          verify_version="${verify_version#0_}"
+          verify_version="${verify_version%%_*}"
+          cd verify
+          cargo run -- --tests "$test_output" "$verify_version"
 
   # The jobs below copied from electrsd repo during import. Not too
   # much further thought given to it.
--- a/integration_test/tests/raw_transactions.rs
+++ b/integration_test/tests/raw_transactions.rs
@@ -717,10 +717,26 @@ fn raw_transactions__get_private_broadcast_info__modelled() {
 
 #[test]
 #[cfg(not(feature = "v30_and_below"))]
-fn raw_transactions__abort_private_broadcast() {
-    let node = BitcoinD::with_wallet(Wallet::None, &[]);
+fn raw_transactions__abort_private_broadcast__modelled() {
+    let node =
+        BitcoinD::with_wallet(Wallet::Default, &["-privatebroadcast=1", "-proxy=127.0.0.1:1"]);
+    node.fund_wallet();
 
-    // Aborting a transaction that is not in the private broadcast queue returns an error in regtest.
-    let txid = "0000000000000000000000000000000000000000000000000000000000000001";
-    assert!(node.client.abort_private_broadcast(txid).is_err());
+    // Create a signed transaction and send it via private broadcast.
+    let tx = create_a_raw_transaction(&node);
+    let signed = node
+        .client
+        .sign_raw_transaction_with_wallet(&tx)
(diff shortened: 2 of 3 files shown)


jamillambert:
I left in the existing `verify` job even though it's run again in integration_tests because it's fast and will fail quickly if there is another issue that verify catches. The additional check added by this PR is a check on the integration tests so fits within the title of that job.

.github/workflows/rust.yaml
```diff
@@ -160,7 +160,18 @@ jobs:
       - name: "Run integration tests"
         env:
           BITCOIND_DOWNLOAD_DIR: ${{ github.workspace }}/.cache/corepc/bitcoind
-        run: cd integration_test && cargo test --features=${{ matrix.version }},download
+        run: |
+          test_output="${{ runner.temp }}/integration-test-${{ matrix.version }}.out"
```
satsfy, line 165: I did a test [CI run](https://github.com/satsfy/corepc/actions/runs/28105273614/job/83217703897?pr=43) on my branch, failing the new integration test with `assert!(false)`. The test does not fail on the integration test; it keeps going to verify, which fails loudly because all subsequent outputs after the failed integration test were not produced and did not fail fast so they are all missing at verify time. The solution is simple:
```suggestion
          set -o pipefail
          test_output="${{ runner.temp }}/integration-test-${{ matrix.version }}.out"
```

satsfy (changes requested):
PR looking good aside from one CI difficulty.

.github/workflows/rust.yaml
jamillambert: Thanks, I added the fix.

satsfy (approved):
tACK da9f01788183fd261db2c82b4b52912187d30cd9

[Ran the CI again](https://github.com/satsfy/corepc/actions/runs/28109113746/job/83231337607), fails as expected now on the integration tests.

integration_test/tests/raw_transactions.rs
```diff
@@ -717,10 +717,25 @@ fn raw_transactions__get_private_broadcast_info__modelled() {
...
+        BitcoinD::with_wallet(Wallet::Default, &["-privatebroadcast=1", "-proxy=127.0.0.1:1"]);
+    node.fund_wallet();
+
+    // Create a signed transaction and send it via private broadcast.
+    let tx = create_a_raw_transaction(&node);
+    let signed: SignRawTransactionWithWallet =
+        node.client.sign_raw_transaction_with_wallet(&tx).expect("signrawtransactionwithwallet");
+    let signed_tx = signed.into_model().expect("SignRawTransaction into model").tx;
+    let send_raw_transaction: SendRawTransaction =
+        node.client.send_raw_transaction(&signed_tx).expect("sendrawtransaction");
```
tcharding: This is all good but the explicit types clutter the code. I rekon we should be careful to only use them when needed so that the reason we have them is _somewhat_ clear. Here we are testing the abort types so we only need to have them explicit.


Perhaps this is cleaner?

```rust
fn raw_transactions__abort_private_broadcast__modelled() {
    let node =
        BitcoinD::with_wallet(Wallet::Default, &["-privatebroadcast=1", "-proxy=127.0.0.1:1"]);
    node.fund_wallet();

    // Create a signed transaction and send it via private broadcast.
    let tx = create_a_raw_transaction(&node);
    let signed = node
        .client
        .sign_raw_transaction_with_wallet(&tx)
        .expect("signrawtransactionwithwallet")
        .into_model()
        .expect("SignRawTransaction into model")
        .tx;
    let send = node.client.send_raw_transaction(&signed).expect("sendrawtransaction");

    let json: AbortPrivateBroadcast =
        node.client.abort_private_broadcast(&send.0).expect("abortprivatebroadcast");
    let model: Result<mtype::AbortPrivateBroadcast, encode::FromHexError> = json.into_model();
    let abort = model.unwrap();

    assert_eq!(abort.removed_transactions, &[signed]);
}
```

tcharding:
Stoked to see this go in. Nice effort.

integration_test/tests/raw_transactions.rs
jamillambert: Done

jamillambert:
Suggested change made.

satsfy (approved):
reACK 0b63ee8d4e758e3cfcbb58fd1835f006fa851bda

The latest force push change is logically equivalent.

tcharding (approved):
ACK 0b63ee8d4e758e3cfcbb58fd1835f006fa851bda

PR #652, opened 2026-06-30
Fix duplicate rustsec issue creation

jamillambert:
When the audit runs on all 3 lockfiles at the same time a problem in multiple lockfiles can result in multiple issues being created for the same RustSec ID.

Run the jobs one at a time so that issues created are seen by the next jobs and no duplicate issue is created.

### DIFF
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -17,6 +17,8 @@ jobs:
       checks: write
     strategy:
       fail-fast: false
+      # Run one at a time, otherwise parallel jobs create the same advisory issue twice.
+      max-parallel: 1
       matrix:
         lockfile: [Cargo.lock, Cargo-minimal.lock, Cargo-recent.lock]
     steps:


.github/workflows/audit.yml
```diff
@@ -17,6 +17,10 @@ jobs:
       checks: write
     strategy:
       fail-fast: false
+      # rustsec/audit-check skips advisories that are already mentioned in an
+      # open issue, but when jobs run concurrently duplicate issues are created.
+      # Run one at a time so that issues created are seen by the next jobs.
```
satsfy: nit: seems verbose, why not "Run one at a time, otherwise parallel jobs create the same advisory issue twice."?

satsfy (approved):
tACK e5889c897b190494f90cb80ff34735106b0b46c9

I tested by running on my fork. It indeed ran jobs in sequence, which were previously run in parallel. The most important part, about issue duplication, can be attested to [here in this CI log run](https://github.com/satsfy/corepc/actions/runs/28558922895/job/84672345367) because the log catches issue duplication emitting log `Seems like RUSTSEC-2026-0190 is mentioned already in the issues/PRs, will not report an issue against it`. [Two](https://github.com/satsfy/corepc/issues/46) [issues](https://github.com/satsfy/corepc/issues/47) were properly created (not duplicates) and no more.

I have just one optional nit.

.github/workflows/audit.yml
jamillambert: Yeah that is better. I changed it.

satsfy (approved):
reACK 2fecab1306ec5e55d3ea30c58b75ce49768e4671

No functional changes and reads cleaner, thanks.

tcharding (approved):
ACK 2fecab1306ec5e55d3ea30c58b75ce49768e4671

Issue #659, opened 2026-07-11
`bitreq::Client` only reuses an HTTP/1.1 connection when the response explicitly contains `Connection: keep-alive`


tnull:
https://github.com/rust-bitcoin/corepc/blob/2d56a779747e0a5df01832af87593ed54d67508c/bitreq/src/connection.rs#L578-L588

HTTP/1.1 connections are persistent by default unless `Connection: close` is present. Servers such as Caddy therefore omit `Connection: keep-alive`. `bitreq` consequently marks a healthy socket unusable and performs a new TCP and TLS handshake for the next request.

Expected behavior:

- HTTP/1.1 or later: reuse unless `Connection: close` is present.
- HTTP/1.0: require explicit `Connection: keep-alive`.

RFC 9112 §9.3: https://www.rfc-editor.org/rfc/rfc9112.html#section-9.3

Related: #562 

Will see to fix as part of #564.

jamillambert:
Issue moved to Forgejo: https://gitea.bitcoin.ninja/rust-bitcoin/corepc/issues/659

PR #660, opened 2026-07-13
`bitreq`: Add RFC 9112 compliance test vectors

tnull:
In #659 we recently discovered that we aren't fully compliant with RFC 9112 / HTTP 1.1.

As a first step of mitigation, we here add some test vectors that cover the central points. We opted to mark the (33!) failing checks as `ignore`d, so we can revisit and fix them in following PRs (instead of cramming everything into one PR).

- Response/cache poisoning: Extra unsolicited bytes can be accepted as the next response (§§6.3, 9.2).
- Request smuggling/injection: Custom methods and headers allow CR/control-character injection; conflicting Host fields are transmitted (§§2.2–5).
- Ambiguous message framing: Requests can contain both Content-Length and Transfer-Encoding; conflicting response lengths are accepted (§§6.1–6.3).
- Incomplete responses accepted: Premature EOF, missing terminal zero chunks, and truncated header sections can be treated as complete (§8).
- Invalid response syntax accepted: Malformed status lines, bare CR, whitespace-prefixed fields, and obsolete line folding are not handled as required (§§2–5).
- Incorrect transfer-coding handling: Coding lists such as gzip, chunked, HTTP/1.0 Transfer-Encoding, and trailers are mishandled (§§6–7).
- Broken connection semantics: HTTP/1.1 persistence incorrectly requires keep-alive; sent Connection: close can be ignored; tokenized options are not parsed correctly (§§9.3, 9.6).
- CONNECT noncompliance: CONNECT uses the wrong request-target, omits the required proxy Host, and treats successful CONNECT responses as ordinary bodies (§§3.2, 6.3).

And more minor:
- Consume 1xx informational responses until the final response.
- Treat clean async EOF as successful for valid close-delimited responses.
- Retry unanswered pipelined requests without immediately re-pipelining.

### DIFF
--- a/bitreq/tests/http_spec_compliance.rs
+++ b/bitreq/tests/http_spec_compliance.rs
@@ -1,0 +1,322 @@
+//! HTTP/1.1 wire-compliance tests for RFC 9112.
+//!
+//! This file is intentionally organized in RFC order. Requirements on servers and forwarding
+//! intermediaries do not apply to bitreq. Sections 10-13, the registries, and the appendices do not
+//! describe protocol behavior implemented by bitreq. TLS record framing and closure alerts from
+//! Sections 9.7 and 9.8 are delegated to the selected TLS backend.
+
+#![cfg(feature = "std")]
+
+use std::io::{Read, Write};
+use std::net::{TcpListener, TcpStream};
+use std::thread;
+use std::time::Duration;
+
+use bitreq::{Error, Method, Request, Response};
+
+const EMPTY_RESPONSE: &[u8] = b"HTTP/1.1 200 OK\r\nContent-Length: 0\r\n\r\n";
+
+fn header_end(bytes: &[u8]) -> Option<usize> {
+    bytes.windows(4).position(|window| window == b"\r\n\r\n").map(|i| i + 4)
+}
+
+fn content_length(head: &[u8]) -> usize {
+    String::from_utf8_lossy(head)
+        .lines()
+        .find_map(|line| {
+            let (name, value) = line.split_once(':')?;
+            name.eq_ignore_ascii_case("content-length")
+                .then(|| value.trim().parse::<usize>().ok())
+                .flatten()
+        })
+        .unwrap_or(0)
+}
+
+fn read_request(stream: &mut TcpStream) -> Vec<u8> {
+    stream.set_read_timeout(Some(Duration::from_secs(5))).unwrap();
+    let mut request = Vec::new();
+    let mut buf = [0u8; 1024];
+    loop {
+        let read = stream.read(&mut buf).unwrap();
+        if read == 0 {
+            break;
+        }
+        request.extend_from_slice(&buf[..read]);
+        if let Some(end) = header_end(&request) {
+            if request.len() >= end + content_length(&request[..end]) {
+                break;
+            }
+        }
+    }
+    request
+}
+
+fn spawn_server(response: Vec<u8>) -> (String, thread::JoinHandle<Vec<u8>>) {
+    let listener = TcpListener::bind("127.0.0.1:0").unwrap();
+    let addr = listener.local_addr().unwrap();
+    let handle = thread::spawn(move || {
+        let (mut stream, _) = listener.accept().unwrap();
+        let request = read_request(&mut stream);
+        stream.write_all(&response).unwrap();
+        request
+    });
+    (format!("http://{addr}"), handle)
+}
+
+fn capture_request(build: impl FnOnce(String) -> Request) -> (Result<Response, Error>, Vec<u8>) {
+    let (base_url, server) = spawn_server(EMPTY_RESPONSE.to_vec());
+    let response = build(base_url).send();
+    (response, server.join().unwrap())
+}
+
+fn response_sync(response: &[u8]) -> Result<Response, Error> {
+    let (url, server) = spawn_server(response.to_vec());
+    let result = bitreq::get(url).send();
+    server.join().unwrap();
+    result
+}
+
+#[cfg(feature = "async")]
+fn response_async(response: &[u8]) -> Result<Response, Error> {
+    let (url, server) = spawn_server(response.to_vec());
+    let runtime = tokio::runtime::Builder::new_current_thread().enable_all().build().unwrap();
+    let result = runtime.block_on(bitreq::get(url).send_async());
+    server.join().unwrap();
+    result
+}
+
+fn check_response_paths(response: &[u8], check: impl Fn(Result<Response, Error>)) {
+    check(response_sync(response));
+    #[cfg(feature = "async")]
+    check(response_async(response));
+}
+
+fn request_text(request: &[u8]) -> &str { std::str::from_utf8(request).unwrap() }
+
+// RFC 9112 Sections 2-5: message syntax, request targets, status lines, and field lines.
+
+#[test]
+fn section_2_1_request_uses_http_message_format() {
+    // RFC 9112 Section 2.1: a message is a start-line, CRLF-delimited fields, an empty line,
+    // and an optional body.
+    let (response, request) = capture_request(|base| {
+        bitreq::post(format!("{base}/resource?answer=42"))
+            .with_header("X-Test", "yes")
+            .with_body(b"body\r\n".to_vec())
+    });
+    response.unwrap();
+    assert!(request.starts_with(b"POST /resource?answer=42 HTTP/1.1\r\n"));
+    let end = header_end(&request).expect("request has a terminating empty line");
+    assert_eq!(&request[end..], b"body\r\n");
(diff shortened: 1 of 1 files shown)


bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+        Box::new(
+            self.windows(needle.len())
+                .enumerate()
+                .filter_map(move |(i, window)| (window == needle).then_some(i)),
+        )
+    }
+}
+
+#[test]
+#[ignore = "TODO: reject bare CR in request protocol elements (RFC 9112 Section 2.2)"]
```
TheBlueMatt, line 144: Do we actually intend to fix all these? I'm happy to see them fixed if we care, but its not clear to me that we do/should, at which point adding tests for them is just noise.

TheBlueMatt (commented):
thanks!

bitreq/tests/http_spec_compliance.rs
tnull: Yes, I think we should. #661 is the start, but I intend to tackle these bit-by-bit in follow-ups.

TheBlueMatt: FWIW I don't think we should bother trying to be super strict on what the calling code passes. In this case specifically where the request is invalid, I think we should just accept it. The only exception would be if the request is bad in a way that will break our request code, ie if it might make two requests because it adds a `\r\n\r\nGET...` or something similar. I think that basically just means rejecting `\r\n`s and *maybe* rejecting standalone `\r` and `\n`s

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+fn section_2_2_sender_rejects_bare_cr() {
+    // RFC 9112 Section 2.2: a sender MUST NOT generate a bare CR in protocol elements.
+    let (response, request) =
+        capture_request(|base| bitreq::get(base).with_header("X-Test", "one\rtwo"));
+    assert!(response.is_err(), "a bare CR must be rejected before transmission");
+    assert!(!request.windows(7).any(|window| window == b"one\rtwo"));
+}
+
+#[test]
+#[ignore = "TODO: reject invalid request field names (RFC 9112 Sections 2.2 and 5)"]
```
TheBlueMatt, line 154: I don't think we should bother rejecting this.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    let authority = base_url.strip_prefix("http://").unwrap();
+    bitreq::get(format!("http://user:password@{authority}/path")).send().unwrap();
+    let request = server.join().unwrap();
+    let text = request_text(&request);
+    assert!(text.contains(&format!("\r\nHost: {authority}\r\n")));
+    assert!(!text.contains("user"));
+}
+
+#[test]
+#[ignore = "TODO: reject conflicting Host fields (RFC 9112 Section 3.2)"]
```
TheBlueMatt, line 190: I don't think we should worry about stuff like this.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    check_response_paths(b"HTTP/1.1 204 \r\nDate: now\r\n\r\n", |result| {
+        let response = result.unwrap();
+        assert_eq!(response.status_code, 204);
+        assert_eq!(response.reason_phrase, "");
+    });
+}
+
+#[test]
+#[ignore = "TODO: validate response status-line grammar (RFC 9112 Sections 2.3 and 4)"]
+fn section_4_invalid_status_lines_are_rejected() {
```
TheBlueMatt, line 269: I don't think we should enforce this.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    let response = b"HTTP/1.1 200 OK\r\nX-Test: one\rtwo\r\nContent-Length: 0\r\n\r\n";
+    check_response_paths(response, |result| {
+        if let Ok(response) = result {
+            assert_eq!(response.headers.get("x-test").map(String::as_str), Some("one two"));
+        }
+    });
+}
+
+#[test]
+#[ignore = "TODO: reject or ignore whitespace-prefixed fields (RFC 9112 Section 2.2)"]
```
TheBlueMatt, line 292: I don't see why we should enforce this.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    });
+    assert!(response.is_err(), "ambiguous request framing must be rejected");
+    let text = request_text(&request).to_ascii_lowercase();
+    assert!(!(text.contains("content-length:") && text.contains("transfer-encoding:")));
+}
+
+#[test]
+#[ignore = "TODO: reject repeated chunked transfer coding (RFC 9112 Section 6.1)"]
+fn section_6_1_sender_rejects_repeated_chunked_coding() {
+    // RFC 9112 Section 6.1: a sender MUST NOT apply chunked transfer coding more than once.
```
TheBlueMatt, line 382: I don't see why we should reject this.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    let (response, request) = capture_request(|base| {
+        bitreq::post(base).with_header("Transfer-Encoding", "chunked, chunked")
+    });
+    assert!(response.is_err(), "repeated chunked coding must be rejected");
+    assert!(!request_text(&request).to_ascii_lowercase().contains("chunked, chunked"));
+}
+
+#[test]
+#[ignore = "TODO: exclude chunked from TE requests (RFC 9112 Section 7.4)"]
```
TheBlueMatt, line 391: Let's please not add parsing of random client headers just to enforce them.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+fn section_7_4_te_does_not_list_chunked() {
+    // RFC 9112 Section 7.4: a client MUST NOT send the chunked coding name in TE.
+    let (response, request) =
+        capture_request(|base| bitreq::get(base).with_header("TE", "trailers, chunked"));
+    assert!(response.is_err(), "TE must not advertise chunked");
+    assert!(!request_text(&request).to_ascii_lowercase().contains("te: trailers, chunked"));
+}
+
+#[test]
+#[ignore = "TODO: add the TE connection option when sending TE (RFC 9112 Section 7.4)"]
```
TheBlueMatt, line 401: Same here.

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    let response = b"HTTP/1.1 100 Continue\r\n\r\n\
+                     HTTP/1.1 200 OK\r\nContent-Length: 5\r\n\r\nhello";
+    check_body_paths(response, |result| assert_eq!(result.unwrap(), b"hello"));
+}
+
+#[test]
+#[ignore = "TODO: treat successful CONNECT responses as tunnels (RFC 9112 Section 6.3)"]
+fn section_6_3_successful_connect_ignores_framing_fields() {
+    // RFC 9112 Section 6.3: a client receiving a successful CONNECT response MUST ignore
+    // Content-Length and Transfer-Encoding because the connection becomes a tunnel.
```
TheBlueMatt, line 435: Are we even going to support parsing this? Its not clear how to parse this if we're pipelining and if a proxy is this broken kinda who cares?

bitreq/tests/http_spec_compliance.rs
```diff
@@ -0,0 +1,956 @@
...
+    // connection close, regardless of Content-Length.
+    let response = b"HTTP/1.1 200 OK\r\nTransfer-Encoding: gzip\r\nContent-Length: 2\r\n\r\nhello";
+    check_body_paths(response, |result| assert_eq!(result.unwrap(), b"hello"));
+}
+
+#[test]
+#[ignore = "TODO: reject Transfer-Encoding in HTTP/1.0 responses (RFC 9112 Section 6.1)"]
+fn section_6_1_http_1_0_transfer_encoding_is_faulty() {
+    // RFC 9112 Section 6.1: a client receiving Transfer-Encoding in HTTP/1.0 MUST treat framing as
+    // faulty and close the connection.
```
TheBlueMatt, line 476: Why is it worth our time to enforce this?

TheBlueMatt (commented):
I skimmed about the first half of these. I think most of them its not worth enforcing - stuff where the server or client are maybe a bit confused but we can parse the resulting stream fine I really don't think we should add code to be strict on the RFC for.

tcharding:

PR #661, opened 2026-07-13
`bitreq`: Fix connection persistence, pool refresh

### LINKED ISSUE #659: `bitreq::Client` only reuses an HTTP/1.1 connection when the response explicitly contains `Connection: keep-alive`

tnull:
Based on #660.
Closes #659.

```
Honor HTTP persistence rules

Reuse HTTP/1.1 connections unless either endpoint requests
closure, and require explicit keep-alive only for HTTP/1.0. This
allows compliant servers to omit redundant keep-alive headers while
still preventing reuse after a close option.
```

```
Refresh async pool entries on reuse 

A cache hit must make its connection the most recently used entry.
Without that refresh, inserting another origin can evict an actively
reused connection and retain an older idle one.
```

### DIFF
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -66,10 +66,13 @@ impl Client {
 
         // Try to get cached connection
         let conn_opt = {
-            let state = self.r#async.lock().unwrap();
+            let mut state = self.r#async.lock().unwrap();
 
             if let Some(conn) = state.connections.get(&owned_key) {
-                Some(Arc::clone(conn))
+                let conn = Arc::clone(conn);
+                state.lru_order.retain(|key| key != &owned_key);
+                state.lru_order.push_back(owned_key.clone());
+                Some(conn)
             } else {
                 None
             }
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -24,11 +24,18 @@ use tokio::sync::Mutex as AsyncMutex;
 
 use crate::request::{ConnectionParams, OwnedConnectionParams, ParsedRequest};
 #[cfg(feature = "async")]
+use crate::response::HttpVersion;
+#[cfg(feature = "async")]
 use crate::Response;
 use crate::{Error, Method, ResponseLazy};
 
 #[cfg(feature = "async")]
 const BACKING_READ_BUFFER_LENGTH: usize = 16 * 1024;
+
+#[cfg(feature = "async")]
+fn has_connection_option(value: &str, option: &str) -> bool {
+    value.split(',').any(|value| value.trim().eq_ignore_ascii_case(option))
+}
 
 type UnsecuredStream = TcpStream;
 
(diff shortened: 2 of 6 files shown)


bitreq/src/connection.rs
```diff
@@ -575,13 +586,19 @@ impl AsyncConnection {
...
                 .await?;
 
-                let mut found_keep_alive = false;
-                if let Some(header) = response.headers.get("connection") {
-                    if header.eq_ignore_ascii_case("keep-alive") {
-                        found_keep_alive = true;
-                    }
-                }
-                if !found_keep_alive {
+                let connection = response.headers.get("connection").map(String::as_str);
```
TheBlueMatt: nit: let's not reallocate the header value but just `eq_ignore_ascii_case` it.

bitreq/src/client.rs
```diff
@@ -66,10 +66,12 @@ impl Client {
 
         // Try to get cached connection
         let conn_opt = {
-            let state = self.r#async.lock().unwrap();
+            let mut state = self.r#async.lock().unwrap();
 
-            if let Some(conn) = state.connections.get(&owned_key) {
-                Some(Arc::clone(conn))
+            if let Some(conn) = state.connections.get(&owned_key).cloned() {
```
TheBlueMatt: nit: I really hate using `clone`/`cloned` on an `Arc`, ideally let's stick with `Arc::clone` so that its clearer what we're doing.

TheBlueMatt (commented):
Thanks, changes in this PR LGTM, had one question in the dep PR though.

bitreq/src/client.rs
tnull: Ah, yeah, agreed.

bitreq/src/connection.rs
tnull: Avoided the allocation, but we can't generally `eq_ignore_ascii_case` at this level (we already do at the right level though).

tnull:
Force-pushed with the following changes:

```diff
diff --git a/bitreq/src/client.rs b/bitreq/src/client.rs
index f2a9bb25..d71c349c 100644
--- a/bitreq/src/client.rs
+++ b/bitreq/src/client.rs
@@ -69,5 +69,6 @@ impl Client {
             let mut state = self.r#async.lock().unwrap();

-            if let Some(conn) = state.connections.get(&owned_key).cloned() {
+            if let Some(conn) = state.connections.get(&owned_key) {
+                let conn = Arc::clone(conn);
                 state.lru_order.retain(|key| key != &owned_key);
                 state.lru_order.push_back(owned_key.clone());
diff --git a/bitreq/src/connection.rs b/bitreq/src/connection.rs
index 1fe7b92c..3b28bad4 100644
--- a/bitreq/src/connection.rs
+++ b/bitreq/src/connection.rs
@@ -587,5 +587,5 @@ impl AsyncConnection {
                 .await?;

-                let connection = response.headers.get("connection").map(String::as_str);
+                let connection = response.headers.get("connection");
                 let received_close =
                     connection.is_some_and(|value| has_connection_option(value, "close"));
```

tcharding:

PR #666, opened 2026-07-16
types: Drop model newtype on Rust primitive types

Abeeujah:
Drops the following types in model which do not model a concrete rust-bitcoin type

- GetBlockCount: newtype on `u64`
- GetDifficulty: newtype on `f64`
- RescanBlockchain: both struct fields (start_height, stop_height) are `u32`


jamillambert (commented):
The changes in total look good, but the commit sequence is a bit confusing to me. For example when you go through them the first one changes a test, but why? the reason that the test needs to be changed hasn't happened yet. And patch 2 doesn't pass `verify` because the verify fix is in a later patch. The work as a whole is good though. I would be happy if it was one patch, but if you want more for personal/green dots in github reasons I am cool with that too, but they need to make sense in order.

Abeeujah:
> I would be happy if it was one patch

Would be squashing right away!

jamillambert (commented):
> * RescanBlockchain: both struct fields (start_height, stop_height) are `u32`

They are both still `i64` in `types`. I assume since everything worked with the conversion to the model where they were `u32` means they can be changed from `i64` to `u32` in types.

jamillambert (approved):
ACK e9dba924d604ffbc60c2b94ac06ff9e7aefcc0ba

tcharding (approved):
ACK e9dba924d604ffbc60c2b94ac06ff9e7aefcc0ba

tcharding:
Thanks man! Nice clear PR, well done.

PR #667, opened 2026-07-16
WIP: Demonstrate usage of the bitcoin-units compat layer

jamillambert:
Depend on the rust-bitcoin compat-layer branch and switch the model layer to stable Amount and Sequence types while preserving the client request helper ergonomics around native bitcoin primitives.

Assisted-by: GPT-5.4

### DIFF
--- a/types/src/v17/wallet/into.rs
+++ b/types/src/v17/wallet/into.rs
@@ -8,8 +8,8 @@ use bitcoin::hex::FromHex;
 use bitcoin::key::{self, PrivateKey, PublicKey};
 use bitcoin::psbt::PsbtParseError;
 use bitcoin::{
-    address, bip32, sign_message, Address, Amount, BlockHash, Psbt, ScriptBuf, SignedAmount,
-    Transaction, Txid, WitnessProgram, WitnessVersion,
+    address, bip32, sign_message, Address, BlockHash, Psbt, ScriptBuf, Transaction, Txid,
+    WitnessProgram, WitnessVersion,
 };
 
 // TODO: Use explicit imports?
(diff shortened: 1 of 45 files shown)


types/src/v17/wallet/into.rs
```diff
@@ -737,7 +740,7 @@ impl WalletCreateFundedPsbt {
         use WalletCreateFundedPsbtError as E;
 
         let psbt = self.psbt.parse::<Psbt>().map_err(E::Psbt)?;
-        let fee = SignedAmount::from_btc(self.fee).map_err(E::Fee)?;
+        let fee = bitcoin::SignedAmount::from_btc(self.fee).map_err(E::Fee)?;
         let change_position = crate::to_u32(self.change_position, "change_position")?;
         Ok(model::WalletCreateFundedPsbt { psbt, fee, change_position })
```
tcharding, line 745: This can be a stable type, right?

tcharding:
Nice! This is pleasant huh. I could bike shed the helper functions and the import alias stuff but as a demo this as awesome, thanks man.

Moving forward we need https://github.com/rust-bitcoin/rust-bitcoin/pull/6248 to land then we can get this in.

By the time we are ready to merge I'd like to see:

- Agree on style of naming stable vs unstable (eg `bitcoin::Amount` / `comtap::Amount` / `StableAmount`)
- Agree / bike shed the helper functions (I see they are super useful when mapping)
- Do all the types in `units 1.0.0.` one per patch across the whole code base (but lets pin everything else down first before we do the whole codebase, I like your 'model module only' approach here to prove it out
- PROFIT!

tcharding:
Please note development has migrated to https://git.rust-bitcoin.org/rust-bitcoin/corepc. Any further comments or pushes here on github may be ignored or lost. Closing since I know @jamillambert  is ok with the migration.

